PR #10, opened 2019-11-08
fix(redhat): stop using Security Data API for patched vulnerabilities

knqyf263:
# Parent PR
https://github.com/aquasecurity/trivy-db/pull/9

# ToDo
- [x] Stop using Security Data API for patched vulnerabilities
- [x] Add tests


knqyf263:
@danielpacak Thanks

PR #18, opened 2019-11-19
test(vulnsrc): add test for vulnsrc.Update

knqyf263:
## Parent
https://github.com/aquasecurity/trivy-db/pull/17


pkg/vulnsrc/vulnsrc.go
```diff
@@ -61,53 +64,89 @@ func init() {
...
 	}
 
-	dbc := db.Config{}
-	dbType := db.TypeFull
-	if light {
-		dbType = db.TypeLight
-	}
-
-	err := dbc.SetMetadata(db.Metadata{
+	fmt.Println(u.dbType)
```
simar7: Remove.

knqyf263: fixed

pkg/vulnsrc/vulnsrc.go
```diff
@@ -61,53 +64,89 @@ func init() {
...
+	dbType := db.TypeFull
+	optimizer = fullOptimizer{dbc: dbConfig}
+
+	if light {
+		dbType = db.TypeLight
+		optimizer = lightOptimizer{dbc: dbConfig}
+	}
+
+	return Updater{
+		dbc:            db.Config{},
```
simar7: Should we reuse `dbConfig` that was previously initialized here?

pkg/vulnsrc/vulnsrc.go
```diff
@@ -61,53 +64,89 @@ func init() {
...
+	dbc            Operation
+	updateMap      map[string]VulnSrc
+	cacheDir       string
+	dbType         db.Type
+	updateInterval time.Duration
+	clock          clock.Clock
+	optimizer      Optimizer
+}
+
+func NewUpdater(cacheDir string, light bool, interval time.Duration) Updater {
```
simar7: Comment: Today this returns an `Updater`, as far as I can see there are no functions of it that do any modifications to any of the `Updater` struct members. Tomorrow this might change and we'll need to return `*Updater` to have those modifications stick.

simar7: But in the spirit of keeping things simple, I'd leave it as is for now. Just wanted to make an observation.

pkg/vulnsrc/vulnsrc.go
```diff
@@ -61,53 +63,88 @@ func init() {
...
-	err := dbc.ForEachSeverity(func(tx *bolt.Tx, cveID string, _ types.Severity) error {
+type Optimizer interface {
+	Optimize() error
+}
+
+type fullOptimizer struct {
+	dbc db.Operations
+}
+
+func (o fullOptimizer) Optimize() error {
```
simar7: Can we include a test for both `Optimize()` versions, light and full? Currently they remain untested.

simar7: My bad. This has been covered https://github.com/aquasecurity/trivy-db/pull/20

pkg/vulnsrc/vulnsrc.go
simar7: Looks like this was resolved in a different PR

PR #22, opened 2019-12-18
Add Dockerfile to build trivy-db locally

mrueg:
This adds a Dockerfile so the trivy-db.gzip can be built locally.

I've made some changes to the github workflow
* use tar.gz instead of the .zip archive for vuln-list (zip is 350MB vs. 121MB as .tar.gz)
* use wget instead of git to get the master tarballs

That should make it faster and I can send a PR for the gh workflow if desired.

Eventually the next step would be to provide a solution to https://github.com/aquasecurity/trivy/issues/294 so trivy can use the container including the trivy.db.gz

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,14 @@
+FROM golang:1.13-alpine as builder
+
+ARG DB_TYPE=trivy
+
+WORKDIR /build
+COPY . /build
+SHELL ["/bin/ash", "-o", "pipefail", "-c"]
+
+RUN apk --no-cache add make
+
+RUN DB_TYPE=${DB_TYPE} make db-all
+
+FROM scratch
+COPY --from=builder /build/assets/trivy*.db.gz .
--- a/Makefile
+++ b/Makefile
@@ -42,3 +42,50 @@ build:
 .PHONY: clean
 clean:
 	rm -rf integration/testdata/fixtures/
+
+$(GOBIN)/bbolt:
+	go get -v github.com/etcd-io/bbolt/cmd/bbolt
+
+export DB_TYPE ?= trivy
+
+ifeq ($(DB_TYPE),trivy-light)
+	DB_ARG := --light
+endif
+
+trivy-db:
+	make build
+
+.PHONY: db-all
+db-all:
+	make build db-fetch
+	make db-build
+	make db-compact
+	make db-compress
+
+.PHONY: db-fetch
+db-fetch:
+	mkdir -p cache/vuln-list cache/ruby-advisory-db cache/rust-advisory-db cache/php-security-advisories cache/nodejs-security-wg cache/python-safety-db
+	wget -qO - https://github.com/aquasecurity/vuln-list/archive/master.tar.gz | tar xz -C cache/vuln-list --strip-components=1
+	wget -qO - https://github.com/rubysec/ruby-advisory-db/archive/master.tar.gz | tar xz -C cache/ruby-advisory-db --strip-components=1
+	wget -qO - https://github.com/RustSec/advisory-db/archive/master.tar.gz | tar xz -C cache/rust-advisory-db --strip-components=1
+	wget -qO - https://github.com/FriendsOfPHP/security-advisories/archive/master.tar.gz | tar xz -C cache/php-security-advisories --strip-components=1
+	wget -qO - https://github.com/nodejs/security-wg/archive/master.tar.gz | tar xz -C cache/nodejs-security-wg --strip-components=1
+	wget -qO - https://github.com/pyupio/safety-db/archive/master.tar.gz | tar xz -C cache/python-safety-db --strip-components=1
+
+.PHONY: db-build
+db-build: trivy-db
+	./trivy-db build $(DB_ARG) --cache-dir cache --update-interval 12h
+
+.PHONY: db-compact
+db-compact: $(GOBIN)/bbolt cache/db/trivy.db
+	mkdir -p assets
+	$(GOBIN)/bbolt compact -o ./assets/$(DB_TYPE).db cache/db/trivy.db
+	rm cache/db/trivy.db
+
+.PHONY: db-compress
+db-compress: assets/$(DB_TYPE).db
+	gzip assets/*
+
+.PHONY: db-clean
+db-clean:
+	rm -rf cache assets
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -28,54 +28,38 @@ jobs:
             dep ensure
         fi
 
-    - name: Prepare dirs
-      run: mkdir cache assets
-
-    - name: Download vuln-list
-      run: |
-        wget --quiet https://github.com/aquasecurity/vuln-list/archive/master.zip
-        unzip -q master.zip
-        mv vuln-list-master cache/vuln-list
-
-    - name: Clone advisories
-      run: |
-        git clone --depth 1 https://github.com/rubysec/ruby-advisory-db.git cache/ruby-advisory-db
-        git clone --depth 1 https://github.com/RustSec/advisory-db.git cache/rust-advisory-db
-        git clone --depth 1 https://github.com/FriendsOfPHP/security-advisories cache/php-security-advisories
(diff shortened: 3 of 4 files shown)


knqyf263:
These improvements are useful for github workflow as well.
>- use tar.gz instead of the .zip archive for vuln-list (zip is 350MB vs. 121MB as .tar.gz)
>- use wget instead of git to get the master tarballs

If you can, would you send a PR?

mrueg:
Sure, I will send one after 12/28.

knqyf263:
Thanks! As for this PR, I'd like to know the use case.

mrueg:
The use case of a Dockerfile is to be able to build a container that includes the database for https://github.com/aquasecurity/trivy/issues/294 This is needed for airgapped environments where the only source you can access is a container image registry (thus you can't access github to download the tarball).

mrueg:
@knqyf263 I'm considering moving the fetch either into the Makefile or into trivy-db directly, so it can be used by github workflow as well as a dockerfile . Any preference from your side?

mrueg:
@knqyf263 any feedback on the Makefile approach?

Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+WORKDIR /build
+COPY . /build
+SHELL ["/bin/ash", "-o", "pipefail", "-c"]
+
+RUN apk --no-cache add make
+
+RUN make build-db
+
+FROM scratch
+COPY --from=builder /build/trivy.db.gz .
```
masahiro331: How about this dockerfile.
Use ARG or ENV.  My preference is ARG.

```
FROM scratch
ARG DBTYPE
ADD https://github.com/aquasecurity/trivy-db/releases/latest/download/${DBTYPE}.db.gz .
```

### build 
Use trivy.db.gz

```
export DBTYPE=trivy
docker built -t ${DBTYPE}-db --build-arg DBTYPE=${DBTYPE} .
```

Use trivy-light.db.gz

```
export DBTYPE=trivy-light
docker built -t ${DBTYPE}-db --build-arg DBTYPE=${DBTYPE} .
```

mrueg: I'd like to avoid adding artifacts to the dockerfile that are volatile, if we can build it from scratch as well. If for some reason the github action pipeline breaks and the release doesn't get published anymore, users can still rebuild from scratch.

masahiro331: Thank you very much.
I think it makes sense.

I have some requests. 
- I would recommend changing the CI to use Makefile
- I think it is better to include Clean target in Makefile.
- Please support trivy-light.db.gz target in Makefile.

mrueg: Thanks, I'll update the PR to include your requested changes. AFAIK, I have no option to test the CI pipeline though.

masahiro331: So sorry, no options.

mrueg: I've added support for the clean target and the light db using DB_TYPE="light" as you suggested.
I'll need to take a look into the CI integration and figure out how to separate the fetching of github vuln db tarballs out.

mrueg:
@masahiro331 I've updated the github workflow and documented building in the README.

Makefile
```diff
@@ -42,3 +42,52 @@ build:
...
+DB_TYPE ?= trivy
+
+ifeq ($(DB_TYPE),trivy-light)
+	DB_ARG := --light
+else
+	undefine DB_ARG
+endif
+
+trivy-db:
+	make build
```
knqyf263: Is this necessary?

Makefile
```diff
@@ -42,3 +42,52 @@ build:
...
+	mkdir -p cache/vuln-list cache/ruby-advisory-db cache/rust-advisory-db cache/php-security-advisories cache/nodejs-security-wg cache/python-safety-db
+	wget -qO - https://github.com/aquasecurity/vuln-list/archive/master.tar.gz | tar xz -C cache/vuln-list --strip-components=1
+	wget -qO - https://github.com/rubysec/ruby-advisory-db/archive/master.tar.gz | tar xz -C cache/ruby-advisory-db --strip-components=1
+	wget -qO - https://github.com/RustSec/advisory-db/archive/master.tar.gz | tar xz -C cache/rust-advisory-db --strip-components=1
+	wget -qO - https://github.com/FriendsOfPHP/security-advisories/archive/master.tar.gz | tar xz -C cache/php-security-advisories --strip-components=1
+	wget -qO - https://github.com/nodejs/security-wg/archive/master.tar.gz | tar xz -C cache/nodejs-security-wg --strip-components=1
+	wget -qO - https://github.com/pyupio/safety-db/archive/master.tar.gz | tar xz -C cache/python-safety-db --strip-components=1
+
+.PHONY: db-build
+db-build: trivy-db
```
knqyf263: Is there any reason you don't specify `build` directly?

```suggestion
db-build: build
```

Makefile
mrueg: This creates a non phony target, so I can ensure that trivy-db as a binary has been built jn later steps.

Makefile
mrueg: If I'd depend on the phony target build here, trivy would get rebuilt every time you call db-build

knqyf263: I understand.

Makefile
```diff
@@ -42,3 +42,52 @@ build:
...
+
+$(GOBIN)/bbolt:
+	go get -v github.com/etcd-io/bbolt/cmd/bbolt
+
+DB_TYPE ?= trivy
+
+ifeq ($(DB_TYPE),trivy-light)
+	DB_ARG := --light
+else
+	undefine DB_ARG
```
knqyf263: `undefine` directive is introduced in `make` 3.82. macOS installs `make` 3.81 as default. Is it possible not to use `undefine`?

mrueg: Good catch, probably that part isn't needed at all. I dropped it.

Makefile
```diff
@@ -42,3 +42,50 @@ build:
...
+
+.PHONY: db-all
+db-all:
+	make build db-fetch
+	make db-build
+	make db-compact
+	make db-compress
+
+.PHONY: db-fetch
+db-fetch:
```
knqyf263: Now, even when security advisories are already under cache directory, `db-fetch` fetches all advisories. Is it possible to avoid it?

Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+ARG DB_TYPE
+
+WORKDIR /build
+COPY . /build
+SHELL ["/bin/ash", "-o", "pipefail", "-c"]
+
+RUN apk --no-cache add make
+
+RUN DB_TYPE=${DB_TYPE} make db-all
```
knqyf263: This line always overwrites `DB_TYPE`. `docker build . -t trivy-db` doesn't seem to work because an empty `DB_TYPE` is passed to `make`.
https://github.com/aquasecurity/trivy-db/pull/22/files#diff-04c6e90faac2675aa89e2176d2eec7d8R35

You can see the error. I think `./assets/.db` should be `./assets/trivy.db`.
```
mkdir -p assets
/go/bin/bbolt compact -o ./assets/.db cache/db/trivy.db
576761856 -> 126279680 bytes (gain=4.57x)
rm cache/db/trivy.db
make[1]: Leaving directory '/build'
make db-compress
make[1]: Entering directory '/build'
gzip assets/*
gzip: assets/*: No such file or directory
make[1]: *** [Makefile:87: db-compress] Error 1
make[1]: Leaving directory '/build'
make: *** [Makefile:63: db-all] Error 2
```

Makefile
mrueg: Initially I wanted to avoid that we have a stale cache directory, so downloading them all seems to be the better option.

I can create a non-phony target for cache/ at the risk of not downloading them at all or having a stale fetch.

Dockerfile
mrueg: Thanks, I'll set the default arg for DB_TYPE to trivy.

Makefile
knqyf263: Fair enough. We can keep it as is.

knqyf263:
Thanks a lot!

PR #32, opened 2020-03-06
Vendor severity map

simar7:
This PR will include Vendor severity for each vendor in trivy-db.

Adresses: https://github.com/aquasecurity/trivy/issues/310


simar7:
Hey @knqyf263 I'd like to get some feedback on the approach so far so I'm going to open a draft PR to have a discussion. 

I've added vendor scores for RedHat and Debian to start off. I'm currently saving Vendor Scores in a map alongside the Severity. An example CVE after the database gets built to look at would be `CVE-2003-1581` where both Debian and RedHat are affected.

simar7:
A quick note on database sizes: The DB sizes for both full and light remain unchanged with the addition of VendorSeverity.

pkg/types/types.go
```diff
@@ -66,14 +68,15 @@ type LastUpdated struct {
...
-	SeverityV3  Severity `json:",omitempty"`
-	References  []string `json:",omitempty"`
-	Title       string   `json:",omitempty"`
-	Description string   `json:",omitempty"`
+	ID             string         `json:",omitempty"` // e.g. CVE-2019-8331, OSVDB-104365
+	CvssScore      float64        `json:",omitempty"`
+	CvssScoreV3    float64        `json:",omitempty"`
+	Severity       Severity       `json:",omitempty"`
+	SeverityV3     Severity       `json:",omitempty"`
+	VendorSeverity VendorSeverity `json:",omitempty"`
```
knqyf263: Why is this necessary?

pkg/vulnsrc/amazon/amazon.go
```diff
@@ -114,11 +114,15 @@ func (vs VulnSrc) commitFunc(tx *bolt.Tx) error {
...
 
+				severity := severityFromPriority(alas.Severity)
 				vuln := types.VulnerabilityDetail{
-					Severity:    severityFromPriority(alas.Severity),
+					Severity:    severity,
 					References:  references,
 					Description: alas.Description,
 					Title:       "",
+					VendorSeverity: types.VendorSeverity{
+						vulnerability.Amazon: severity,
```
knqyf263: It always stores the same value as `Severity`.

pkg/vulnsrc/vulnerability/vulnerability.go
```diff
@@ -44,6 +44,27 @@ func getSeverity(details map[string]types.VulnerabilityDetail) types.Severity {
...
+			continue
+		//case d.CvssScore > 0:
+		//	return scoreToSeverity(d.CvssScore)
+		//case d.CvssScoreV3 > 0:
+		//	return scoreToSeverity(d.CvssScoreV3)
+		//case d.Severity != 0:
+		//	return d.Severity
+		//case d.SeverityV3 != 0:
+		//	return d.SeverityV3
+		case d.VendorSeverity[source] > 0:
```
knqyf263: You can use `d.Severity`, right?

pkg/vulnsrc/vulnerability/vulnerability.go
```diff
@@ -44,6 +44,27 @@ func getSeverity(details map[string]types.VulnerabilityDetail) types.Severity {
 	return types.SeverityUnknown
 }
 
+func getVendorSeverity(details map[string]types.VulnerabilityDetail) types.VendorSeverity {
+	vs := make(types.VendorSeverity)
+	for _, source := range sources {
```
knqyf263: This `sources` should be different from `getSeverity` because some vendors use the same value as NVD. We don't have to store such a severity in DB.

pkg/vulnsrc/vulnerability/vulnerability.go
```diff
@@ -44,6 +44,27 @@ func getSeverity(details map[string]types.VulnerabilityDetail) types.Severity {
...
+		case !ok:
+			continue
+		//case d.CvssScore > 0:
+		//	return scoreToSeverity(d.CvssScore)
+		//case d.CvssScoreV3 > 0:
+		//	return scoreToSeverity(d.CvssScoreV3)
+		//case d.Severity != 0:
+		//	return d.Severity
+		//case d.SeverityV3 != 0:
+		//	return d.SeverityV3
```
knqyf263: The vendors sometimes don't have their own severity yet, especially immediately after the vulnerability is found. To support the case, we should store NVD severity as well. Of course, sometimes NVD doesn't have the severity, while the vendor has.

knqyf263:
How much is it?
>A quick note on database sizes: The DB sizes for both full and light remain unchanged with the addition of VendorSeverity.

knqyf263:
>In the case of light-db which severity will we use?

Yes, it is a problem... I adopted the different schema to reduce the DB size, which is not suitable for vendor severity. We display the different severity in the case of light-db or add another bucket.

simar7:
> How much is it?
> 
> > A quick note on database sizes: The DB sizes for both full and light remain unchanged with the addition of VendorSeverity.

Master:
```
-rw-------  1 simar  staff   3.9M Mar  6 18:30 trivy-light.db.gz
-rw-------  1 simar  staff    15M Mar  6 18:00 trivy.db.gz
```

New: 
```
-rw-------  1 simar  staff   3.9M Mar  6 18:30 trivy-light.db.gz
-rw-------  1 simar  staff    15M Mar  6 18:00 trivy.db.gz
```

pkg/types/types.go
simar7: It is not, I tried to create a new field so that I don't have to change existing logic and as a result keep the diff smaller. Once we figure out the necessary implementation details required, we may be able to remove this.

pkg/vulnsrc/amazon/amazon.go
simar7: Yes, is that not right?

simar7: My idea is to use the vulnerability that comes from vendor like this https://github.com/aquasecurity/trivy/compare/vendor-score-priority?expand=1#diff-d85a008c0dbe874b6d2b0caf6eee787cR53-R58 

If this doesn't make sense, please let me know.

pkg/vulnsrc/vulnerability/vulnerability.go
simar7: What should the sources be then?

pkg/vulnsrc/vulnerability/vulnerability.go
simar7: Are you suggesting we need both `Severity` and `VendorSeverity`?

pkg/types/types.go
simar7: I removed it: https://github.com/aquasecurity/trivy-db/pull/32/commits/0c57dacb01b3db8366a9c3aa07efa6e0ed284122

knqyf263:
@simar7 Thanks! Also, zstd improve the size significantly.

pkg/vulnsrc/vulnsrc.go
```diff
@@ -137,39 +137,58 @@ type Optimizer interface {
...
+		References:  references,
+	}
+
+	// assign vendor severities
+	vs := make(types.VendorSeverity)
+	vd, _ := o.dbConfig.GetVulnerabilityDetail(cveID)
+	for vendor, detail := range vd {
+		vs[vendor] = detail.Severity
+	}
+	vuln.VendorSeverity = vs
```
knqyf263: I think `GetDetai` should return `vs` as well.

pkg/vulnsrc/vulnsrc.go
```diff
@@ -137,39 +137,58 @@ type Optimizer interface {
...
+		Title:       title,
+		Description: description,
+		Severity:    severity.String(),
+		References:  references,
+	}
+
+	// assign vendor severities
+	vs := make(types.VendorSeverity)
+	vd, _ := o.dbConfig.GetVulnerabilityDetail(cveID)
+	for vendor, detail := range vd {
```
knqyf263: As we discussed before, we should not use all vendor severities because some vendors may use the same severity as NVD.

knqyf263: Also, we have to investigate the severity of public databases to detect vulnerabilities of libraries. We have the list here.
https://github.com/aquasecurity/trivy-db/blob/master/Makefile#L69-L73

pkg/vulnsrc/amazon/amazon.go
```diff
@@ -114,8 +114,9 @@ func (vs VulnSrc) commitFunc(tx *bolt.Tx) error {
 					references = append(references, ref.Href)
 				}
 
+				severity := severityFromPriority(alas.Severity)
```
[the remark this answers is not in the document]
simar7: Good catch, no this is old cruft that needs inlining. Fixed here 45b037d82aeda92948f4c4a55820fe340d625a04

pkg/vulnsrc/vulnsrc.go
simar7: OK I updated it here 7270ae49ccc114d62b3109b3b5e5b34c441db0d5

pkg/vulnsrc/vulnsrc.go
simar7: OK. I updated it here 7270ae49ccc114d62b3109b3b5e5b34c441db0d5

pkg/vulnsrc/vulnsrc.go
simar7: > As we discussed before, we should not use all vendor severities because some vendors may use the same severity as NVD.

Even if they do store the same as NVD, I'd prefer keeping the logic simple rather than removing them. To me removing them feels premature optimization. As we can see, even with keeping them the size is hardly any different.


> Also, we have to investigate the severity of public databases to detect vulnerabilities of libraries. We have the list here.
> https://github.com/aquasecurity/trivy-db/blob/master/Makefile#L69-L73

I'm not sure what you imply by this. Could you give an example of how to use these databases? Is it part of this PR or can we move it in a different PR?

knqyf263: They also have severities. If we detect vulnerabilities of npm and show the severity of Red Hat, it is weird. We have to display the severity of the npm database.
https://github.com/nodejs/security-wg/blob/master/vuln/npm/208.json#L20

knqyf263: >Even if they do store the same as NVD, I'd prefer keeping the logic simple rather than removing them.

I got it. We will keep as is.

pkg/vulnsrc/vulnsrc.go
knqyf263: It looks good now.

pkg/vulnsrc/vulnsrc.go
simar7: > They also have severities. If we detect vulnerabilities of npm and show the severity of Red Hat, it is weird. We have to display the severity of the npm database.
> https://github.com/nodejs/security-wg/blob/master/vuln/npm/208.json#L20

As discussed offline, you can see that with this commit it now favours the priority that the vendor score has: https://github.com/aquasecurity/trivy-db/pull/32/commits/99cdc188683d9f1b7e659ab2f9e1fbe669f683eb 

An example output of a CVE 
image

pkg/vulnsrc/vulnerability/vulnerability.go
```diff
@@ -15,15 +15,38 @@ var (
...
+		switch {
+		case detail.Severity != types.SeverityUnknown:
+			vs[vendor] = detail.Severity
+		case detail.CvssScore != 0:
+			vs[vendor] = scoreToSeverity(detail.CvssScore)
+		case detail.CvssScoreV3 != 0:
+			vs[vendor] = scoreToSeverity(detail.CvssScoreV3)
+		default:
+			vs[vendor] = types.SeverityUnknown
+		}
```
knqyf263: Are these lines necessary?

simar7: It is not, if my understanding is correct we can take NVD if vendor is not present. I'll remove this. https://github.com/aquasecurity/trivy-db/pull/32/commits/d65430e7cb51687599c40063d825235b14ad2373

simar7:
@knqyf263 I think I've covered all the comments you had. Once we merge this in, we can start using it in Trivy to prefer vendor severity over NVD if available!

knqyf263:
@simar7 I think we have to implement Trivy side before merging this PR. Especially, I'm wondering if we can display the severity from the databases of programming languages.

simar7:
> @simar7 I think we have to implement Trivy side before merging this PR. Especially, I'm wondering if we can display the severity from the databases of programming languages. It seems to be able to show the severity based on OS of the image, but we should check programming languages as well. If we find any problem on Trivy side, we may need to fix this repository again. This repository affects the databases in GitHub Releases, so we need to be careful of the merge.
> 
> For the debug, you can build this database on your local machine and use it in Trivy.

Roger that. The Trivy PR related to this is available here: https://github.com/aquasecurity/trivy/pull/433

simar7:
As discussed offline, now light DB will also include VendorSeverity: https://github.com/aquasecurity/trivy-db/pull/32/commits/a11aca42bc75aa9a857cb501b073237db52c4ed0

As a result the light DB has marginally increased in size:

Old:
```
➜  assets git:(vendor-severity-map) ✗ ls  -lrth
total 8816
-rw-------  1 simar  staff   3.9M Mar 6 18:30 trivy-light.db.gz
```

New:
```
➜  assets git:(vendor-severity-map) ✗ ls  -lrth
total 8816
-rw-------  1 simar  staff   4.3M Mar 26 18:12 trivy-light.db.gz
```

pkg/vulnsrc/vulnerability/vulnerability.go
```diff
@@ -15,15 +15,36 @@ var (
...
+	}
+
+	return getSeverity(details), getVendorSeverity(details), getTitle(details), getDescription(details), getReferences(details)
+}
+
+func getVendorSeverity(details map[string]types.VulnerabilityDetail) types.VendorSeverity {
+	vs := make(types.VendorSeverity)
+	for vendor, detail := range details {
+		switch {
+		case detail.Severity != types.SeverityUnknown:
```
knqyf263: Could you add SeverityV3 as well?

pkg/vulnsrc/vulnerability/vulnerability_test.go
```diff
@@ -0,0 +1,55 @@
...
+		return map[string]types.VulnerabilityDetail{
+			vulnerability.RedHat: {
+				ID:          "CVE-2020-1234",
+				CvssScore:   4.3,
+				CvssScoreV3: 5.6,
+				Severity:    types.SeverityHigh,
+				SeverityV3:  types.SeverityCritical,
+				Title:       "test vulnerability",
+				Description: "a test vulnerability where vendor rates it lower than NVD",
+			},
```
knqyf263: Recently, Red Hat shows only CVSSv3.
https://access.redhat.com/security/cve/cve-2020-8597

So, we should add the case which doesn't have CvssScore/Severity.

```
			vulnerability.RedHat: {
				ID:          "CVE-2020-1234",
				CvssScoreV3: 5.6,
				SeverityV3:  types.SeverityCritical,
				Title:       "test vulnerability",
				Description: "a test vulnerability where vendor rates it lower than NVD",
			},
```

pkg/vulnsrc/vulnsrc.go
```diff
@@ -137,60 +137,86 @@ type Optimizer interface {
...
 	return nil
 
 }
 
+// TODO: Until we can dependency inject, we need to monkey patch sadly
+var (
+	getDetailFunc = vulnerability.GetDetail
+)
+
+func (o fullOptimizer) fullOptimize(cveID string, tx *bolt.Tx) error {
```
knqyf263: nit: I feel it is natural `tx` is a first argument.
```suggestion
func (o fullOptimizer) fullOptimize(tx *bolt.Tx, cveID string) error {
```

pkg/vulnsrc/vulnsrc.go
```diff
@@ -137,60 +137,86 @@ type Optimizer interface {
...
 		return xerrors.Errorf("failed to delete vulnerability detail bucket: %w", err)
 	}
 	return nil
 }
+
+func (o lightOptimizer) lightOptimize(cveID string, tx *bolt.Tx) error {
+	// get correct severity
+	severity, vendorSeverity, _, _, _ := getDetailFunc(cveID)
+	vuln := types.Vulnerability{
+		Severity:       severity.String(),
```
knqyf263: Is this necessary? We have `Severity` in the full database for backward compatibility because an old trivy client reads the severity from `Severity` key, but it doesn't look necessary in the light database.

pkg/vulnsrc/vulnsrc.go
```diff
@@ -137,60 +137,86 @@ type Optimizer interface {
...
+var (
+	getDetailFunc = vulnerability.GetDetail
+)
+
+func (o fullOptimizer) fullOptimize(cveID string, tx *bolt.Tx) error {
+	severity, vs, title, description, references := getDetailFunc(cveID)
+	vuln := types.Vulnerability{
+		Title:          title,
+		Description:    description,
+		Severity:       severity.String(),
```
knqyf263: We can add a comment here. 
```suggestion
		Severity:       severity.String(), // TODO: We have to keep this key until we deprecate
```

pkg/vulnsrc/vulnerability/vulnerability.go
simar7: Added

pkg/vulnsrc/vulnerability/vulnerability_test.go
simar7: Added a few more tests.

pkg/vulnsrc/vulnsrc.go
simar7: Roger.

pkg/vulnsrc/vulnsrc.go
simar7: It's not but I'm keeping this to make logic flow easier when it comes to using it. Mixing sources by using information from Full DB when it's a request for Light DB and vice versa isn't a good pattern IMO.

pkg/vulnsrc/vulnsrc.go
simar7: Fixed

pkg/vulnsrc/vulnsrc.go
knqyf263: Now, a new client no longer needs `Severity` in both light and full DB, right? Would you explain why sources are mixed a bit more?

simar7: Ah – I got your point. Sorry I misunderstood earlier. I will remove this.

simar7: Removed https://github.com/aquasecurity/trivy-db/pull/32/commits/c47c05449be94fe880be11e46c0f6aed18631e26

pkg/vulnsrc/vulnerability/vulnerability.go
```diff
@@ -15,15 +15,38 @@ var (
...
+			vs[vendor] = detail.Severity
+		case detail.SeverityV3 != types.SeverityUnknown:
+			vs[vendor] = detail.SeverityV3
+		case detail.CvssScore != 0:
+			vs[vendor] = scoreToSeverity(detail.CvssScore)
+		case detail.CvssScoreV3 != 0:
+			vs[vendor] = scoreToSeverity(detail.CvssScoreV3)
```
knqyf263, line 46: I'm wondering why you are using different logic. Is there any reason?
https://github.com/aquasecurity/trivy-db/blob/c47c05449be94fe880be11e46c0f6aed18631e26/pkg/vulnsrc/vulnerability/vulnerability.go#L57-L60

pkg/vulnsrc/vulnsrc.go
```diff
@@ -137,60 +137,85 @@ type Optimizer interface {
 }
 
 type fullOptimizer struct {
-	dbc db.Operation
+	dbConfig db.VulnOperation
```
knqyf263, line 140: Where is this used?

pkg/vulnsrc/vulnsrc.go
```diff
@@ -137,60 +137,85 @@ type Optimizer interface {
...
+
+	if err := o.dbOp.PutVulnerability(tx, cveID, vuln); err != nil {
+		return xerrors.Errorf("failed to put vulnerability: %w", err)
+	}
+	return nil
+}
+
 type lightOptimizer struct {
-	dbc db.Operation
+	dbConfig db.VulnOperation
```
knqyf263, line 186: ditto

pkg/vulnsrc/vulnsrc.go
simar7: https://github.com/aquasecurity/trivy-db/blob/c47c05449be94fe880be11e46c0f6aed18631e26/pkg/vulnsrc/vulnsrc_test.go#L415

pkg/vulnsrc/vulnsrc.go
simar7: https://github.com/aquasecurity/trivy-db/blob/c47c05449be94fe880be11e46c0f6aed18631e26/pkg/vulnsrc/vulnsrc_test.go#L480

pkg/vulnsrc/vulnerability/vulnerability.go
simar7: Actually good catch! This is a bug. The score can be negative as we have seen here https://github.com/nodejs/security-wg/pull/91/files

So this should strictly be `> 0`. I have added a test and fixed here: https://github.com/aquasecurity/trivy-db/pull/32/commits/cb0286151707122fe537bfc7c0c6db2dae979c75

pkg/vulnsrc/vulnsrc.go
knqyf263: I found it in the test code, but I couldn't find it in vulnsrc.go. If it is not called in vulnsrc.go, why is this necessary?

pkg/vulnsrc/vulnerability/vulnerability.go
knqyf263: Yes, we have a negative score.

pkg/vulnsrc/vulnsrc.go
knqyf263: I found it in the test code, but I couldn't find it in vulnsrc.go. If it is not called in business logic, why is this necessary?

simar7: Ah I overlooked. I had initially intended it for being able to inject different behaviours for `GetVulnerabilityDetail` but I never got around finishing it and now it's not needed anymore since I added tests for optimizers on their own with different behaviours of `GetDetailFunc`. I have removed this for now https://github.com/aquasecurity/trivy-db/pull/32/commits/f9156e305e100444602cc79ed5930dee879517fe

pkg/vulnsrc/vulnsrc.go
```diff
@@ -136,60 +136,83 @@ type Optimizer interface {
 }
 
 type fullOptimizer struct {
-	dbc db.Operation
+	dbOp db.Operation
```
knqyf263, line 139: This change makes this PR bigger. I think we should revert this change now. If you actually want to change this name, we can do it in another PR. But if there is a strong reason, we can keep it as is in this PR.
I'm guessing this is because you added and removed `dbConfig`.

simar7: OK I've changed it.

PR #34, opened 2020-03-14
db: Add JSON tags and tests for it

simar7:
### Motivation
Currently if the fields are not set, we will return empty JSON, which isn't so appealing to look at through something like `trivy -version` output.

This PR also  adds tests as there  were none.

Part of: https://github.com/aquasecurity/harbor-scanner-trivy/issues/48

### DIFF
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -49,10 +49,10 @@ type Operation interface {
 }
 
 type Metadata struct {
-	Version    int
-	Type       Type
-	NextUpdate time.Time
-	UpdatedAt  time.Time
+	Version    int        `json:",omitempty"`
+	Type       Type       `json:",omitempty"`
+	NextUpdate *time.Time `json:",omitempty"`
+	UpdatedAt  *time.Time `json:",omitempty"`
 }
 
 type Config struct {
@@ -95,7 +95,7 @@ func (dbc Config) GetVersion() int {
 func (dbc Config) GetMetadata() (Metadata, error) {
 	var metadata Metadata
 	value, err := Config{}.get("trivy", "metadata", "data")
-	if err != nil {
+	if err != nil { // FIXME: get only returns nil, will this ever be true?
 		return Metadata{}, err
 	}
 	if err = json.Unmarshal(value, &metadata); err != nil {
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -49,10 +49,10 @@ type Operation interface {
 }
 
 type Metadata struct {
-	Version    int        `json:",omitempty"`
-	Type       Type       `json:",omitempty"`
-	NextUpdate *time.Time `json:",omitempty"`
-	UpdatedAt  *time.Time `json:",omitempty"`
+	Version    int  `json:",omitempty"`
+	Type       Type `json:",omitempty"`
+	NextUpdate time.Time
+	UpdatedAt  time.Time
 }
 
 type Config struct {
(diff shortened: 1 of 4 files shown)


pkg/db/db.go
```diff
@@ -95,7 +95,7 @@ func (dbc Config) GetVersion() int {
 func (dbc Config) GetMetadata() (Metadata, error) {
 	var metadata Metadata
 	value, err := Config{}.get("trivy", "metadata", "data")
-	if err != nil {
+	if err != nil { // FIXME: get only returns nil, will this ever be true?
```
knqyf263: There's a possibility that `db.View` fails.

pkg/db/db.go
```diff
@@ -49,10 +49,10 @@ type Operation interface {
...
 
 type Metadata struct {
-	Version    int
-	Type       Type
-	NextUpdate time.Time
-	UpdatedAt  time.Time
+	Version    int        `json:",omitempty"`
+	Type       Type       `json:",omitempty"`
+	NextUpdate *time.Time `json:",omitempty"`
+	UpdatedAt  *time.Time `json:",omitempty"`
```
knqyf263: We can't change types easily because old trivy clients will crash.

knqyf263: I think `NextUpdate` and `UpdatedAt` are always not empty, while `Metadata` itself is sometimes empty. I know you want to omit these values if they are empty, but it doesn't have enough effects here. If possible, I'd like to avoid the pointer because we have to do nil check anywhere. If we forget it, it introduces panic. I mean effort and achievement are not balanced.

simar7: You are correct, they aren't empty but have an initialized zero value of: `0001-01-01 00:00:00 +0000 UTC`. 

I also agree we should leave them as is and not use pointers. Simple is better. I was trying to make the output look better by removing them if zero. I agree it's not worth the added complexity of maintaining it.

pkg/vulnsrc/vulnsrc.go
```diff
@@ -119,11 +119,14 @@ func (u Updater) Update(targets []string) error {
...
+	timeUpdatedAt := u.clock.Now().UTC()
+	timeNextUpdate := timeUpdatedAt.Add(u.updateInterval)
+
 	err := u.dbc.SetMetadata(db.Metadata{
 		Version:    db.SchemaVersion,
 		Type:       u.dbType,
-		NextUpdate: u.clock.Now().UTC().Add(u.updateInterval),
-		UpdatedAt:  u.clock.Now().UTC(),
+		NextUpdate: timeNextUpdate,
+		UpdatedAt:  timeUpdatedAt,
```
knqyf263: nit: IMHO, if the meaning of the value is difficult to understand, we should define another variable for ease of understanding. For example,

```
return timeUpdatedAt.Add(u.updateInterval)
```
is a bit unclear, so in this case, I prefer the following.

```
timeNextUpdate := timeUpdatedAt.Add(u.updateInterval)
return timeNextUpdate
```



In this case, I feel it looks the same as the following.
```suggestion
		NextUpdate: u.clock.Now().UTC().Add(u.updateInterval),
		UpdatedAt: u.clock.Now().UTC(),
```

I'm guessing you changed here to get pointers of these values and forgot to change it back, but let me know if you have any reason.

simar7: Yeah I understand. Although doing a `clock.Now()` at two different places will lead to two different values of "now" since when things were about to get updated. In this particular case it's not a big deal. 

Yes I forgot to remove this after the switchback from pointers! Thanks for catching it.

PR #35, opened 2020-03-30
fix(db): convert a fatal error to a panic and recover it

knqyf263:
```
unexpected fault address 0x3132c008
fatal error: fault
[signal SIGBUS: bus error code=0x2 addr=0x3132c008 pc=0x10d7ca0]

goroutine 1 [running]:
runtime.throw(0x1126ccd, 0x5)
        /usr/local/Cellar/go/1.14.1/libexec/src/runtime/panic.go:1114 +0x72 fp=0xc000105c08 sp=0xc000105bd8 pc=0x1030f22
runtime.sigpanic()
        /usr/local/Cellar/go/1.14.1/libexec/src/runtime/signal_unix.go:692 +0x443 fp=0xc000105c38 sp=0xc000105c08 pc=0x10473f3
go.etcd.io/bbolt.(*freelist).read(0xc00011a000, 0x3132c000)
        /Users/teppei/src/github.com/etcd-io/bbolt/freelist.go:285 +0x40 fp=0xc000105d00 sp=0xc000105c38 pc=0x10d7ca0
go.etcd.io/bbolt.(*DB).loadFreelist.func1()
        /Users/teppei/src/github.com/etcd-io/bbolt/db.go:316 +0xcf fp=0xc000105d38 sp=0xc000105d00 pc=0x10e0b5f
sync.(*Once).doSlow(0xc00007c368, 0xc000105d98)
        /usr/local/Cellar/go/1.14.1/libexec/src/sync/once.go:66 +0xec fp=0xc000105d88 sp=0xc000105d38 pc=0x107444c
sync.(*Once).Do(...)
        /usr/local/Cellar/go/1.14.1/libexec/src/sync/once.go:57
go.etcd.io/bbolt.(*DB).loadFreelist(0xc00007c200)
        /Users/teppei/src/github.com/etcd-io/bbolt/db.go:309 +0x6a fp=0xc000105db8 sp=0xc000105d88 pc=0x10d3a9a
go.etcd.io/bbolt.Open(0x112732c, 0x8, 0x180, 0x120fee0, 0x9, 0x20, 0x1104440)
        /Users/teppei/src/github.com/etcd-io/bbolt/db.go:286 +0x36b fp=0xc000105ea0 sp=0xc000105db8 pc=0x10d355b
main.run(0x0, 0x0)
...

goroutine 6 [chan send]:
os/signal.signal_enable(0x953321700000002)
        /usr/local/Cellar/go/1.14.1/libexec/src/runtime/sigqueue.go:219 +0x78
os/signal.enableSignal(...)
        /usr/local/Cellar/go/1.14.1/libexec/src/os/signal/signal_unix.go:51
os/signal.Notify.func2(0x2)
        /usr/local/Cellar/go/1.14.1/libexec/src/os/signal/signal.go:150 +0x8e
os/signal.Notify(0xc00008c000, 0xc0000327b8, 0x1, 0x1)
        /usr/local/Cellar/go/1.14.1/libexec/src/os/signal/signal.go:162 +0x170
github.com/pkg/profile.Start.func10(0xc00006e040)
        /Users/teppei/pkg/mod/github.com/pkg/profile@v1.4.0/profile.go:273 +0x89
created by github.com/pkg/profile.Start
        /Users/teppei/pkg/mod/github.com/pkg/profile@v1.4.0/profile.go:271 +0x1e7

goroutine 17 [runnable]:
os/signal.loop()
        /usr/local/Cellar/go/1.14.1/libexec/src/os/signal/signal_unix.go:21
created by os/signal.Notify.func1
        /usr/local/Cellar/go/1.14.1/libexec/src/os/signal/signal.go:127 +0x44

Process finished with exit code 2
```


pkg/db/db_test.go
```diff
@@ -12,6 +14,62 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func TestInit(t *testing.T) {
+	tests := []struct {
+		name    string
+		dbPath  string
+		wantErr bool
```
danielpacak: Is it used somewhere?

pkg/db/db.go
```diff
@@ -65,6 +66,19 @@ func Init(cacheDir string) (err error) {
 		return xerrors.Errorf("failed to mkdir: %w", err)
 	}
 
+	// bbolt sometimes occurs the fatal error of "unexpected fault address".
+	// In that case, the local DB should be broken and needs to be removed.
+	debug.SetPanicOnFault(true)
```
danielpacak: I'm thinking whether we should set this flag in the upstream trivy project. Somewhere in the main() method.

pkg/db/db_test.go
knqyf263: Nice catch. Not necessary.

pkg/db/db.go
knqyf263: It is reasonable to catch faults at an unexpected (non-nil) address from BoltDB because they use a memory-mapped file, but I'm not sure we should catch all faults in trivy. So, I set this flag only in this function. Then, this flag is disabled in the `defer`. 
https://github.com/aquasecurity/trivy-db/pull/35/files#diff-ac4e04b0d1bbf8235592bdbb82938144R79

Does it make sense to you? Actually, I'm not sure what we should do here.

danielpacak: Yup. Got your point. If this code is not accessed by multiple goroutines we could leave it as it is.

knqyf263: Thanks. This is a function to initialize a database, so it is intended to be called once at the beginning.

PR #40, opened 2020-04-23
Add Ubuntu 13.10 and 20.04

mrueg:
13.10 was missed, 20.04 is the new LTS.


knqyf263:
Hi @mrueg, I don't know why, but GitHub Actions doesn't work. Would you run `git commit --ammend` or commit some dummy change to re-run tests?

mrueg:
@knqyf263 sure, i triggered it.

knqyf263:
Thanks!

PR #41, opened 2020-04-26
fix(vulnsrc): remove an import of aquasecurity/trivy

knqyf263:
https://github.com/aquasecurity/trivy-db/pull/32 imported `aquasecurity/trivy/pkg/vulnsrc/vulnerability` for constants, but it is wrong. Those constants are actually under `aquasecurity/trivy-db/pkg/vulnsrc/vulnerability`. This causes the following issue. `trivy-db` is intended to be imported by `trivy`, so it must not import `trivy` to avoid a cycle import.

```
$ go mod tidy
go: finding module for package github.com/aquasecurity/trivy/pkg/vulnsrc/vulnerability
github.com/aquasecurity/trivy/pkg/vulnerability imports
        github.com/aquasecurity/trivy-db/pkg/vulnsrc/vulnerability tested by
        github.com/aquasecurity/trivy-db/pkg/vulnsrc/vulnerability.test imports
        github.com/aquasecurity/trivy/pkg/vulnsrc/vulnerability: no matching versions for query "latest"
```


simar7:
Ah good catch!

PR #42, opened 2020-04-27
mod: Update bbolt go module name for downstream imports.

simar7:
bbolt project updated it's module name starting v1.3.4: https://github.com/etcd-io/bbolt/commit/e65d4d5d27e82efe8e16175a11ec2f6e8067c990#diff-37aff102a57d3d7b797f152915a6dc16

image

This breaks any imports of v1.3.4. One example is Trivy where trivy-db can no longer if imported. Another way to fix is use replace statements (https://github.com/etcd-io/bbolt/issues/211). But I feel if the name is here to stay, let's just updated it.
This PR updates the imports based on the new module name.


knqyf263:
I want to merge the following PR, so let me merge this PR.
https://github.com/aquasecurity/trivy/pull/476

PR #43, opened 2020-04-28
Include CVSS vector info

simar7:
Addresses: https://github.com/aquasecurity/trivy/issues/92


simar7:
This currently only includes RedHat vectors. Correct me if I'm wrong @knqyf263 but I believe there are others too?

knqyf263:
Did you check actual JSONs? There doesn't seem to be baseMetricsV3 in JSON.
https://github.com/aquasecurity/vuln-list/tree/master/nvd/2020

We use NVD JSON Version 1.0, but they provide Version 1.1 now. We may need to migrate to Version 1.1.
https://nvd.nist.gov/vuln/data-feeds#JSON_FEED

simar7:
> Did you check actual JSONs? There doesn't seem to be baseMetricV3 in JSON.
> https://github.com/aquasecurity/vuln-list/tree/master/nvd/2020
> 


Yes I have. I tried debugging this inside of vuln-list and I do see the `baseMetricV3` JSON object as part of the data received. You can see it in the debugger below. 

image

> We use NVD JSON Version 1.0, but now they provide Version 1.1 including basemetricV3. We may need to migrate to Version 1.1.
> https://nvd.nist.gov/vuln/data-feeds#JSON_FEED
> 

Actually the difference between 1.0 and 1.1 are insignificant and not the cause here. https://gist.github.com/simar7/3578eefa18b8883006e70cf6b2ed1f68

> Would you look into it? The update script is here.
> https://github.com/aquasecurity/vuln-list-update/blob/master/nvd/nvd.go

Now as to why this is not included in the data that we have is unknown to me. My guess is that since `baseMetricV3` was released, the older CVE definitions are not updated. Like for example this https://github.com/aquasecurity/vuln-list/blob/master/nvd/2020/CVE-2020-0001.json was updated on January 14th last and hasn't been since.

How is the generation of definitions run? How are older definitions like the one above upserted?

knqyf263:
The schema looks the same, but did you try to download JSON files for 1.0 and 1.1 from NVD and compare them?
https://nvd.nist.gov/vuln/data-feeds

The reason looks simple to me. I just didn't find CVSSv3 in JSON Version 1.0. For example, this is a JSON file for CVE-2020.
https://nvd.nist.gov/feeds/json/cve/1.0/nvdcve-1.0-2020.json.gz 

We don't have the definition. We download JSON files from NVD and store them in `vuln-list`.
https://github.com/aquasecurity/vuln-list-update/blob/master/nvd/nvd.go#L19-L21

simar7:
> The schema looks the same, but did you try to download JSON files for 1.0 and 1.1 from NVD and compare them?
> https://nvd.nist.gov/vuln/data-feeds
> 
> The reason looks simple to me. I just didn't find CVSSv3 in JSON Version 1.0. For example, this is a JSON file for CVE-2020.
> https://nvd.nist.gov/feeds/json/cve/1.0/nvdcve-1.0-2020.json.gz
> 
> We don't have the definition. We download JSON files from NVD and store them in `vuln-list`.
> https://github.com/aquasecurity/vuln-list-update/blob/master/nvd/nvd.go#L19-L21

You are correct. I updated vuln-list-update with this PR https://github.com/aquasecurity/vuln-list-update/pull/33

knqyf263:
Thanks. After vuln-list is updated, I'll try building the database on this branch, and approve it.

simar7:
> Thanks. After vuln-list is updated, I'll try building the database on this branch, and approve it.

It's updated now, you can see a sample here https://github.com/aquasecurity/vuln-list/blob/master/nvd/2020/CVE-2020-0001.json

pkg/vulnsrc/vulnerability/vulnerability.go
```diff
@@ -20,16 +20,41 @@ var (
...
+		case VectorV2 != "" && VectorV3 == "":
+			vv[vendor] = map[string]string{
+				"v2": VectorV2,
+			}
+		case VectorV2 == "" && VectorV3 != "":
+			vv[vendor] = map[string]string{
+				"v3": VectorV3,
+			}
+		}
+	}
```
knqyf263, line 56: If you add `json:omitempty`, these lines are not necessary, right?

```suggestion
		vv[vendor = CVSSVector {
			V2: VectorV2,
			V3: VectorV3,
		}
```

simar7: It is still needed since it's a nested struct, because omitempty won't work on it. You can see this test will fail if you remove this logic: https://github.com/aquasecurity/trivy-db/pull/43/files#diff-bc5f6ed1d12bdab4d1aeca18ca3003d3R24

I also agree that I don't like this logic and would like to get rid of it, do you have another idea?

pkg/vulnsrc/vulnerability/vulnerability.go
```diff
@@ -20,16 +20,41 @@ var (
...
 
-	return getSeverity(details), getVendorSeverity(details), getTitle(details), getDescription(details), getReferences(details)
+	return getSeverity(details), getVendorSeverity(details), getVendorVectors(details), getTitle(details), getDescription(details), getReferences(details)
+}
+
+func getVendorVectors(details map[string]types.VulnerabilityDetail) types.VendorVectors {
+	vv := make(types.VendorVectors)
+	for vendor, detail := range details {
+		VectorV2 := detail.CvssVector
+		VectorV3 := detail.CvssVectorV3
```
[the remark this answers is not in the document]
simar7: fixed

pkg/types/types.go
```diff
@@ -11,6 +11,9 @@ type Severity int
 
 type VendorSeverity map[string]Severity
 
+type CVSSVector map[string]string
```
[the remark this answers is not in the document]
simar7: I'm not torn on either approach, I can switch to yours.

pkg/vulnsrc/vulnerability/vulnerability.go
knqyf263: I think it is simple to fix it. You just need to ignore the case when both values are empty.

```
		if detail.CvssVector == "" && detail.CvssVectorV3 == "" {
			continue
		}
		vv[vendor] = types.CVSSVector{
			V2: detail.CvssVector,
			V3: detail.CvssVectorV3,
		}
```

simar7: Thanks, that did the trick. I pushed an update: https://github.com/aquasecurity/trivy-db/pull/43/commits/a48e663c7d4669fa065b049fd2cab611e1693e39

pkg/vulnsrc/vulnerability/vulnerability_test.go
```diff
@@ -56,12 +61,21 @@ func TestGetDetail(t *testing.T) {
...
-			expectedDescription:    "a test vulnerability where vendor rates it lower than NVD",
-			expectedRefs:           []string{"http://foo-bar.com/baz"},
+			expectedVendorVectors: types.VendorVectors{
+				RedHat: types.CVSSVector{
+					V2: "AV:N/AC:M/Au:N/C:N/I:P/A:N",
+					V3: "CVSS:3.0/AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:H/A:H",
+				},
+				Ubuntu: types.CVSSVector{
+					V3: "CVSS:3.0/AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:H/A:H",
+				},
```
knqyf263, line 71: nit: `ubuntu` doesn't have CVSS vector, right? It's ok since this is a test, but someone perhaps would misunderstand that ubuntu should return CVSS vector, looking at this expectation. I prefer using actual data, but this is nit-pick.

pkg/types/types.go
```diff
@@ -68,14 +74,16 @@ type LastUpdated struct {
...
-	Severity    Severity `json:",omitempty"`
-	SeverityV3  Severity `json:",omitempty"`
-	References  []string `json:",omitempty"`
-	Title       string   `json:",omitempty"`
-	Description string   `json:",omitempty"`
+	ID           string   `json:",omitempty"` // e.g. CVE-2019-8331, OSVDB-104365
+	CvssScore    float64  `json:",omitempty"`
+	CvssVector   string   `json:",omitempty"`
+	CvssScoreV3  float64  `json:",omitempty"`
+	CvssVectorV3 string   `json:",omitempty"`
```
knqyf263, line 81: info: We might be able to re-use `CVSSVector`
```suggestion
	CvssScore    float64  `json:",omitempty"`
	CvssScoreV3  float64  `json:",omitempty"`
	CvssVector   CVSSVector   `json:",omitempty"`
```

pkg/vulnsrc/vulnerability/vulnerability.go
```diff
@@ -20,16 +20,30 @@ var (
...
+func getVendorVectors(details map[string]types.VulnerabilityDetail) types.VendorVectors {
+	vv := make(types.VendorVectors)
+	for vendor, detail := range details {
+		if detail.CvssVector == "" && detail.CvssVectorV3 == "" {
+			continue
+		}
+		vv[vendor] = types.CVSSVector{
+			V2: detail.CvssVector,
+			V3: detail.CvssVectorV3,
+		}
```
knqyf263, line 44: info: If we re-use CVSSVector, we can write something like the following.

```suggestion
		vv[vendor] = detail.CvssVector
```

But, it's no big deal 😉

pkg/vulnsrc/vulnerability/vulnerability_test.go
simar7: Yeah I see your point. Although it's correct what you say since Ubuntu doesn't have vectors, I only added this because I would like to have more than once vendor to play with in a test.

pkg/vulnsrc/vulnerability/vulnerability.go
simar7: I tried your suggestion but I would suggest let's keep it this way. After trying, I like the analog between the V2 and V3 by having a field for each.

PR #48, opened 2020-05-27
vulnsrc: Add CVSS scores to output in addition to vectors

simar7:
Addresses: https://github.com/aquasecurity/trivy/issues/92

image

### DIFF
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -12,11 +12,16 @@ type Severity int
 type VendorSeverity map[string]Severity
 
 type CVSSVector struct {
-	V2      string  `json:"v2,omitempty"`
-	V2Score float64 `json:"v2_score,omitempty"`
-	V3      string  `json:"v3,omitempty"`
-	V3Score float64 `json:"v3_score,omitempty"`
+	V2     string `json:"v2,omitempty"`
+	V3     string `json:"v3,omitempty"`
+	Scores CVSSScore
 }
+
+type CVSSScore struct {
+	V2 float64 `json:"v2,omitempty"`
+	V3 float64 `json:"v3,omitempty"`
+}
+
 type VendorVectors map[string]CVSSVector
 
 const (
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -12,8 +12,10 @@ type Severity int
 type VendorSeverity map[string]Severity
 
 type CVSSVector struct {
-	V2 string `json:"v2,omitempty"`
-	V3 string `json:"v3,omitempty"`
+	V2      string  `json:"v2,omitempty"`
+	V2Score float64 `json:"v2_score,omitempty"`
+	V3      string  `json:"v3,omitempty"`
+	V3Score float64 `json:"v3_score,omitempty"`
 }
 type VendorVectors map[string]CVSSVector
 
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -11,13 +11,13 @@ type Severity int
 
 type VendorSeverity map[string]Severity
 
-type CVSS struct {
-	V2Vector string  `json:"v2_vector,omitempty"`
-	V2Score  float64 `json:"v2_score,omitempty"`
-	V3Vector string  `json:"v3_vector,omitempty"`
-	V3Score  float64 `json:"v3_score,omitempty"`
+type CVSSVector struct {
+	V2      string  `json:"v2,omitempty"`
+	V2Score float64 `json:"v2_score,omitempty"`
+	V3      string  `json:"v3,omitempty"`
+	V3Score float64 `json:"v3_score,omitempty"`
 }
-type VendorVectors map[string]CVSS
+type VendorVectors map[string]CVSSVector
 
 const (
 	SeverityUnknown Severity = iota
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -18,7 +18,13 @@ type CVSS struct {
 	V3Score float64 `json:"v3_score,omitempty"`
 }
 
+type CVSSVector struct {
+	V2 string `json:"v2,omitempty"`
+	V3 string `json:"v3,omitempty"`
+}
+
 type VendorCVSS map[string]CVSS
+type VendorVectors map[string]CVSSVector
 
 const (
 	SeverityUnknown Severity = iota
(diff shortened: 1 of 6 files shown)


pkg/vulnsrc/vulnerability/vulnerability.go
```diff
@@ -12,6 +12,7 @@ import (
 
 var (
 	sources = []string{Nvd, RedHat, Debian, DebianOVAL, Alpine, Amazon, OracleOVAL, SuseCVRF, Photon,
+		//sources = []string{Nvd, RedHat, Debian, DebianOVAL, Ubuntu, Alpine, Amazon, OracleOVAL, SuseCVRF, Photon,
```
simar7, line 15: @knqyf263 why is `Ubuntu` missing from this list? You can see this test currently fails if it's not included https://github.com/aquasecurity/trivy-db/pull/48/files#diff-bc5f6ed1d12bdab4d1aeca18ca3003d3R112-R143

simar7: I added it: https://github.com/aquasecurity/trivy-db/pull/48/commits/3d3689f41fc146e64a894fe9cdd472f6d46ae747

pkg/types/types.go
```diff
@@ -12,8 +12,10 @@ type Severity int
 type VendorSeverity map[string]Severity
 
 type CVSSVector struct {
-	V2 string `json:"v2,omitempty"`
-	V3 string `json:"v3,omitempty"`
+	V2      string  `json:"v2,omitempty"`
+	V2Score float64 `json:"v2_score,omitempty"`
+	V3      string  `json:"v3,omitempty"`
+	V3Score float64 `json:"v3_score,omitempty"`
```
knqyf263, line 18: I feel score is not vector. How about
```suggestion
	V2      string  `json:"v2,omitempty"`
	V3      string  `json:"v3,omitempty"`
	
type CVSSScore struct {
	V2 float64 `json:"v2,omitempty"`
	V3 float64 `json:"v3,omitempty"`
}
```

Or

```
type CVSS struct {
	V2Score float64 `json:"v2,omitempty"`
	V3Score float64 `json:"v3,omitempty"`
	V2Vector      string  `json:"v2,omitempty"`
	V3Vector     string  `json:"v3,omitempty"`
}
```

simar7: Got it. Let's go with the second approach. https://github.com/aquasecurity/trivy-db/pull/48/commits/53c090dc1bb65c875d16c54efea5b9aa6dd69aa0

simar7: To keep changes backwards compatible, I've changed the approach to introduce a new struct instead (Approach #1) https://github.com/aquasecurity/trivy-db/pull/48/commits/9199281e1c1e51903efc19762830bbfacb240469

knqyf263: How about the following with the first approach?


```
type Vulnerability struct {
	Title          string         `json:",omitempty"`
	Description    string         `json:",omitempty"`
	Severity       string         `json:",omitempty"`
	VendorSeverity VendorSeverity `json:",omitempty"`
	VendorScores  VendorScores  `json:",omitempty"`
	VendorVectors  VendorVectors  `json:",omitempty"`
	References     []string       `json:",omitempty"`
}
```

Or, 

```
type Vulnerability struct {
	Title          string         `json:",omitempty"`
	Description    string         `json:",omitempty"`
	Severity       string         `json:",omitempty"`
	VendorSeverity VendorSeverity `json:",omitempty"`
	VendorVectors  VendorVectors  `json:",omitempty"` // only for backward compatibility
	VendorCVSSs  VendorVectors  `json:",omitempty"`
	References     []string       `json:",omitempty"`
}
```

with the second approach.

To be honest, I feel this is a bit weird 😁

```
type CVSSVector struct {
	V2     string `json:"v2,omitempty"`
	V3     string `json:"v3,omitempty"`
	Scores CVSSScore
}
```

simar7: OK, what do you want `VendorScores` to be of type then? Is it a `map[string]CVSSScore` where `CVSSScore` is `type CVSSScore struct { V2 float64, V3 float64 }` or something else?

If it's the case we'll have to add additional logic to `vulnerability.go` to add a new method called `getVendorScores` which will be called by `GetDetail`. I feel it's unnecessary.

Maybe we could rename like the following, if that works for you:

```
type CVSSVectorsAndScores struct {
	V2     string `json:"v2,omitempty"`
	V3     string `json:"v3,omitempty"`
	Scores CVSSScore
}
```

But I feel this discussion doesn't add much technical value at this point.

knqyf263: Hmm...It looks different from my suggestion... It seems to break compatibility because the same key `VendorVectors` uses a different type `VendorCVSS`. I think an old trivy can't decode a new JSON and returns an error.

simar7: > Hmm...It looks different from my suggestion... It seems to break compatibility because the same key `VendorVectors` uses a different type `VendorCVSS`. I think an old trivy can't decode a new JSON and returns an error.

Sorry I misread your earlier comment. I addressed it here again: https://github.com/aquasecurity/trivy-db/pull/48/commits/febb18e734ec752faa9d9d4773af3af4e784c7a0


I just tried with the master version of Trivy, it works fine (no errors) but it doesn't show any VendorVector information (CVSS Vectors) after this new commit. This is as expected because now we store `CVSS` field instead of `VendorVector` so old Trivy doesn't know about it. 

I addressed it here https://github.com/aquasecurity/trivy-db/pull/48/commits/2e0299ee4f62afdd7384a474618aac79b045f338 now old Trivy can use new DB and display `VendorVectors` and new Trivy can display `CVSS` which includes both.

pkg/types/types.go
```diff
@@ -11,10 +11,19 @@ type Severity int
 
 type VendorSeverity map[string]Severity
 
+type CVSS struct {
+	V2      string  `json:"v2,omitempty"`
```
knqyf263, line 15: I suggested V2Vector because V2 doesn’t mean vector, but do you prefer V2?

simar7: I have no preference, I updated as you requested: https://github.com/aquasecurity/trivy-db/pull/48/commits/14729673df77fc74e1424c295102f05dc629f996

PR #52, opened 2020-07-10
fix(vulnsrc): fix error values referencing wrong vuln src

kgeorgiou:
Noticed that some error values referenced wrong VulnSrc names 🧐 

Also tried to make some of the VulnSrc name capitalization more consistent within their files (e.g `amazon` vs `Amazon`)


knqyf263:
Thanks!

Issue #44, opened 2020-05-10
Split CVE-IDs in safety-db


knqyf263:
safety-db sometimes has multiple CVE-IDs.
https://github.com/pyupio/safety-db/blob/9cb84541a691044512dab8422dc2a36383e0fff3/data/insecure_full.json#L14132

![image](https://user-images.githubusercontent.com/2253692/81504008-a4ec2580-92ef-11ea-91af-42b192b1ee43.png)

They should be stored separately.

knqyf263:
Hi @rahul2393, could you fix it?

rahul2393:
Sure!!

rahul2393:
So for every such case where CVE is like this
```
        {
            "advisory": "Zulip 2.1.2 includes a corrected fix for CVE-2019-19775 (the original fix was affected by an unfixed security bug in Python's urllib, CVE-2015-2104). It also adds authentication for redis and memcached even in configurations where these are running on localhost, for add hardening against attacks from malicious processes running on the Zulip server.",
            "cve": "CVE-2019-19775,CVE-2015-2104",
            "id": "pyup.io-38114",
            "specs": [
                "<2.1.2"
            ],
            "v": "<2.1.2"
        },
```

We need to update is like this?
```
        {
            "advisory": "Zulip 2.1.2 includes a corrected fix for CVE-2019-19775 (the original fix was affected by an unfixed security bug in Python's urllib, CVE-2015-2104). It also adds authentication for redis and memcached even in configurations where these are running on localhost, for add hardening against attacks from malicious processes running on the Zulip server.",
            "cve": "CVE-2019-19775",
            "id": "pyup.io-38114",
            "specs": [
                "<2.1.2"
            ],
            "v": "<2.1.2"
        },
       {
            "advisory": "Zulip 2.1.2 includes a corrected fix for CVE-2019-19775 (the original fix was affected by an unfixed security bug in Python's urllib, CVE-2015-2104). It also adds authentication for redis and memcached even in configurations where these are running on localhost, for add hardening against attacks from malicious processes running on the Zulip server.",
            "cve": "CVE-2015-2104",
            "id": "pyup.io-38114",
            "specs": [
                "<2.1.2"
            ],
            "v": "<2.1.2"
        },
```

knqyf263:
Exactly

PR #59, opened 2020-07-20
Split CVE-IDs in safety-db

rahul2393:
[Issue](https://github.com/aquasecurity/trivy-db/issues/44)

### DIFF
--- a/pkg/vulnsrc/python/python_test.go
+++ b/pkg/vulnsrc/python/python_test.go
@@ -1,0 +1,44 @@
+package python
+
+import (
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/aquasecurity/trivy-db/pkg/db"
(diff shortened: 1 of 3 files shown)


pkg/vulnsrc/python/python_test.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			db.Init("testdata")
+			vulSec := NewVulnSrc()
+			err := vulSec.Update(tc.args.dir)
+			assert.NoError(t, err, tc.name)
+			advisories, err := vulSec.dbc.GetAdvisories(vulnerability.PythonSafetyDB, "zulip")
+			assert.NoError(t, err, tc.name)
+			assert.Equal(t, 7, len(advisories), "expected 7 advisories for zulip found: %v", len(advisories))
```
knqyf263, line 40: I think we have to assert not only the length but only the content.

PR #61, opened 2020-07-22
Split CVE-IDs in safety-db

rahul2393:
[Issue](https://github.com/aquasecurity/trivy-db/issues/44)


pkg/vulnsrc/python/python_test.go
```diff
@@ -0,0 +1,75 @@
...
+			db.Init("testdata")
+			vulSec := NewVulnSrc()
+			err := vulSec.Update(tc.args.dir)
+			assert.NoError(t, err, tc.name)
+			advisories, err := vulSec.dbc.GetAdvisories(vulnerability.PythonSafetyDB, "zulip")
+			assert.NoError(t, err, tc.name)
+			sort.Slice(advisories[:], func(i, j int) bool {
+				return strings.Compare(advisories[i].VulnerabilityID, advisories[j].VulnerabilityID) <= 0
+			})
+			assert.True(t, reflect.DeepEqual(tc.expectedAdvisory,advisories), "expected %v for package zulip, got: %v",tc.expectedAdvisory, advisories)
```
knqyf263, line 71: You can use assert.Equal.

pkg/vulnsrc/python/python_test.go
```diff
@@ -0,0 +1,75 @@
...
+	}{
+		{
+			name: "happy path",
+			args: args{
+				dir: "testdata/fixtures",
+			},
+			expectedAdvisory: []types.Advisory {
+				{
+					VulnerabilityID: "CVE-2018-9986",
+				},
```
knqyf263, line 37: I may be missing something, but why doesn't it include `Spec`?

rahul2393: @knqyf263 can you please help explain what is `Spec`?

pkg/vulnsrc/python/python_test.go
rahul2393: @knqyf263 Done

pkg/vulnsrc/python/python_test.go
```diff
@@ -0,0 +1,79 @@
...
+					VulnerabilityID: "CVE-2020-9444",
+					Specs: []string{"<2.1.3"},
+				},
+			},
+			expectErr: false,
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			db.Init("testdata")
```
knqyf263, line 66:
```suggestion
			err := db.Init("testdata")
			require.NoError(t, err)
```

pkg/vulnsrc/python/python_test.go
```diff
@@ -0,0 +1,79 @@
...
+			vulSec := NewVulnSrc()
+			err := vulSec.Update(tc.args.dir)
+			assert.NoError(t, err, tc.name)
+			advisories, err := vulSec.Get( "zulip")
+			assert.NoError(t, err, tc.name)
+			sort.Slice(advisories[:], func(i, j int) bool {
+				return strings.Compare(advisories[i].VulnerabilityID, advisories[j].VulnerabilityID) <= 0
+			})
+			assert.Equal(t, tc.expectedAdvisory,advisories, "expected %v for package zulip, got: %v",tc.expectedAdvisory, advisories)
+			os.RemoveAll("testdata/db")
```
knqyf263, line 76: If a panic happens above, this will be never called. I think we should put it under the error handling of `db.Init`.

```suggestion
			defer os.RemoveAll("testdata/db")
```

pkg/vulnsrc/python/python_test.go
```diff
@@ -0,0 +1,79 @@
...
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			db.Init("testdata")
+			vulSec := NewVulnSrc()
+			err := vulSec.Update(tc.args.dir)
+			assert.NoError(t, err, tc.name)
+			advisories, err := vulSec.Get( "zulip")
+			assert.NoError(t, err, tc.name)
```
knqyf263, line 71: If `vulnSec.Get` returns an error, `advisories` would be nil. It might cause panic after this line.

```suggestion
			require.NoError(t, err, tc.name)
```

pkg/vulnsrc/python/python_test.go
```diff
@@ -0,0 +1,79 @@
...
+			},
+			expectErr: false,
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			db.Init("testdata")
+			vulSec := NewVulnSrc()
+			err := vulSec.Update(tc.args.dir)
+			assert.NoError(t, err, tc.name)
```
knqyf263, line 69: Also, we perhaps should replace assert with require.

rahul2393: @knqyf263 👍

pkg/vulnsrc/python/python_test.go
rahul2393: Done

Issue #60, opened 2020-07-22
Skip rejected vulnerabilities even when Red Hat doesn't say so


knqyf263:
In the case of CVE-2018-10754, Red Hat doesn't say it is rejected.
https://access.redhat.com/security/cve/CVE-2018-10754

But NVD describes it is rejected.
https://nvd.nist.gov/vuln/detail/CVE-2018-10754

In this case, we have to skip this vulnerability even if Red Hat doesn't say so.

knqyf263:
@rahul2393 Could you look into it?

This PR just skips inserting a vulnerability from NVD, but Red Hat still inserts the vulnerability.
https://github.com/aquasecurity/trivy-db/pull/57

Here are the steps in trivy-db:
- Insert vulnerabilities of NVD into trivy-db
  - https://github.com/aquasecurity/trivy-db/tree/master/pkg/vulnsrc/nvd
- Insert vulnerabilities of Red Hat into trivy-db
  - https://github.com/aquasecurity/trivy-db/tree/master/pkg/vulnsrc/redhat
- Insert vulnerabilities...
- They have duplicate data like title, description, and references. To remove them, trivy-db optimizes all buckets.
  - https://github.com/aquasecurity/trivy-db/blob/master/pkg/vulnsrc/vulnsrc.go#L147

I think it is difficult to skip inserting rejected vulnerabilities in the Red Hat step because the process doesn't know NVD data. We probably have to skip rejected vulnerabilities in the optimization. Note that we have two optimizers.
https://github.com/aquasecurity/trivy-db/blob/master/pkg/vulnsrc/vulnsrc.go#L183
https://github.com/aquasecurity/trivy-db/blob/master/pkg/vulnsrc/vulnsrc.go#L209

If you have any other idea, let me know, please.

rahul2393:
Sure.

rahul2393:
@knqyf263 I agree with you, I think instead of rejecting vulnerability at the source level (which we are doing currently), at optimizer level we need to mark a vulnerability rejected.

PR #62, opened 2020-07-30
Skip vulnerability if marked rejected by any of sources

rahul2393:
[Issue](https://github.com/aquasecurity/trivy-db/issues/60)


pkg/vulnsrc/vulnerability/vulnerability.go
```diff
@@ -176,6 +179,19 @@ func getReferences(details map[string]types.VulnerabilityDetail) []string {
 	return refs
 }
 
+func getRejectedStatus(details map[string]types.VulnerabilityDetail) bool {
+	for _, source := range sources {
+		d, ok := details[source]
+		if !ok {
+			continue
+		}
+		if strings.HasPrefix(d.Description, rejectVulnerability) {
```
knqyf263, line 188: `strings.Contains` may be sufficient. Could you check if `strings.Contains` has a problem? I mean we have to make sure that there is no vulnerability which includes `** REJECT **` in the description, but is not rejected vulnerability.
If we can use `strings.Contains`, we may not need the process adding `** REJECT **` prefix in each file like `nvd.go` and `redhat.go`.

pkg/vulnsrc/vulnsrc.go
```diff
@@ -208,7 +211,10 @@ func (o lightOptimizer) Optimize() error {
 
 func (o lightOptimizer) lightOptimize(cveID string, tx *bolt.Tx) error {
 	// get correct severity
-	vuln := getDetailFunc(cveID)
+	vuln, isRejected := getDetailFunc(cveID)
+	if isRejected {
+		return nil
+	}
```
knqyf263, line 217: In addition, we have to remove the security advisory of this CVE-ID. We have 3 types of buckets.

- vulnerability
  - includes vulnerability detail
  - it is created while optimization
  - it is used to display vulnerability information in a scan result
- vulnerability-detail
  - includes vulnerability detail, but there are duplication
  - it is removed while optimization
- security advisory (such as `Red Hat Enterprise Linux 6`)
  - includes fixed versions for vulnerability detection

We prevent rejected vulnerabilities from being inserted into the `vulnerability` bucket, but those vulnerabilities are already inserted into the `security advisory` bucket. 
https://github.com/aquasecurity/trivy-db/blob/862ec93db08b0bf995c5c0c9e6e08578f668f74b/pkg/vulnsrc/redhat/redhat.go#L144

It means they are detected without detail. We need to remove those CVE-IDs in the optimization process.

pkg/vulnsrc/vulnerability/vulnerability.go
rahul2393: @knqyf263 I prefixed the rejected because in case of redhat we take the first not empty description as vulnerability description, and there can be the case that 2nd or 3rd description specifies the infor for rejection
if we are 100% sure that it won't break, then we can use strings.Contains
https://github.com/aquasecurity/trivy-db/blob/master/pkg/vulnsrc/nvd/nvd.go#L80

pkg/vulnsrc/vulnsrc.go
rahul2393: Instead of putting and deleting the advisory, if I add one field in vulnerability detail for advisory and during optimization add advisories for CVE not rejected will it make sense?

knqyf263: Can we do it even though `Advisory` is a different struct?

pkg/vulnsrc/vulnerability/vulnerability.go
knqyf263: Yes, I know. This is the reason I said as follows.
>Could you check if strings.Contains has a problem? I mean we have to make sure that there is no vulnerability which includes ** REJECT ** in the description, but is not rejected vulnerability.

pkg/vulnsrc/vulnsrc.go
rahul2393: Yes, we can add a field type []Advisory in vulnerability detail for a vulnerability and then we can add it in optimization

knqyf263: OK, so let's go ahead with it, but I'd like to see how it's going to be. Let me review it before you'll write tests and other small fixes.

rahul2393: @knqyf263 Please check, I still need to fix tests

pkg/vulnsrc/vulnerability/vulnerability.go
```diff
@@ -21,15 +27,20 @@ var (
...
 		log.Println(err)
-		return types.Vulnerability{}
+		return types.Vulnerability{}, false
 	} else if len(details) == 0 {
-		return types.Vulnerability{}
+		return types.Vulnerability{}, false
+	}
+	isRejected := getRejectedStatus(details)
+	if !isRejected {
+		if err = saveAdvisories(dbc, tx, vulnID, details); err != nil {
```
knqyf263, line 40: This function is to get detail, so we should not update data.

pkg/vulnsrc/vulnerability/vulnerability.go
```diff
@@ -190,3 +214,19 @@ func scoreToSeverity(score float64) types.Severity {
 		return types.SeverityUnknown
 	}
 }
+
+func saveAdvisories(dbc db.Operation, tx *bolt.Tx, vulnID string, details map[string]types.VulnerabilityDetail) error {
+	for _, source := range sources {
+		d, ok := details[source]
+		if !ok {
+			continue
+		}
```
knqyf263, line 223: We should just iterate `details`.
```suggestion
	for _, d := range details {
```

knqyf263:
@rahul2393 It's a very big change, so I want to make sure there is no difference except for rejected vulnerabilities before and after this change. Could you write a test script to compare two databases? It's like `boltdiff`. It would be great if we can see the result in CI.

script/compare.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+func main() {
+	flag.Parse()
+	oldVuln, oldAdvis := readFile(*oldBboltFile)
+	newVuln, newAdv := readFile(*newBboltFile)
+	fmt.Printf("got %v vulnerabilities and %v advisories from old DB\n", len(oldVuln), len(oldAdvis))
+	fmt.Printf("got %v vulnerabilities and %v advisories from new DB\n", len(newVuln), len(newAdv))
+	for k := range oldVuln {
+		if _, ok := newVuln[k]; !ok {
```
knqyf263, line 26: Could you compare the value as well? You can use `reflect.DeepEqual` or other libraries such as https://github.com/google/go-cmp or https://github.com/kylelemons/godebug.

script/compare.go
```diff
@@ -0,0 +1,70 @@
...
+			fmt.Printf("vulnerability %v does not exist in new %v", k, oldVuln[k])
+		}
+	}
+	for k := range oldAdvis {
+		if _, ok := newAdv[k]; !ok {
```
knqyf263, line 31: ditto

PR #63, opened 2020-08-11
nvd: Add published and last modified timestamps

simar7:


pkg/types/types.go
```diff
@@ -102,14 +104,16 @@ type Advisory struct {
...
+	Title            string         `json:",omitempty"`
+	Description      string         `json:",omitempty"`
+	Severity         string         `json:",omitempty"` // Deprecated: Severity is only for backwards compatibility. Use VendorSeverity instead.
+	CweIDs           []string       `json:",omitempty"` // e.g. CWE-78, CWE-89
+	VendorSeverity   VendorSeverity `json:",omitempty"`
+	VendorVectors    VendorVectors  `json:",omitempty"` // Deprecated: VendorVectors is only for backwards compatibility. Use CVSS instead.
+	CVSS             VendorCVSS     `json:",omitempty"`
+	References       []string       `json:",omitempty"`
+	PublishedDate    string         `json:",omitempty"`
+	LastModifiedDate string         `json:",omitempty"`
```
knqyf263, line 116: I think they should be time.Time, considering a use case.

simar7: I tried that but I realized that the data we save for vuln-list is not a proper time.RFC3339 value https://github.com/aquasecurity/vuln-list/blob/master/nvd/2016/CVE-2016-0016.json#L176-L177

What do you think we should do? Fix vuln-list with proper format or write a custom unmarshaler for trivy-db to decode this format?

Go default json.unmarshal only seems to work on standard time.RFC3339 formats:
```
=== RUN   TestVulnSrc_Update/happy_path
    TestVulnSrc_Update/happy_path: nvd_test.go:69: 
        	Error Trace:	nvd_test.go:69
        	Error:      	Received unexpected error:
        	            	error in NVD walk:
        	            	    github.com/aquasecurity/trivy-db/pkg/vulnsrc/nvd.VulnSrc.Update
        	            	        /Users/simar/repos/trivy-db/pkg/vulnsrc/nvd/nvd.go:53
        	            	  - error in file walk:
        	            	    github.com/aquasecurity/trivy-db/pkg/utils.FileWalk
        	            	        /Users/simar/repos/trivy-db/pkg/utils/utils.go:50
        	            	  - failed to decode NVD JSON:
        	            	    github.com/aquasecurity/trivy-db/pkg/vulnsrc/nvd.VulnSrc.Update.func1
        	            	        /Users/simar/repos/trivy-db/pkg/vulnsrc/nvd/nvd.go:46
        	            	  - parsing time ""2020-01-01T01:01Z"" as ""2006-01-02T15:04:05Z07:00"": cannot parse "Z"" as ":"
        	Test:       	TestVulnSrc_Update/happy_path
```

knqyf263: What if we keep them `string` in `nvd/types.go` and convert `string` into `time.Time` after unmarshal?

simar7: Aha! Good point. I made that change https://github.com/aquasecurity/trivy-db/pull/63/commits/37c4d7cf19d8e5d454b4713e2bc296e4141153c6

knqyf263:
I approved, but there are some conflicts.

PR #64, opened 2020-08-16
Update content depends on target order, sorting targets based on priority

rahul2393:


pkg/vulnsrc/vulnsrc.go
```diff
@@ -114,7 +119,9 @@ func NewUpdater(cacheDir string, light bool, interval time.Duration) Updater {
 
 func (u Updater) Update(targets []string) error {
 	log.Println("Updating vulnerability database...")
-
+	sort.Slice(targets, func(i, j int) bool {
+		return strings.Compare(targets[i], targets[j]) <= 0
+	})
```
knqyf263, line 124: If we sort here, we don't have to sort in `init`.

rahul2393: @knqyf263 Done

PR #65, opened 2020-08-20
test: Add a target to create integration test db

simar7:


Makefile
```diff
@@ -91,3 +91,13 @@ db-compress: assets/$(DB_TYPE)/$(DB_TYPE).db assets/$(DB_TYPE)/metadata.json
 .PHONY: db-clean
 db-clean:
 	rm -rf cache assets
+
+.PHONY: create-test-db
+create-test-db:
+	make db-clean
+	make db-fetch
+	cd cache/vuln-list && git reset --hard 7cce93b3cb1921b9915a4de8ca7dc37d18452806
```
knqyf263, line 99: I'm curious why it works. I thought `master.tar.gz` is not a git repository.
https://github.com/aquasecurity/trivy-db/blob/master/Makefile#L68

knqyf263: @simar7 any progress?

knqyf263: Also, I didn't find 7cce93b3cb1921b9915a4de8ca7dc37d18452806.

knqyf263: How about `50bfd86d29e7357b7987947a876100dbac7fcf33`?

```
commit 50bfd86d29e7357b7987947a876100dbac7fcf33
Author: Travis CI User <travis@example.org>
Date:   Sat Dec 28 10:10:50 2019 +0000

    Debian Security Bug Tracker
```

simar7: Actually you're right, this is not correct. It does a `git reset` on the trivy-db repo which is why it doesn't seem to "fail". `vuln-list` is not a git repo. Let me take a look into this a bit more.

simar7: Any reason you want to use `50bfd86d29e7357b7987947a876100dbac7fcf33 ` why not `8b786696ee5c47495e19f9453b467e729b844a17` (the latest master as of writing this comment)?

simar7: I added the fix here https://github.com/aquasecurity/trivy-db/pull/65/commits/07c27df4a3055b15d2adc011f7345af32df0641b

knqyf263: Ah, OK. I forgot GHSA. I usually look at the difference in vulnerability detected in integration tests before and after a commit as follows.
https://github.com/aquasecurity/trivy/blob/4dc2534dfacdf985154c6a3590ce62633dcc529f/integration/testdata/alpine-310.gitlab.golden

The latest DB produced too many differences and it was difficult to look into all of them. I wanted to use as old a DB as possible. What about https://github.com/aquasecurity/vuln-list/commit/8f40e0ae016df0be4148b1b5936ade4aab06a5bc?

Makefile
```diff
@@ -91,3 +90,22 @@ db-compress: assets/$(DB_TYPE)/$(DB_TYPE).db assets/$(DB_TYPE)/metadata.json
...
+
+.PHONY: db-fetch-vuln-list-master
+db-fetch-vuln-list-master:
+	mkdir -p cache/vuln-list
+	wget -qO - https://github.com/aquasecurity/vuln-list/archive/master.tar.gz | tar xz -C cache/vuln-list --strip-components=1
+
+.PHONY: db-fetch-vuln-list-fixed
+db-fetch-vuln-list-fixed:
+	mkdir -p cache/vuln-list
+	wget -qO - https://github.com/aquasecurity/vuln-list/archive/8b786696ee5c47495e19f9453b467e729b844a17.tar.gz | tar xz -C cache/vuln-list --strip-components=1
```
knqyf263, line 102: I didn't know we can download `tar.gz` specifying a commit hash. Nice!

simar7: Yes I also recently found this out!

Makefile
simar7: That looks OK but I see another issue:
```
2020/10/21 17:53:06 Updating python-safety-db data...
2020/10/21 17:53:06 error in python-safety-db update: failed to update python vulnerabilities: failed to decode JSON: json: cannot unmarshal object into Go value of type []python.RawAdvisory
make[1]: *** [db-build] Error 1
make: *** [create-test-db] Error 2
```

Did safety db change its format?

knqyf263: Ah, yes, I merged the master branch and it should've been addressed.

knqyf263 (approved):

PR #69, opened 2020-08-25
fix(vulnerability): use pointer of time.Time

knqyf263:
When NVD doesn't have information, it displays "0001-01-01T00:00:00" now, but I think we should omit it.

### DIFF
--- a/pkg/utils/time.go
+++ b/pkg/utils/time.go
@@ -1,0 +1,15 @@
+package utils
+
+import (
+	"log"
+	"time"
+)
+
+func MustTimeParse(value string) *time.Time {
+	t, err := time.Parse(time.RFC3339, value)
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	return &t
+}
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.13
 
 require (
 	github.com/BurntSushi/toml v0.3.1
-	github.com/aquasecurity/bolt-fixtures v0.0.0-20200825112230-c0f517aea2ed // indirect
+	github.com/aquasecurity/bolt-fixtures v0.0.0-20200825112230-c0f517aea2ed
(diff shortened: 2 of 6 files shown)


pkg/utils/time.go
```diff
@@ -0,0 +1,15 @@
...
+)
+
+func MustTimeParse(value string) *time.Time {
+	t, err := time.Parse(time.RFC3339, value)
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	return &t
+}
```
simar7, line 15: It's nice to have this function so we can inline its usage as it wraps `time.Parse` which cannot be inlined. But this function is also not test friendly for cases where it will return an error since it will panic out. 

Is there a way to limit the usage of a function in Go to test callers only? IOW, this function can only be called from a `*_test.go` test file and no where else.

knqyf263: >It's nice to have this function so we can inline its usage as it wraps time.Parse which cannot be inlined. But this function is also not test friendly for cases where it will return an error since it will panic out.

No, this is intended to be used in test cases as follows.

```
	testCases := []struct {
		name    string
		want    types.VulnerabilityDetail
	}{
		{
			name:  "happy path",
			want: types.VulnerabilityDetail{
				...
				LastModifiedDate: utils.MustTimeParse("2020-01-01T01:01:00Z"),
				PublishedDate:    utils.MustTimeParse("2001-01-01T01:01:00Z"),
			},
		},
	}
```

A developer passes the string to `MustTimeParse`, so it's not user-input. It simplifies safe initialization of time.Time used in test cases like `regexp.MustCompile`.
https://golang.org/pkg/regexp/#MustCompile

>Is there a way to limit the usage of a function in Go to test callers only? IOW, this function can only be called from a *_test.go test file and no where else.

As far as I know, there is no way to restrict the scope. If we define this function in a test file, it cannot be called from a non-test file, but it also doesn't allow other test files to call it. We have to define the same function in each test file.
@danielpacak Do you know something to do that?

PR #72, opened 2020-09-21
Function getSeverity will always return severity out of CvssV2

GalDevel:
Hello,

While using Trivy, I noticed in a problem in the severity I was getting.
Looking at  CVE-2017-0817, Trivy reported the severity as MEDIUM.
But when looking at NIST, the CvssV3 severity is HIGH, while the old CvssV2 severity is MEDIUM.

This led me to believe that there is a bug in Trivy while composing the vulnerability, after looking at the code and at the CVEs from NVD I found the following:
* In the function [getSeverity](https://github.com/aquasecurity/trivy-db/blob/7cce93b3cb1921b9915a4de8ca7dc37d18452806/pkg/vulnsrc/vulnerability/vulnerability.go#L127) in vulnerability.go - you can clearly see that `case d.CvssScore > 0:` will always occur.
NIST started using CvssV3 around ~2015, but they also keep using CvssV2.
* Function [getVendorSeverity](https://github.com/aquasecurity/trivy-db/blob/7cce93b3cb1921b9915a4de8ca7dc37d18452806/pkg/vulnsrc/vulnerability/vulnerability.go#L110) in vulnerability.go - seems to also suffer this issue.
* [Trivy CVE-2017-0817](https://github.com/aquasecurity/vuln-list/blob/0a6d9e889100b8e677240aa2b611e19e5f82e764/nvd/2017/CVE-2017-0817.json#L221)
[NIST CVE-2017-0817](https://nvd.nist.gov/vuln/detail/CVE-2017-0817)

I would love to hear from you regarding this PR.

Best,
Gal.


knqyf263:
Trivy prioritizes CVSSv2 on purpose. I think CVSSv3 always scores too high. But there is room for further consideration. If many users prefer CVSSv3, we can switch it. cc: @lizrice and @simar7

simar7:
> Trivy prioritizes CVSSv2 on purpose. I think CVSSv3 always scores too high. But there is room for further consideration. If many users prefer CVSSv3, we can switch it. cc: @lizrice and @simar7

My vote would be CVSSv3 as well. At the time of writing that code we still preferred to go with V2 as we didn't have enough data points of user input. But as you can see from these 100+ data points of user input on this [poll](https://twitter.com/lizrice/status/1242394151271960576) by @lizrice  V3 is more preferred.

lizrice:
Agreed, let's move to v3

stevegileno:
Once there is a projected timeframe for completing the move to v3, that would be great to know. Thank you @lizrice !

simar7 (approved):
This lgtm! @knqyf263 would you like to have another look?

simar7:
ping @knqyf263 :)

knqyf263:
I announced the migration. This PR affects Trivy DB just after being merged, so we're planning to merge this PR on October 19th.
https://github.com/aquasecurity/trivy/discussions/678

knqyf263 (approved):

PR #74, opened 2020-10-04
Fix update python vulnerabilities no longer work

### LINKED ISSUE #73: make db-build is broken
```
2020/10/04 00:37:32 Updating python-safety-db data...
2020/10/04 00:37:32 error in python-safety-db update: failed to update python vulnerabilities: failed to decode JSON: json: cannot unmarshal object into Go value of type []python.RawAdvisory
make: *** [db-build] Error 1
Error: Makefile:77: recipe for target 'db-build' failed
Error: Process completed with exit code 2.
```

https://github.com/aquasecurity/trivy-db/actions/runs/287205703

The root case is:

* https://github.com/pyupio/safety-db/commit/b412631d3d1b8fc873804d78f75c3f187a745989 changed the format of `insecure_full.json` .
* https://github.com/RustSec/advisory-db/issues/414 changed the format of advisory-db.

mozillazg:
Refer 
* #73 
* https://github.com/pyupio/safety-db

### DIFF
--- a/pkg/vulnsrc/python/python.go
+++ b/pkg/vulnsrc/python/python.go
@@ -2,6 +2,7 @@ package python
 
 import (
 	"encoding/json"
+	"io"
 	"os"
 	"path/filepath"
 	"strings"
@@ -68,8 +69,8 @@ func (vs VulnSrc) update(repoPath string) error {
 
 	// for detecting vulnerabilities
 	var advisoryDB AdvisoryDB
-	if err = json.NewDecoder(f).Decode(&advisoryDB); err != nil {
-		return xerrors.Errorf("failed to decode JSON: %w", err)
+	if advisoryDB, err = decodeAdvisoryDB(f); err != nil {
+		return xerrors.Errorf("failed to decode AdvisoryDB: %w", err)
 	}
 
 	// for displaying vulnerability detail
@@ -135,4 +136,25 @@ func (vs VulnSrc) Get(pkgName string) ([]Advisory, error) {
 		results = append(results, advisory)
 	}
 	return results, nil
+}
+
+func decodeAdvisoryDB(r io.Reader) (AdvisoryDB, error) {
+	var obj map[string]interface{}
+	advisoryDB := AdvisoryDB{}
+	if err := json.NewDecoder(r).Decode(&obj); err != nil {
+		return nil, xerrors.Errorf("failed to decode JSON: %w", err)
+	}
+	for k, v := range obj {
+		slice, ok := v.([]interface{})
+		if !ok {
+			continue
+		}
+		b, _ := json.Marshal(slice)
+		var ad []RawAdvisory
+		if err := json.Unmarshal(b, &ad); err != nil {
+			return nil, xerrors.Errorf("failed to decode JSON: %w", err)
+		}
+		advisoryDB[k] = ad
+	}
+	return advisoryDB, nil
 }
--- a/pkg/vulnsrc/python/python.go
+++ b/pkg/vulnsrc/python/python.go
@@ -2,7 +2,6 @@ package python
 
 import (
 	"encoding/json"
-	"io"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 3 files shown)


pkg/vulnsrc/python/python.go
```diff
@@ -136,3 +137,24 @@ func (vs VulnSrc) Get(pkgName string) ([]Advisory, error) {
 	}
 	return results, nil
 }
+
+func decodeAdvisoryDB(r io.Reader) (AdvisoryDB, error) {
```
knqyf263, line 141: What if defining `UnmarshalJSON` of `AdvisoryDB`?

pkg/vulnsrc/python/python.go
```diff
@@ -136,3 +137,24 @@ func (vs VulnSrc) Get(pkgName string) ([]Advisory, error) {
...
+	var obj map[string]interface{}
```
knqyf263, line 142:
```suggestion
	var obj map[string]json.RawMessage
```

Also, we may be able to use `json.RawMessage` in order to avoid marshaling and unmarshaling.
https://play.golang.org/p/4VFqyBdH9Iu

pkg/vulnsrc/python/python.go
```diff
@@ -136,3 +137,24 @@ func (vs VulnSrc) Get(pkgName string) ([]Advisory, error) {
...
+	advisoryDB := AdvisoryDB{}
+	if err := json.NewDecoder(r).Decode(&obj); err != nil {
+		return nil, xerrors.Errorf("failed to decode JSON: %w", err)
+	}
+	for k, v := range obj {
+		slice, ok := v.([]interface{})
+		if !ok {
+			continue
+		}
```
knqyf263, line 151:
```suggestion
		if k == "$meta" {
			continue
		}
```

This is more readable to me. I know your implementation can handle when other data structures are added in the future. But if `$meta` will be changed to slice, an error occurs. Both implementations have pros and cons. If a prefix `$` always means metadata, we can write it as follows.

```
		if strings.HasPrefix(k) == "$" {
			continue
		}
```

At the moment, we can just skip `$meta`.

knqyf263 (commented):
Thanks for the contribution!

pkg/vulnsrc/python/python.go
mozillazg: ok

pkg/vulnsrc/python/python.go
mozillazg: ok

pkg/vulnsrc/python/python_test.go
```diff
@@ -92,3 +93,52 @@ func TestVulnSrc_Update(t *testing.T) {
 		})
 	}
 }
+
+func Test_decodeAdvisoryDB(t *testing.T) {
```
knqyf263, line 97:
```suggestion
func TestAdvisoryDB_ UnmarshalJSON(t *testing.T) {
```

pkg/vulnsrc/python/python_test.go
```diff
@@ -92,3 +93,52 @@ func TestVulnSrc_Update(t *testing.T) {
...
+	b, _ := ioutil.ReadFile("testdata/fixtures/python-safety-db/data/insecure_full.json")
```
knqyf263, line 98:
```suggestion
	b, err := ioutil.ReadFile("testdata/fixtures/python-safety-db/data/insecure_full.json")
	require.NoError(t, err)
```

pkg/vulnsrc/python/python_test.go
```diff
@@ -92,3 +93,52 @@ func TestVulnSrc_Update(t *testing.T) {
...
+			{
+				ID:       "pyup.io-36979",
+				Advisory: "kuber 9.0.0a1 bumps urllib3 version to pick up security fix for CVE-2018-20060.",
+				Cve:      "",
+				Specs:    []string{"<9.0.0a1"},
+				Version:  "<9.0.0a1",
+			},
+		},
+	}
+	require.Equal(t, expectedAdvisory, advisories, "expected %v, got: %v", expectedAdvisory, advisories)
```
knqyf263, line 143:
```suggestion
	assert.Equal(t, expectedAdvisory, advisories)
```

`"expected %v, got: %v"` will be displayed automatically by `testify`

knqyf263 (changes requested):

knqyf263 (approved):
Great contribution! Thanks a lot!!

PR #75, opened 2020-10-04
Fix update rust vulnerabilities no longer work

mozillazg:
Refer

* #73
* https://github.com/RustSec/advisory-db/issues/414


pkg/vulnsrc/cargo/cargo.go
```diff
@@ -92,14 +94,17 @@ func (vs VulnSrc) walk(tx *bolt.Tx, root string) error {
 		}
 		buf, err := ioutil.ReadFile(path)
 		if err != nil {
-			return xerrors.Errorf("failed to read a file: %w", err)
+			return xerrors.Errorf("failed to read a file(%s): %w", path, err)
 		}
+		codeBlock, title, description := parseAdvisoryMarkdown(buf)
```
mozillazg, line 99: Can't define `UnmarshalTOML` like `UnmarshalJSON`

pkg/vulnsrc/cargo/cargo.go
```diff
@@ -145,3 +150,53 @@ func (vs VulnSrc) Get(pkgName string) ([]Advisory, error) {
 	}
 	return results, nil
 }
+
+func parseAdvisoryMarkdown(src []byte) (codeBlock, title, description string) {
+	md := markdown.New()
+	tokens := md.Parse(src)
```
knqyf263, line 156: `github.com/golang-commonmark/markdown` adds more dependencies. Less dependency is less problem. Can we go with the second way according to this comment? I believe an advisory consists of one code block and one section. I'm wondering if we can parse it ourselves. It looks a bit overkill in our case. If it is difficult, we can use the markdown library.
https://github.com/RustSec/advisory-db/issues/414#issuecomment-702197689

mozillazg: No longer depend on github.com/golang-commonmark/markdown

pkg/vulnsrc/cargo/cargo.go
```diff
@@ -22,6 +24,10 @@ const (
 	cargoDir = "rust-advisory-db"
 )
 
+// https://github.com/RustSec/advisory-db/issues/414#issuecomment-702197689
+var regexAdvisoryMarkdown = regexp.MustCompile(
+	`(?sm)^\x60\x60\x60toml\s*?$(.*?)^\x60\x60\x60\s*?$\s*^#\s*([^\r\n]+)\s*?$(.*)`)
```
knqyf263, line 29: I feel `regex` is difficult to maintain. Could you use `scanner.Scan` and `scanner.Text`?
https://golang.org/pkg/bufio/#Scanner

I'm sorry to bother you again. If you are busy, I'll work on it since we have to fix this issue as soon as possible. Note that we have to handle an empty line well.

mozillazg: Have changed use bufio.Scanner instead of regexp.

mozillazg:
Feel free to work on it if there are any issues.

knqyf263 (approved):

PR #80, opened 2020-10-14
remove trivy-db binary from repo

### LINKED ISSUE #79: Binary committed to the repo?
Hi @simar7 any reason why the trivy-db binary is part of the repo - https://github.com/aquasecurity/trivy-db/blob/master/cmd/trivy-db/trivy-db? I have a PR to remove it, please feel free to close it, if it's irrelevant. Thanks!

neha-viswanathan:
Closes #79


knqyf263 (approved):
I think there is no reason to store the binary. @simar7?

simar7:
ah yes – you're right, this is cruft and probably was accidentally added. thanks for the PR!

simar7 (approved):

PR #82, opened 2020-10-14
feat: use golangci-lint-action

### LINKED ISSUE #81: Use golangci GitHub Action
We can replace [this step](https://github.com/aquasecurity/trivy-db/blob/086d97d517369361ca78e5f063f51854b0cb2b58/.github/workflows/go.yml#L18) with the [`golangci-lint-action`](https://github.com/golangci/golangci-lint-action).

I shall submit a PR if this a valid issue.

sudiptog81:
* utilise golangci-lint-action instead of installing and running it
* closes #81

### DIFF
--- a/pkg/vulnsrc/python/python_test.go
+++ b/pkg/vulnsrc/python/python_test.go
@@ -14,9 +14,6 @@ import (
 )
 
 func TestVulnSrc_Update(t *testing.T) {
-	type args struct {
-		dir string
-	}
 	testCases := []struct {
 		name string
 		args struct {
--- a/pkg/vulnsrc/python/python_test.go
+++ b/pkg/vulnsrc/python/python_test.go
@@ -14,11 +14,12 @@ import (
 )
 
 func TestVulnSrc_Update(t *testing.T) {
+	type args struct {	
+		dir string	
+	}
 	testCases := []struct {
 		name string
-		args struct {
-			dir string
-		}
+		args args
 		expectedAdvisory []Advisory
 		expectErr        bool
 	}{
--- a/pkg/vulnsrc/python/python_test.go
+++ b/pkg/vulnsrc/python/python_test.go
@@ -14,8 +14,8 @@ import (
 )
 
 func TestVulnSrc_Update(t *testing.T) {
-	type args struct {	
-		dir string	
+	type args struct {
+		dir string
 	}
 	testCases := []struct {
 		name string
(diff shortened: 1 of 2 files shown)


sudiptog81:
These error logs might be of interest.

```logs
##[error]Error return value of `(*go.etcd.io/bbolt.Bucket).ForEach` is not checked (errcheck)
##[error]Error return value of `(*go.etcd.io/bbolt.Bucket).ForEach` is not checked (errcheck)
##[error]S1031: unnecessary nil check around range (gosimple)
##[error]type `args` is unused (unused)
```

knqyf263:
As for `errcheck`, you can add `nolint`. We probably should fix errors from `gosimple` and `unused`.

sudiptog81:
The following errors might give an insight. Ran `golangci-lint run --fix -D errcheck` to get this.

``` bash
pkg/vulnsrc/python/python_test.go:78:5: S1031: unnecessary nil check around range (gosimple)
                                if adv != nil {
                                ^
pkg/vulnsrc/python/python_test.go:17:7: type `args` is unused (unused)
```

knqyf263:
You can remove the if statement.

sudiptog81:
The test is passing now.

knqyf263:
Do you have any reason to disable `unused`?

sudiptog81:
The tests were failing due to https://github.com/aquasecurity/trivy-db/blob/master/pkg/vulnsrc/python/python_test.go#L17-L19 which may be necessary but causes the `unused` linter rule to fail.

knqyf263:
It looks like unnecessary. You can remove it.

sudiptog81:
I had gone ahead and removed those lines. The test is failing a https://github.com/aquasecurity/trivy-db/runs/1271265059#step:4:70.

pkg/vulnsrc/python/python_test.go
```diff
@@ -14,9 +14,6 @@ import (
 )
 
 func TestVulnSrc_Update(t *testing.T) {
-	type args struct {
-		dir string
-	}
 	testCases := []struct {
 		name string
 		args struct {
```
knqyf263, line 19: What if defining this struct outside like before? And use it as follows.

```
	type args struct {	
		dir string	
	}
	...
	testCases := []struct {
	...
		args args
```

sudiptog81: The tests are now passing after making the suggested changes.

knqyf263 (approved):

knqyf263:
Thanks.

PR #83, opened 2020-10-19
CSharp constant in const.go

Johannestegner:
As a part in https://github.com/aquasecurity/trivy/pull/686, this pull request adds `CSharp` constant in the `vulnsrc/vulnerability/const.go` file.

### DIFF
--- a/pkg/vulnsrc/vulnerability/const.go
+++ b/pkg/vulnsrc/vulnerability/const.go
@@ -34,4 +34,5 @@ const (
 	Pip      = "pip"
 	RubyGems = "rubygems"
 	Cargo    = "cargo"
+	Dotnet   = "dotnet"
 )


chgl:
this is probably quite nit-picky, but should it be "c-sharp" or something more generic like "dotnet"? It seems like this will support scanning any nuget/.net projects irrespective of the language.

Johannestegner:
Hi there @chgl! That was my initial thought too actually, but seeing all the other language constants was a language rather than a ecosystem, I wanted to keep it in line with the standard :)

simar7 (approved):

knqyf263:
Could you replace c-sharp with dotnet?

knqyf263:
Unfortunately, there is a conflict now. Would you resolve it?

Johannestegner:
> 
> 
> Unfortunately, there is a conflict now. Would you resolve it?

Haha, yes ofc, I should have noticed that :)

Johannestegner:
Rebased :)

knqyf263:
Thanks! Please let me merge the PR after finalizing https://github.com/aquasecurity/trivy/pull/686. We may need further change.

knqyf263:
Oops, I migrated from master to main and this PR was deleted by mistake...

Johannestegner:
No problems! As long as the const gets in it's all good :)

PR #85, opened 2020-10-24
fix(ghsa): handle empty CVE-ID

knqyf263:
GHSA recently added an empty CVE-ID as follows.

```
      {
        "Type": "CVE",
        "Value": ""
      }
```

https://github.com/aquasecurity/vuln-list/commit/1972b6aa3742ccf03b69c4cc1ea98756e3df5fe9#diff-6cf63dea72daa86aeb93472481cace60e174754ca91d55a2005092d1fe169518

Inserting into the DB has been failing because this was an unexpected value.


knqyf263:
@simar7 It was hotfix and I merged this PR without approval. Let me know if you have any feedback even after the PR was merged.

PR #89, opened 2020-12-17
Fix github cron workflow

### LINKED ISSUE #87: Automated vuln-db publishing has broken
I got an alert that our trivy vulndb wasn't fresh anymore. Seems like the last release was three days ago, when they'd normally be every 12 hours.

image

Looking at the github actions history, it looks like they've just [started failing](https://github.com/aquasecurity/trivy-db/runs/1554027650?check_suite_focus=true): 

```
Run make db-fetch
make: *** No rule to make target 'db-fetch'.  Stop.
Error: Process completed with exit code 2.
```

image

It seems the `db-fetch` Makefile target was changed in @simar7's PR here: https://github.com/aquasecurity/trivy-db/pull/65/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L60

Sounds like we probably want to update the github actions script to use the new `make db-fetch-langs` and possibly `db-fetch-vuln-list-master` too??

iamdanfox:
The Makefile targets were renamed in this PR https://github.com/aquasecurity/trivy-db/pull/65 but the github cron workflow is trying to call the old one and failing.

fixes #87

### DIFF
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -21,7 +21,7 @@ jobs:
       uses: actions/checkout@v1
 
     - name: Download vuln-list and advisories
-      run: make db-fetch
+      run: make db-fetch-langs db-fetch-vuln-list-main
 
     - name: Build the binary
       run: make build


knqyf263 (approved):

knqyf263:
oops. Thank you so much!

iamdanfox:
No problemo, trivy has been absolute pleasure to use, thanks for your work!

PR #108, opened 2021-03-06
Support Ubuntu 20.10

mozillazg:
Ref https://github.com/aquasecurity/trivy/pull/876#issuecomment-791881220


knqyf263:
Ubuntu advisories just say "groovy", not "20.10". It would be great if we can get the mapping anywhere so that we wouldn't manage the mapping manually. Anyway, thanks for your great contribution!
https://github.com/aquasecurity/vuln-list/blob/main/ubuntu/2021/CVE-2021-0308.json#L35-L38

knqyf263 (approved):

PR #111, opened 2021-03-21
feat: support GitLab Advisory Database

masahiro331:
https://github.com/aquasecurity/trivy/issues/940


pkg/vulnsrc/glad/glad_test.go
```diff
@@ -0,0 +1,74 @@
...
+		},
+		{
+			name: "happy path maven",
+			dir:  filepath.Join("testdata", "happy"),
+			wantValues: []want{
+				{
+					key:   []string{"advisory-detail", "CVE-2018-1196", "maven::GitLab Advisory Database", "org.springframework.boot:spring-boot"},
+					value: `{"PatchedVersions":["1.5.10.RELEASE"],"VulnerableVersions":["(,1.5.10)"]}`},
+			},
+		},
```
knqyf263, line 43: Looks like they can be aggregated.

pkg/vulnsrc/glad/glad_test.go
```diff
@@ -0,0 +1,74 @@
...
+		wantValues []want
+		wantErr    string
+	}{
+		{
+			name: "happy path go",
+			dir:  filepath.Join("testdata", "happy"),
+			wantValues: []want{
+				{
+					key:   []string{"advisory-detail", "CVE-2016-1905", "go::GitLab Advisory Database", "k8s.io/kubernetes"},
+					value: `{"PatchedVersions":["v1.2.0"],"VulnerableVersions":["\u003cv1.2.0"]}`,
```
knqyf263, line 31: It might be better to define this value as a golden file.

pkg/vulnsrc/glad/glad_test.go
```diff
@@ -0,0 +1,74 @@
...
+				},
+			},
+		},
+		{
+			name: "happy path maven",
+			dir:  filepath.Join("testdata", "happy"),
+			wantValues: []want{
+				{
+					key:   []string{"advisory-detail", "CVE-2018-1196", "maven::GitLab Advisory Database", "org.springframework.boot:spring-boot"},
+					value: `{"PatchedVersions":["1.5.10.RELEASE"],"VulnerableVersions":["(,1.5.10)"]}`},
```
knqyf263, line 41: ditto

pkg/vulnsrc/glad/glad_test.go
```diff
@@ -0,0 +1,74 @@
...
+	tests := []struct {
+		name       string
+		dir        string
+		wantValues []want
+		wantErr    string
+	}{
+		{
+			name: "happy path go",
+			dir:  filepath.Join("testdata", "happy"),
+			wantValues: []want{
```
knqyf263, line 28: We should test the `vulnerability-test` bucket as well.

masahiro331: Is it different from vulnerability-detail?

knqyf263 (approved):

Issue #113, opened 2021-04-21
feat: add CPE


knqyf263:
Add CPE information to the `vulnerability` bucket.
https://github.com/aquasecurity/vuln-list/blob/main/nvd/2021/CVE-2021-0109.json#L6-L24

PR #114, opened 2021-04-21
Added NVD CPE details

rahul2393:
[Issue](https://github.com/aquasecurity/trivy-db/issues/113)

### DIFF
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -95,6 +95,7 @@ type VulnerabilityDetail struct {
 	Description      string     `json:",omitempty"`
 	PublishedDate    *time.Time `json:",omitempty"`
 	LastModifiedDate *time.Time `json:",omitempty"`
+	CPEDetails       string     `json:",omitempty"`
 }
 
 type AdvisoryDetail struct {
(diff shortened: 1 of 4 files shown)


pkg/types/types.go
```diff
@@ -95,6 +95,7 @@ type VulnerabilityDetail struct {
 	Description      string     `json:",omitempty"`
 	PublishedDate    *time.Time `json:",omitempty"`
 	LastModifiedDate *time.Time `json:",omitempty"`
+	CPEDetails       string     `json:",omitempty"`
```
knqyf263, line 98: We should have a struct for CPEDetails, not string.

rahul2393: noted, I thought we want to optimize the storage here, so I tried storing as JSON string

knqyf263: In any cases, we marshal VulnerabilityDetail to JSON in BoltDB, so the disk size would be almost the same. If we store `CPEDetails` as string, we need to unmarshal twice.

rahul2393: Ok, pushing the change

PR #119, opened 2021-05-27
Add Arch Linux vulns Updater functionality

santhosh1729:
This commit adds functionality to insert arch linux vulns to db

Data in bolt db after insert:
bolt_db_for_arch_linux_data


pkg/vulnsrc/arch-linux/archlinux.go
```diff
@@ -0,0 +1,113 @@
...
+		return types.SeverityLow
+	case "Medium":
+		return types.SeverityMedium
+	case "High":
+		return types.SeverityHigh
+	case "Critical":
+		return types.SeverityCritical
+	}
+	return types.SeverityUnknown
+}
```
knqyf263, line 113:
```suggestion
func convertSeverity(sev string) types.Severity {
	severity, _ := types.NewSeverity(strings.ToUpper(sev))
	return severity
}
```

knqyf263 (commented):
Looks like we have to handle affected versions.
https://security.archlinux.org/AVG-2037

pkg/vulnsrc/arch-linux/archlinux.go
santhosh1729: PR is updated with above suggested changes i.e storing affected version and using convertSeverity function

knqyf263 (approved):

PR #126, opened 2021-06-10
chore: publish docker and ecr

krol3:
Separate of push steps to publish docker and ecr


knqyf263:
Thanks. The issue was fixed now.

PR #132, opened 2021-07-18
Added support for unfixed/open vulnerabilities in alpine.

tonaim:


knqyf263:
We decided not to support unfixed vulnerabilities of Alpine Linux for now.

PR #135, opened 2021-08-05
fix(python): apply package name conversion

AndreyLevchenko:


pkg/vulnsrc/python/python.go
```diff
@@ -154,3 +156,8 @@ func (ad AdvisoryDB) UnmarshalJSON(data []byte) error {
 	}
 	return nil
 }
+func ToLowerCasePythonPackage(pkg string) string {
+	pkg = strings.ToLower(pkg)
+	pkg = strings.ReplaceAll(pkg, "_", "-")
+	return pkg
+}
```
knqyf263, line 163: Could you add a commend and reference?
https://www.python.org/dev/peps/pep-0426/#name

knqyf263 (approved):

PR #139, opened 2021-08-18
feat(python): add pypa advisory support

AndreyLevchenko:


knqyf263:
We supported PyPA advisories https://github.com/aquasecurity/trivy-db/pull/161

PR #140, opened 2021-08-23
refactor(amazon): remove dependency of vuln-list-update

knqyf263:
trivy-db should not depend on vuln-list-update.


tonaim (approved):
Looks good to me.

PR #141, opened 2021-08-24
chore(ci): use go install

knqyf263:
The behavior of `go get` changed.


knqyf263:
I confirmed this change worked well.
https://github.com/aquasecurity/trivy-db/runs/3411416761?check_suite_focus=true

tonaim (approved):

PR #147, opened 2021-09-02
fix(ruby): remove advisories from OSVDB

afdesk:
OSVDB doesn't allow commercial usage without agreement.
Skip those advisories so that Trivy can be used in commercials.

Fixes #1208 in github.com/aquasecurity/trivy/


knqyf263 (changes requested):
In addition, we have to filter out OSVDB URLs.
https://github.com/rubysec/ruby-advisory-db/blob/871af3b5e0612093f587210d8af92697f3e77a9c/rubies/ruby/CVE-2011-3389.yml#L5

PR #148, opened 2021-09-03
fix(ruby): remove advisories from OSVDB

afdesk:
OSVDB doesn't allow commercial usage without agreement.
Skipping OSVDB* files so that Trivy can be used in commercials.
    
also remove the defunct `osvdb.org` website from URL.

Fixes [aquasecurity/trivy/#1208](https://github.com/aquasecurity/trivy/issues/1208)

### DIFF
--- a/pkg/vulnsrc/bundler/bundler.go
+++ b/pkg/vulnsrc/bundler/bundler.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strings"
 
 	bolt "go.etcd.io/bbolt"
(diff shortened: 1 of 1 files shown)


knqyf263:
Fix https://github.com/aquasecurity/trivy/issues/1208

knqyf263 (approved):

knqyf263:
@afdesk Also, could you update this document? We no longer depend on OSVDB and can use "Ruby Advisory Database" for commercial usage.
https://github.com/aquasecurity/trivy/blob/main/docs/vulnerability/detection/data-source.md

PR #166, opened 2021-12-21
chore: bump up Go to 1.17

knqyf263:
The job is failing now due to the old Go version in Dockerfile.
https://github.com/aquasecurity/trivy-db/runs/4589164324?check_suite_focus=true


krol3 (approved):
I tested running this branch with the publish-gh . 
https://github.com/aquasecurity/trivy-db/runs/4600089983?check_suite_focus=true

PR #180, opened 2022-01-18
fix(osv/python): skip GHSA-IDs

DmitriyLewen:
## Description
Added code to skip GHSA-IDs when OSV database is processed

## Related issues
- aquasecurity/trivy-db/issues/179


pkg/vulnsrc/osv/osv_test.go
```diff
@@ -108,3 +109,76 @@ func TestVulnSrc_Update(t *testing.T) {
...
+
+func TestVulnSrc_PythonGHSAIDsIsSkipped(t *testing.T) {
+	tests := []struct {
+		name          string
+		dir           string
+		pythonBuckets []wantKV
+		rustBuckets   []wantKV
+	}{
+		{
+			name: "happy path, when python GHSA-ID is skipped",
```
knqyf263, line 121: I think we can merge this case to TestVulnSrc_Update.

pkg/vulnsrc/osv/osv.go
```diff
@@ -82,6 +82,11 @@ func (vs VulnSrc) Update(dir string) error {
 func (vs VulnSrc) save(eco ecosystem, entries []Entry) error {
 	err := vs.dbc.BatchUpdate(func(tx *bolt.Tx) error {
 		for _, entry := range entries {
+			//GHSA-IDs are already stored via ghsa package.
```
knqyf263, line 85: It should be done when decoding to JSON.

knqyf263 (changes requested):

pkg/vulnsrc/osv/osv.go
DmitriyLewen: Moved

pkg/vulnsrc/osv/osv_test.go
DmitriyLewen: Done

knqyf263 (approved):

PR #183, opened 2022-01-25
feat(vulnsrc): add data sources

knqyf263:
## Description
Add a new bucket "data-source" to store data source names and URLs. They will be shown in the Trivy result.


pkg/db/db.go
```diff
@@ -141,12 +148,18 @@ func (dbc Config) forEach(rootBucket, nestedBucket string) (value map[string][]b
 				continue
 			}
 
+			source, _ := dbc.getDataSource(tx, r)
```
afdesk, line 151: It seems the error will be skipped. maybe we'll print it to logs?

afdesk (approved):

pkg/db/db.go
knqyf263: Added. https://github.com/aquasecurity/trivy-db/pull/183/commits/17ba4e367997fdea18270948f8b996c028a58d38

afdesk (approved):

pkg/db/db.go
afdesk: approved. that's great!

PR #213, opened 2022-05-21
fix(mariner): ignore "Not Applicable" in the Patchable.

masahiro331:
Fix bug.

https://github.com/aquasecurity/trivy-db/runs/6535136541?check_suite_focus=true

# Description 
Mariner OVAL's Definition.MetadataPatchable contains a "Not Applicable" value in addition to the bool value, so it is changed to ignore it.
This modification is a workaround.


pkg/vulnsrc/mariner/oval/oval.go
```diff
@@ -21,6 +22,11 @@ func ParseDefinitions(dir string) ([]Definition, error) {
 	err := utils.FileWalk(dir, func(r io.Reader, path string) error {
 		var def Definition
 		if err := json.NewDecoder(r).Decode(&def); err != nil {
+			// Definition.Metadata.Patchable has a bool and "Not Applicable".
+			// Workaround: Ignore "Not Applicable"
+			if strings.HasSuffix(err.Error(), "unmarshal \"Not Applicable\" into bool") {
```
masahiro331, line 27: https://github.com/golang/go/blob/go1.18/src/encoding/json/decode.go#L919
Also this error.

I use HasSuffix, is there a better way?

pkg/vulnsrc/mariner/mariner.go
```diff
@@ -206,8 +206,11 @@ func (vs VulnSrc) commit(tx *bolt.Tx, platformName string, entries []Entry) erro
 	for _, entry := range entries {
 		cveID := entry.Metadata.Reference.RefID
 		advisory := types.Advisory{}
-		if entry.Metadata.Patchable {
+		// Definition.Metadata.Patchable has a bool and "Not Applicable" string.
+		if entry.Metadata.Patchable == "true" {
```
knqyf263, line 210: We may want to care about the upper letters.

```suggestion
		if strings.ToLower(entry.Metadata.Patchable) == "true" {
```

knqyf263 (approved):

PR #219, opened 2022-06-02
fix(redhat-oval): changed tag for Entry.Arches

DmitriyLewen:
## Description
Json tag for `Entry.Arches` has been renamed from `Arch` to `Arches`

### DIFF
--- a/pkg/vulnsrc/redhat-oval/types.go
+++ b/pkg/vulnsrc/redhat-oval/types.go
@@ -160,7 +160,7 @@ type Definition struct {
 type Entry struct {
 	FixedVersion string `json:",omitempty"`
 	Cves         []CveEntry
-	Arches       []string `json:"Arch,omitempty"`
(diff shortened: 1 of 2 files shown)


pkg/vulnsrc/redhat-oval/types.go
```diff
@@ -160,7 +160,7 @@ type Definition struct {
 type Entry struct {
 	FixedVersion string `json:",omitempty"`
 	Cves         []CveEntry
-	Arches       []string `json:"Arch,omitempty"`
+	Arches       []string `json:"Arches,omitempty"`
```
knqyf263, line 163: You can remove `Arches`, then Go uses the same name as the field name.

```suggestion
	Arches       []string `json:",omitempty"`
```

DmitriyLewen: looks good, thanks!

knqyf263 (approved):

PR #221, opened 2022-06-06
fix(oracle-oval): Support multiple ELSAs per CVE

### LINKED ISSUE #220: DB does not contain all advisories for Oracle Linux
Oracle sometimes issues multiple ELSAs for a single CVE.  One may be for the "normal flavor", and one for the "FIPS flavor".

See for example [ELSA-2021-4451](https://linux.oracle.com/errata/ELSA-2021-4451.html) and [ELSA-2022-9221](https://linux.oracle.com/errata/ELSA-2022-9221.html)

Both these ELSAs address [CVE-2021-20231](https://linux.oracle.com/cve/CVE-2021-20231.html) for `gnutls`, but note that ELSA-2022-9221 contains fix versions for the FIPS flavor (`gnutls-3.6.16-4.0.1.el8_fips`), while ELSA-2021-4451 is for the normal flavor (`gnutls-3.6.16-4.el8`).

It appears that the database only supports one advisory per vulnerability ID.  When trivy-db populates the database advisories via `PutAdvisoryDetail()` (with the CVE ID as the key), this will result in a "last one wins" situation.  In this specific case, the normal advisory is processed first, and then the FIPS advisory, meaning the database only contains the advisory for the FIPS version.

I believe I have confirmed this through a simple `bbolt` traversal of the cached trivy database -
```
b := tx.Bucket([]byte("Oracle Linux 8")).Bucket([]byte("gnutls"))
b.ForEach(func(k, v []byte) error {
    fmt.Printf("%s - %s\n", string(k), v)
    return nil
})
```
Gives the output:
```
CVE-2019-3829 - {"FixedVersion":"3.6.8-8.el8"}
CVE-2019-3836 - {"FixedVersion":"3.6.8-8.el8"}
CVE-2020-11501 - {"FixedVersion":"3.6.8-10.el8_2"}
CVE-2020-13777 - {"FixedVersion":"3.6.8-11.el8_2"}
CVE-2020-24659 - {"FixedVersion":"3.6.14-7.el8_3"}
CVE-2021-20231 - {"FixedVersion":"10:3.6.16-4.0.1.el8_fips"}
CVE-2021-20232 - {"FixedVersion":"10:3.6.16-4.0.1.el8_fips"}
CVE-2021-20305 - {"FixedVersion":"3.6.14-8.el8_3"}
CVE-2021-3580 - {"FixedVersion":"10:3.6.16-4.0.1.el8_fips"}
```

This may also be a problem with `ksplice` flavors as well.

The trivy scanner has related issues differentiating between the flavors (see https://github.com/aquasecurity/trivy/issues/1967 most recently), but if the database only contains 1 of the 1-3 advisories, then addressing the flavor parsing in the scanner will still result in false negatives as the other flavors are not present in the database.

bpfoster:
This is notably important for when there are additional FIPS and ksplice package patches available.

Adjust persisted advisory details object to be more like the RedHat object - each CVE key contains an array of `Entries` - each entry contains the fix version and associated ELSAs for different package flavors.
We keep the `FixedVersion` field at the top of the object to preserve backwards compatibility with older clients.

Since ELSA title and description can be different between the two, do not specify those in the vulnerability details, failing back to the defaults.

Inspecting the database for something like `Oracle Linux 8 / gnutls`, you can see it now changes from:
```
CVE-2019-3829 - {
  "FixedVersion":"3.6.8-8.el8"
}
CVE-2019-3836 - {
  "FixedVersion":"3.6.8-8.el8"
}
CVE-2020-11501 - {
  "FixedVersion":"3.6.8-10.el8_2"
}
CVE-2020-13777 - {
  "FixedVersion":"3.6.8-11.el8_2"
}
CVE-2020-24659 - {
  "FixedVersion":"3.6.14-7.el8_3"
}
CVE-2021-20231 - {
  "FixedVersion":"10:3.6.16-4.0.1.el8_fips"     // incorrect - this results in reported vulnerabilities on a patched normal package
}
CVE-2021-20232 - {
  "FixedVersion":"10:3.6.16-4.0.1.el8_fips"     // incorrect - this results in reported vulnerabilities on a patched normal package
}
CVE-2021-20305 - {
  "FixedVersion":"3.6.14-8.el8_3"
}
CVE-2021-3580 - {
  "FixedVersion":"10:3.6.16-4.0.1.el8_fips"     // incorrect - this results in reported vulnerabilities on a patched normal package
}
```
to:
```
CVE-2019-3829 - {
  "FixedVersion":"3.6.8-8.el8",
  "Entries": [
    {
      "FixedVersion": "3.6.8-8.el8",
      "VendorIDs": [
        "ELSA-2019-3600"
      ]
    }
  ]
}
CVE-2019-3836 - {
  "FixedVersion":"3.6.8-8.el8",
  "Entries": [
    {
      "FixedVersion": "3.6.8-8.el8",
      "VendorIDs": [
        "ELSA-2019-3600"
      ]
    }
  ]
}
CVE-2020-11501 - {
  "FixedVersion":"3.6.8-10.el8_2",
  "Entries": [
    {
      "FixedVersion": "3.6.8-10.el8_2",
      "VendorIDs": [
        "ELSA-2020-1998"
      ]
    }
  ]
}
CVE-2020-13777 - {
  "FixedVersion":"3.6.8-11.el8_2",
  "Entries": [
    {
      "FixedVersion": "3.6.8-11.el8_2",
      "VendorIDs": [
        "ELSA-2020-2637"
      ]
    }
  ]
}
CVE-2020-24659 - {
  "FixedVersion":"3.6.14-7.el8_3",
  "Entries": [
    {
      "FixedVersion": "3.6.14-7.el8_3",
      "VendorIDs": [
        "ELSA-2020-5483"
      ]
    }
  ]
}
CVE-2021-20231 - {
  "FixedVersion":"3.6.16-4.el8",
  "Entries": [
    {
      "FixedVersion": "3.6.16-4.el8",
      "VendorIDs": [
        "ELSA-2021-4451"
      ]
    },
    {
      "FixedVersion": "10:3.6.16-4.0.1.el8_fips",
      "VendorIDs": [
        "ELSA-2022-9221"
      ]
    }
  ]
}
CVE-2021-20232 - {
  "FixedVersion":"3.6.16-4.el8",
  "Entries": [
    {
      "FixedVersion": "3.6.16-4.el8",
      "VendorIDs": [
        "ELSA-2021-4451"
      ]
    },
    {
      "FixedVersion": "10:3.6.16-4.0.1.el8_fips",
      "VendorIDs": [
        "ELSA-2022-9221"
      ]
    }
  ]
}
CVE-2021-20305 - {
  "FixedVersion":"3.6.14-8.el8_3",
  "Entries": [
    {
      "FixedVersion": "3.6.14-8.el8_3",
      "VendorIDs": [
        "ELSA-2021-1206"
      ]
    }
  ]
}
CVE-2021-3580 - {
  "FixedVersion":"3.6.16-4.el8",
  "Entries": [
    {
      "FixedVersion": "3.6.16-4.el8",
      "VendorIDs": [
        "ELSA-2021-4451"
      ]
    },
    {
      "FixedVersion": "10:3.6.16-4.0.1.el8_fips",
      "VendorIDs": [
        "ELSA-2022-9221"
      ]
    }
  ]
}
```

See that previously, CVEs CVE-2021-20231, CVE-2021-20232 and CVE-2021-3580 only had fix versions for the FIPS package.  We now have fix version for both the FIPS package and the normal one.

This also now supports multiple ELSAs per CVE, rolling up to the latest fixed version.  For example:
```
Oracle Linux 8 / kernel-uek
CVE-2020-28374 - {
  "FixedVersion": "5.4.17-2102.202.5.el8uek",
  "Entries": [
    {
      "FixedVersion": "5.4.17-2102.202.5.el8uek",
      "VendorIDs": [
        "ELSA-2021-9006",
        "ELSA-2021-9037",
        "ELSA-2021-9306"
      ]
    }
  ]
}
```

Much like FIPS, Ksplice handling is also improved and accounted for:
```
Oracle Linux 8 / glibc
CVE-2021-3326 - {
  "FixedVersion": "2.28-151.0.1.el8"
  "Entries": [
    {
      "FixedVersion": "2.28-151.0.1.el8",
      "VendorIDs": [
        "ELSA-2021-1585"
      ]
    },
    {
      "FixedVersion": "2:2.28-151.0.1.ksplice2.el8",
      "VendorIDs": [
        "ELSA-2021-9280",
        "ELSA-2021-9344"
      ]
    }
  ]
}
```

fixes #220


DmitriyLewen:
Hello @bpfoster 
I am working with this problem

What do you think about ELSAs with different fixed versions for one CVE?
e.g.:
[CVE-2021-23133](https://linux.oracle.com/cve/CVE-2021-23133.html):
```
Oracle Linux version 8 (kernel-uek)  ELSA-2021-9306  2021-06-15
Oracle Linux version 8 (kernel-uek)  ELSA-2021-9362  2021-07-16
```

from [ELSA-2021-9306](https://linux.oracle.com/errata/ELSA-2021-9306.html):
```
Oracle Linux 8 (aarch64)kernel-uek-5.4.17-2102.202.5.el8uek.src.rpm
```
from [ELSA-2021-9362](https://linux.oracle.com/errata/ELSA-2021-9362.html):
```
Oracle Linux 8 (aarch64)kernel-uek-5.4.17-2102.203.5.el8uek.src.rpm
```

pkg/vulnsrc/oracle-oval/oracle-oval.go
```diff
@@ -84,6 +84,13 @@ func (vs VulnSrc) save(ovals []OracleOVAL) error {
 }
 
 func (vs VulnSrc) commit(tx *bolt.Tx, ovals []OracleOVAL) error {
+	advisories := make(map[string]map[string]map[string]Advisory)
```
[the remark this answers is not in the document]
bpfoster: These are great suggestions

pkg/vulnsrc/oracle-oval/oracle-oval.go
```diff
@@ -110,50 +117,75 @@ func (vs VulnSrc) commit(tx *bolt.Tx, ovals []OracleOVAL) error {
...
+			return xerrors.Errorf("failed to save Oracle Linux OVAL vulnerability: %w", err)
+		}
+	}
 
-			// for optimization
-			if err := vs.dbc.PutVulnerabilityID(tx, vulnID); err != nil {
-				return xerrors.Errorf("failed to save the vulnerability ID: %w", err)
-			}
+	for bkt, advisory := range advisories {
+		if err := vs.dbc.PutAdvisoryDetail(tx, bkt.vulnID, bkt.pkgName, []string{bkt.platform}, advisory); err != nil {
```
bpfoster, line 143: Since this is a schema change, how is it coordinated with trivy, and are there any backwards compatibility concerns?

pkg/vulnsrc/oracle-oval/oracle-oval.go
```diff
@@ -183,16 +215,86 @@ func walkOracle(cri Criteria, osVer string, pkgs []AffectedPackage) []AffectedPa
...
+	}
+
+	return detail
+}
+
+func mergeEntries(advisory Advisory, pkg AffectedPackage, elsaID string) Advisory {
+	affectedFlavor := GetPackageFlavor(pkg.Package.FixedVersion)
+
+	found := false
+	for i, entry := range advisory.Entries {
```
bpfoster, line 242: This loop could be eliminated if we had a nested map here.  `bucket -> flavor -> Entry`.  I explored this but it resulted in random ordering of entries, which caused [tests](https://github.com/aquasecurity/trivy-db/pull/221/files#diff-c3694ef355672019be1ce23f5f5e6d131bfbd8c1c7e6cc93e7b1a9b8a0e6d90bR441) to always fail.  I didn't see another way around it other than sorting prior to saving the advisory details.

pkg/vulnsrc/oracle-oval/oracle-oval.go
```diff
@@ -183,16 +215,86 @@ func walkOracle(cri Criteria, osVer string, pkgs []AffectedPackage) []AffectedPa
...
+		}
+	}
+	return false
+}
+
+// GetPackageFlavor Determine the package "flavor" based on its version string
+//   - normal
+//   - FIPS validated
+//   - ksplice userspace
+func GetPackageFlavor(version string) PackageFlavor {
```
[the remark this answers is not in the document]
DmitriyLewen: I think it is good idea.

pkg/vulnsrc/oracle-oval/oracle-oval.go
DmitriyLewen: I think your current version is good.

pkg/vulnsrc/oracle-oval/oracle-oval.go
DmitriyLewen: It is a good question.
@knqyf263  What do you this about this?
Do we need to create a new version of the Trivy-db schema?

pkg/vulnsrc/oracle-oval/oracle-oval.go
```diff
@@ -183,16 +215,86 @@ func walkOracle(cri Criteria, osVer string, pkgs []AffectedPackage) []AffectedPa
...
+			if version.NewVersion(entry.FixedVersion).LessThan(version.NewVersion(pkg.Package.FixedVersion)) {
+				advisory.Entries[i].FixedVersion = pkg.Package.FixedVersion
+			}
+
+			// Add the ELSA ID to the vendor ID list
+			if !ustrings.InSlice(elsaID, entry.VendorIDs) {
+				advisory.Entries[i].VendorIDs = append(entry.VendorIDs, elsaID)
+			}
+
+			break
```
DmitriyLewen, line 257: we can just return `advisory`.

pkg/vulnsrc/oracle-oval/oracle-oval.go
knqyf263: We need to keep backward compatibility. It is quite a big impact.

bpfoster: Do you have any recommendations on how to do that here?

DmitriyLewen: We will save packages without flavor to: [Advisory.FixedVersion](https://github.com/aquasecurity/trivy-db/blob/39d8a6798e072884afa497c4939036dbafd3b2e4/pkg/types/types.go#L116)
then old Trivy will only support packages without flover.
For new Trivy we will save all versions to [Advisory.PatchedVersions](https://github.com/aquasecurity/trivy-db/blob/39d8a6798e072884afa497c4939036dbafd3b2e4/pkg/types/types.go#L122)

in total:
1) we keep backward compatibility.
2) `Trivy` currently has false positive for packages with some flower versions. After updating `trivy-db`, old Trivy versions will only work with packages without flavor. This work will be correct.
3) New Trivy will work correctly with all flovers.

I'm not sure if this is a good idea, but I don't have any other thoughts.
@bpfoster What do you think about this?

bpfoster: I was hopeful trivy/trivy-db had run into schema changes before and had a solution in hand already :upside_down_face:

@DmitriyLewen That could work.  One downside there that pops out to me (however minor) is we would no longer be able to separate ELSAs by the flavors, since they'd be stored in [Advisory.VendorIDs](https://github.com/aquasecurity/trivy-db/blob/39d8a6798e072884afa497c4939036dbafd3b2e4/pkg/types/types.go#L99).  That is, it would report the ELSAs specific to FIPS/Ksplice packages on the "normal" package, and vice versa.

Playing off your idea, what if instead of reverting the DB object back to [types/types.Advisory](https://github.com/aquasecurity/trivy-db/blob/39d8a6798e072884afa497c4939036dbafd3b2e4/pkg/types/types.go), we adjust the new [oracleoval/types.Advisory](https://github.com/aquasecurity/trivy-db/pull/221/files#diff-c3702d608c4ed2a3af1f11815fb27c57e8556bb2f5d6a3e85d3dbe6ab23a65f6R51-R53):
```
type Advisory struct {
	Entries []Entry `json:",omitempty"`
        // Add this for backwards compat:
        FixedVersion    string `json:",omitempty"`
}
```
Similar to what you said, we populate `FixedVersion` with the normal flavor for backwards compatibility, and `Entries` are ignored by old clients.  The new logic stays the same, and there is an easy path to removing backwards compatibility in the future (drop `FixedVersion` field) while maintaining the ability to separate vendor IDs (and potentially add other future functionality like architecture support).


The other options I can think of are more involved...
- Rev the database version and maintain the 2 schemas/versions for some time.  Not sure the infrastructure is in place to support this already.
- Write the new advisory details structure to another DB bucket.  Same issue that for some (undetermined) time, the same data would be double-stored in the database.

bpfoster: To recap and capture the evolving changes in this PR, I'm now proposing:
Old advisory DB format:
```
Path: Oracle Linux 8 → kernel-uek → CVE-2021-23133
Key: CVE-2021-23133
Value:
{
  "FixedVersion": "5.4.17-2102.203.5.el8uek"
}
```

New advisory DB format:
```
Path: Oracle Linux 8 → kernel-uek → CVE-2021-23133
Key: CVE-2021-23133
Value:
{
  "FixedVersion": "5.4.17-2102.203.5.el8uek"
  "Entries": [
    {
      "FixedVersion": "5.4.17-2102.203.5.el8uek",
      "VendorIDs": [
        "ELSA-2021-9306"
      ]
    },
    {
      "FixedVersion": "5.4.17-2102.203.5.el8uek_fips",
      "VendorIDs": [
        "ELSA-2021-9408"
      ]
    }
  ]
}
```

DmitriyLewen: > One downside there that pops out to me (however minor) is we would no longer be able to separate ELSAs by the flavors, since they'd be stored in [Advisory.VendorIDs](https://github.com/aquasecurity/trivy-db/blob/39d8a6798e072884afa497c4939036dbafd3b2e4/pkg/types/types.go#L99). That is, it would report the ELSAs specific to FIPS/Ksplice packages on the "normal" package, and vice versa.

oh... i missed this. 

> Playing off your idea, what if instead of reverting the DB object back to [types/types.Advisory](https://github.com/aquasecurity/trivy-db/blob/39d8a6798e072884afa497c4939036dbafd3b2e4/pkg/types/types.go), we adjust the new [oracleoval/types.Advisory](https://github.com/aquasecurity/trivy-db/pull/221/files#diff-c3702d608c4ed2a3af1f11815fb27c57e8556bb2f5d6a3e85d3dbe6ab23a65f6R51-R53):

It sounds good. Can you apply these changes?

bpfoster: > Can you apply these changes?

Done

knqyf263:
I'm so sorry to be late.

>I was hopeful trivy/trivy-db had run into schema changes before and had a solution in hand already 🙃

We have it as Trivy DB is now version 2.

>Rev the database version and maintain the 2 schemas/versions for some time. Not sure the infrastructure is in place to support this already.

As you said, we have to maintain multiple versions. We already maintain Trivy DB v1 and v2 and it is already hard. Adding v3 makes it harder. We should avoid it as much as possible.

What if we use ELSA-ID as a key? Did you already discuss it? I'm sorry if I missed it.

```
ELSA-2019-3600 - {
  "FixedVersion":"3.6.8-8.el8",
}
ELSA-2020-1998 - {
  "FixedVersion":"3.6.8-10.el8_2",
}
```

It increases the database size since we cannot reuse the CVE-ID detail in vulnerability-detail. We may want to have CveIDs.

```
ELSA-2019-3600 - {
  "FixedVersion":"3.6.8-8.el8",
  "CveIDs": []string{"CVE-2019-3829", "CVE-2019-3836"}.
}
```

But it doesn't work if there are several fixed versions in the same ELSA-ID.

bpfoster:
> What if we use ELSA-ID as a key? Did you already discuss it? I'm sorry if I missed it.

I did initially take that approach.  There were some complications with it and I switched back to using CVE as the key.  It is most likely feasible to use ELSA ID as the key, but I think we have come to a solution with CVE ID as the key that works for all identified cases and still preserves pretty good backwards compatibility.  Is there something missing there that ELSA ID as the key solves?

I see a couple concerns with the schema you've proposed:
1. Unless I'm overlooking something, backwards compatibility is diminished - because ELSA ID is used as the key and not the CVE ID, older clients utilizing the database will not be able to resolve the CVE and will report results like the following:
```
┌────────────────────────┬────────────────┬──────────┬─────────────────────┬──────────────────────────┬─────────────────────────────────────────────────────┐
│        Library         │ Vulnerability  │ Severity │  Installed Version  │      Fixed Version       │                        Title                        │
├────────────────────────┼────────────────┼──────────┼─────────────────────┼──────────────────────────┼─────────────────────────────────────────────────────┤
│ cyrus-sasl-lib         │ ELSA-2022-0658 │ HIGH     │ 2.1.27-5.el8        │ 2.1.27-6.el8_5           │ https://linux.oracle.com/errata/ELSA-2022-0658.html │
├────────────────────────┼────────────────┤          ├─────────────────────┼──────────────────────────┼─────────────────────────────────────────────────────┤
│ expat                  │ ELSA-2022-0951 │          │ 2.2.5-4.el8         │ 2.2.5-4.el8_5.3          │ https://linux.oracle.com/errata/ELSA-2022-0951.html │
│                        ├────────────────┤          │                     ├──────────────────────────┼─────────────────────────────────────────────────────┤
│                        │ ELSA-2022-9232 │          │                     │ 2.2.5-4.0.1.el8_5.3      │ https://linux.oracle.com/errata/ELSA-2022-9232.html │
├────────────────────────┼────────────────┼──────────┼─────────────────────┼──────────────────────────┼─────────────────────────────────────────────────────┤
```

Keeping CVE ID as the key yields this on an older client:
```
┌────────────────────────┬────────────────┬──────────┬─────────────────────┬──────────────────────┬──────────────────────────────────────────────────────────────┐
│        Library         │ Vulnerability  │ Severity │  Installed Version  │    Fixed Version     │                            Title                             │
├────────────────────────┼────────────────┼──────────┼─────────────────────┼──────────────────────┼──────────────────────────────────────────────────────────────┤
│ cyrus-sasl-lib         │ CVE-2022-24407 │ HIGH     │ 2.1.27-5.el8        │ 2.1.27-6.el8_5       │ cyrus-sasl: failure to properly escape SQL input allows an   │
│                        │                │          │                     │                      │ attacker to execute...                                       │
│                        │                │          │                     │                      │ https://avd.aquasec.com/nvd/cve-2022-24407                   │
├────────────────────────┼────────────────┼──────────┼─────────────────────┼──────────────────────┼──────────────────────────────────────────────────────────────┤
│ expat                  │ CVE-2022-25235 │ CRITICAL │ 2.2.5-4.el8         │ 2.2.5-4.el8_5.3      │ expat: Malformed 2- and 3-byte UTF-8 sequences can lead to   │
│                        │                │          │                     │                      │ arbitrary code...                                            │
│                        │                │          │                     │                      │ https://avd.aquasec.com/nvd/cve-2022-25235                   │
│                        ├────────────────┤          │                     │                      ├──────────────────────────────────────────────────────────────┤
│                        │ CVE-2022-25236 │          │                     │                      │ expat: Namespace-separator characters in "xmlns[:prefix]"    │
│                        │                │          │                     │                      │ attribute values can lead to arbitrary code...               │
│                        │                │          │                     │                      │ https://avd.aquasec.com/nvd/cve-2022-25236                   │
├────────────────────────┼────────────────┼──────────┼─────────────────────┼──────────────────────┼──────────────────────────────────────────────────────────────┤
```

2. It is not as flexible as the `Entries` approach.  Trivy does not currently support it, but once the dust settles on this I would propose adding support for architectures on the entry.  There are cases where an ELSA only applies to 1 architecture ([ELSA-2022-9496](https://linux.oracle.com/errata/ELSA-2022-9496.html) for example), and there are cases where an ELSA has different versions for different architectures ([ELSA-2022-4803](https://linux.oracle.com/errata/ELSA-2022-4803.html) for example).  Introducing an `Arches` field to `Entry` would be a non-breaking change.

All that said, if you have suggestions on how to improve this, I'm happy to make them so we can move this forward.

bpfoster:
@knqyf263 @DmitriyLewen I apologize if my last reply was too strong.  I'd be happy to make some changes here if there's something you'd like done.

bpfoster:
@knqyf263 @DmitriyLewen checking in... anything needed here?  Thanks!

DmitriyLewen:
@knqyf263 if i remember correctly - if we use ELSAs we still need to change db structure. Because ELSA can include multiple versions (e.g. default, fips and ksplice) and we can't save them in [FixedVersion]() field. We need to use a different field for this, ut old Trivy uses this field to get version.
uses this field to get the version.
Then we need to store the versions in 2 fields.

@bpfoster i am right?

bpfoster:
@DmitriyLewen I don't think there's anything technically restricting ELSAs from including multiple _flavors_ (default, fips, and ksplice), but it looks like it's Oracle's practice to split them into separate ELSAs.  All the one's I've looked at are specific to the flavor.  

While out of scope of this change, I would like to point out that the ELSAs support multiple _architectures_, and there are cases that a single ELSA has different versions for different architectures.  See [ELSA-2022-4803](https://linux.oracle.com/errata/ELSA-2022-4803.html) for example, which has `rsyslog-8.24.0-57.0.4.el7_9.3` for aarch64 and `rsyslog-8.24.0-57.0.1.el7_9.3` for x86_64.

Trivy does not currently support architectures for Oracle Linux, but does for RedHat.  If we change the key to ELSA ID and leave the database schema as-is with just a `FixedVersion`, you could not support this in the future.  If however we change the schema to what you and I had discussed earlier with a list of `Entries` (similar to the way RedHat is handled), then we could easily add architecture to the entries (also similar to how the RedHat sources are handeled).

If the changes in this PR are merged in and you have interest in it, I would be happy to submit a change request to add architecture support in the future.  If you'd rather fix it all at once, I'd be happy to incorporate that into this PR as well.

From that standpoint, we could use ELSA ID as the key, and use the modified value schema to support this, if there is benefit in using ELSA ID as the key instead of CVE ID.

It is unfortunate that we duplicate the version of the default package in the value, but I'm not sure there's an alternative that both provides the flexibility to solve all these issues and maintains backwards compatibility.

I did have one concern with using ELSA ID as the key, regardless of the structure of the value.
Since we've been emphasizing backwards compatibility - I believe, but could very well be mistaken, that this would provide a degraded report for older clients.  This is not a huge issue, but worth mentioning.

Because the existing oracle vulnsrc does not translate the DB vulnerability ID to CVE ID like the [redhat one does](https://github.com/aquasecurity/trivy-db/blob/main/pkg/vulnsrc/redhat-oval/redhat-oval.go#L276-L282), the report on an older client will show the ELSA ID as the vulnerability ID instead of CVE ID, and won't include the vulnerability information:
```
┌────────────────────────┬────────────────┬──────────┬─────────────────────┬──────────────────────────┬─────────────────────────────────────────────────────┐
│        Library         │ Vulnerability  │ Severity │  Installed Version  │      Fixed Version       │                        Title                        │
├────────────────────────┼────────────────┼──────────┼─────────────────────┼──────────────────────────┼─────────────────────────────────────────────────────┤
│ cyrus-sasl-lib         │ ELSA-2022-0658 │ HIGH     │ 2.1.27-5.el8        │ 2.1.27-6.el8_5           │ https://linux.oracle.com/errata/ELSA-2022-0658.html │
├────────────────────────┼────────────────┤          ├─────────────────────┼──────────────────────────┼─────────────────────────────────────────────────────┤
│ expat                  │ ELSA-2022-0951 │          │ 2.2.5-4.el8         │ 2.2.5-4.el8_5.3          │ https://linux.oracle.com/errata/ELSA-2022-0951.html │
│                        ├────────────────┤          │                     ├──────────────────────────┼─────────────────────────────────────────────────────┤
│                        │ ELSA-2022-9232 │          │                     │ 2.2.5-4.0.1.el8_5.3      │ https://linux.oracle.com/errata/ELSA-2022-9232.html │
├────────────────────────┼────────────────┼──────────┼─────────────────────┼──────────────────────────┼─────────────────────────────────────────────────────┤
``` 


I hope that clears up my earlier comments.  If I'm wrong on any analysis, please feel free to correct me, and I'm happy to make any changes requested.  Looking forward to getting this resolved so I can stop getting false positives on FIPS versions :smile: .

pkg/vulnsrc/oracle-oval/oracle-oval.go
```diff
@@ -156,21 +168,118 @@ func (vs *VulnSrc) commit(tx *bolt.Tx, ovals []OracleOVAL) error {
...
+	return fixedVersion, versions
+}
+
+type PkgFlavor string
+
+const (
+	NormalPackageFlavor   PkgFlavor = "normal"
+	FipsPackageFlavor     PkgFlavor = "fips"
+	Ksplice1PackageFlavor PkgFlavor = "ksplice1"
+	Ksplice2PackageFlavor PkgFlavor = "ksplice2"
```
DmitriyLewen, line 256: We don't use `ksplice1` flavor for `ksplice2` (and vice versa) in Trivy:
https://github.com/aquasecurity/trivy/blob/dd0a64a1cf0cd76e6f81e3ff55fa6ccb95ce3c3d/pkg/detector/ospkg/oracle/oracle.go#L72-L77

So i separated `ksplice1` and `ksplice2`.
After these changes db size increased to `0.9MB`:
```
➜  du -sh ./assets-with-flavors
539M	./assets-with-flavors
➜  du -sh ./assets             
540M	./assets 
```

DmitriyLewen (commented):
research on the choice of structure and dimensions of the various options can be seen here - https://github.com/aquasecurity/trivy-db/pull/452

pkg/vulnsrc/oracle-oval/oracle-oval.go
bpfoster: It's been long enough that I've lost track of all the details on this, but I did want to pass along some information I got from @tvierling (from the Oracle Linux support/development group) when I was working though this.  I won't quote his email directly without his permission, but the gist of it is that `ksplice` is the flavor, and the number after `ksplice` is just a revision / build number (which should be considered.. `x.y.z.ksplice2` > `x.y.z.ksplice1`).  So I don't think you want to separate out as different flavors.

DmitriyLewen: I'm glad you're still here and helping with this task.
Sorry we're working on this after so long...

Also, I want to thank you for your work! Our work with you helped me remember the details and update this PR.

> but the gist of it is that ksplice is the flavor, and the number after ksplice is just a revision / build number (which should be considered.. x.y.z.ksplice2 > x.y.z.ksplice1). So I don't think you want to separate out as different flavors.

I originally used your `PackageFlavor` function. 
But I was confused that we separate ksplice1 and ksplice2 in Trivy and i updated PR.

But since the person from Oracle Linux says that we shouldn't divide them, then there is a mistake in our logic Trivy.

Thanks! I will revert last commit.

DmitriyLewen: done - https://github.com/aquasecurity/trivy-db/pull/221/commits/a135499d61b191053c033a5dc22ef2dd24e244b5

knqyf263:
@santhosh1729 Can you also please review this PR?

knqyf263:
@santhosh1729 If it looks good to you, please approve this PR just for the record that Aqua approves it.

pkg/vulnsrc/oracle-oval/oracle-oval.go
```diff
@@ -192,10 +295,36 @@ func (o *Oracle) Put(tx *bolt.Tx, input PutInput) error {
...
+		return nil, xerrors.Errorf("unable to iterate advisories: %w", err)
 	}
+	var advisories []types.Advisory
+	for vulnID, v := range rawAdvisories {
+		var adv types.Advisories
+		if err = json.Unmarshal(v.Content, &adv); err != nil {
+			return nil, xerrors.Errorf("failed to unmarshal advisory JSON: %w", err)
+		}
+
+		// For backward compatibility
```
knqyf263, line 309: This is needed only when users keep using old databases, right? I don't think people use the old database for a month. We can set a deadline.

```suggestion
		// For backward compatibility (This code can be deleted after Dec 19th, 2024)
```

knqyf263 (commented):
Looks almost good. Left a nit comment

pkg/vulnsrc/oracle-oval/oracle-oval.go
```diff
@@ -156,21 +168,112 @@ func (vs *VulnSrc) commit(tx *bolt.Tx, ovals []OracleOVAL) error {
...
+				ver = savedVer
+			}
+		}
+		fixedVers[flavor] = ver
+	}
+
+	versions := lo.Values(fixedVers)
+	slices.Sort(versions)
+
+	fixedVersion := fixedVers[NormalPackageFlavor]
```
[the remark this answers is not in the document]
DmitriyLewen: This is not entirely correct (which is why this PR was created), but it is only needed for backward compatibility, so let's leave the previous logic
Updated in https://github.com/aquasecurity/trivy-db/pull/221/commits/cb14132c0558fc680363299730dd9638e47bea84

pkg/vulnsrc/oracle-oval/oracle-oval.go
DmitriyLewen: You are right.
Updated.

knqyf263 (approved):

knqyf263:
I'm waiting for approval from @santhosh1729.

santhosh1729:
@knqyf263 changes look good to me, It resolves the the issue that we are experiencing

knqyf263:
@DmitriyLewen Did you confirm the new DB works with the current Trivy?

DmitriyLewen:
Trivy v0.57.1 correctrly works with new DB:
```bash
➜  7954 trivy -q image oraclelinux:9  

oraclelinux:9 (oracle 9.4)

Total: 20 (UNKNOWN: 0, LOW: 13, MEDIUM: 7, HIGH: 0, CRITICAL: 0)

┌───────────────────────┬────────────────┬──────────┬────────┬─────────────────────────┬───────────────────────┬───────────────────────────────────────────────────────────┐
│        Library        │ Vulnerability  │ Severity │ Status │    Installed Version    │     Fixed Version     │                           Title                           │
├───────────────────────┼────────────────┼──────────┼────────┼─────────────────────────┼───────────────────────┼───────────────────────────────────────────────────────────┤
│ expat                 │ CVE-2024-50602 │ MEDIUM   │ fixed  │ 2.5.0-2.el9_4.1         │ 2.5.0-3.el9_5.1       │ libexpat: expat: DoS via XML_ResumeParser                 │
│                       │                │          │        │                         │                       │ https://avd.aquasec.com/nvd/cve-2024-50602                │
├───────────────────────┼────────────────┤          │        ├─────────────────────────┼───────────────────────┼───────────────────────────────────────────────────────────┤
```
изображение

knqyf263:
Thanks! Merging.

DmitriyLewen:
@knqyf263  i updated https://github.com/aquasecurity/trivy/pull/7858

PR #223, opened 2022-06-23
fix(go): no lowercase for go packages

afdesk:
There are Go packages with names that contain letters in different cases, eg: `github.com/Masterminds/vcs`.
we shouldn't normalize them.

### DIFF
--- a/pkg/vulnsrc/vulnerability/vulnerability.go
+++ b/pkg/vulnsrc/vulnerability/vulnerability.go
@@ -210,9 +210,8 @@ func NormalizePkgName(ecosystem types.Ecosystem, pkgName string) string {
 		// and MUST consider hyphens and underscores to be equivalent.
 		pkgName = strings.ToLower(pkgName)
 		pkgName = strings.ReplaceAll(pkgName, "_", "-")
-	}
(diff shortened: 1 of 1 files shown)


pkg/vulnsrc/vulnerability/vulnerability.go
```diff
@@ -210,7 +210,9 @@ func NormalizePkgName(ecosystem types.Ecosystem, pkgName string) string {
 		// and MUST consider hyphens and underscores to be equivalent.
 		pkgName = strings.ToLower(pkgName)
 		pkgName = strings.ReplaceAll(pkgName, "_", "-")
-	} else if ecosystem != NuGet { // Nuget is case-sensitive
+	}
+	// Nuget is case-sensitive, Go has case-sensitive packages
+	if ecosystem != NuGet && ecosystem != Go {
```
knqyf263, line 215: Pip also gets into this if-statement. It should be `else if`.

afdesk: oh, sure. my mistake.

knqyf263 (approved):

PR #228, opened 2022-07-14
ci: rollback `oras` to v0.12.0

afdesk:
`oras` was rollbacked to v0.12.0, because now v0.13.0 (the latest version) contains bugs.

### DIFF
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -51,6 +51,7 @@ jobs:
 
       - name: Upload assets to GHCR
         run: |
+          # oras was rollbacked to v0.12.0, because now v0.13.0 (the latest version) contains bugs
           curl -LO https://github.com/oras-project/oras/releases/download/v0.12.0/oras_0.12.0_linux_amd64.tar.gz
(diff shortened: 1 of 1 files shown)


.github/workflows/cron.yml
```diff
@@ -51,9 +51,13 @@ jobs:
 
       - name: Upload assets to GHCR
         run: |
+          # oras was rollbacked to v0.12.0, because now v0.13.0 (the latest version) contains bugs
```
knqyf263, line 54: We should have a GitHub link here.

.github/workflows/cron.yml
```diff
@@ -51,9 +51,13 @@ jobs:
...
+          curl -LO https://github.com/oras-project/oras/releases/download/v0.12.0/oras_0.12.0_linux_amd64.tar.gz
+          tar -xvf ./oras_0.12.0_linux_amd64.tar.gz
```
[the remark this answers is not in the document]
afdesk: done

.github/workflows/cron.yml
afdesk: > We should have a GitHub link here.

i've added a link to my issue, right?

PR #233, opened 2022-07-23
feat: add k8s outdated api

### LINKED ISSUE #231: Add k8s outdated api data source
it is require to add k8s [outdated api data](https://github.com/aquasecurity/trivy-db-data/) source to trivy-db for `trivy` and `trivy-operator` to implement k8s outdated api resources view

chen-keinan:
Signed-off-by: chenk <hen.keinan@gmail.com>
Add support to outdated api

Close #231


chen-keinan:
@knqyf263 Pls take a look when you have time

pkg/db/k8sdetail.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"encoding/json"
+	"github.com/aquasecurity/trivy-db/pkg/types"
+	bolt "go.etcd.io/bbolt"
+	"golang.org/x/xerrors"
+)
+
+const k8sBucket = "k8s"
+
+func (dbc Config) AddK8sOutdatedAPI(tx *bolt.Tx, key string, apis interface{}) (err error) {
```
DmitriyLewen, line 12: Looks like there should be `PutK8sOutdatedAPI`. 
@chen-keinan wdyt?

pkg/k8s/api/outdated.go
```diff
@@ -0,0 +1,93 @@
...
+	return nil
+}
+
+func (vs Outdated) save(advisories []Advisory) error {
+	err := vs.dbc.BatchUpdate(func(tx *bolt.Tx) error {
+		if err := vs.dbc.PutK8sDataSource(tx, dataSource, source); err != nil {
+			return err
+		}
+		for _, adv := range advisories {
+			if err := vs.dbc.AddK8sOutdatedAPI(tx, dataType, adv); err != nil {
```
DmitriyLewen, line 75: 1 small question: we save all advisories in 1 bucket. We don't need to get one or more(not all) `OutdatedAPI`?

pkg/k8s/api/outdated.go
```diff
@@ -0,0 +1,93 @@
...
+	if err = vs.save(advisories); err != nil {
+		return xerrors.Errorf("error in outdated api save: %w", err)
+	}
+
+	return nil
+}
+
+func (vs Outdated) save(advisories []Advisory) error {
+	err := vs.dbc.BatchUpdate(func(tx *bolt.Tx) error {
+		if err := vs.dbc.PutK8sDataSource(tx, dataSource, source); err != nil {
```
DmitriyLewen, line 71: i am not sure, but, perhaps, we need to save `DataSourse` with all `Datasourses`?
I was confused when i didn't find `DataSourse` in default bucket.

chen-keinan: we are trying to separate vulnerabilities sources from other sources , as we might have different sources type in the future

pkg/k8s/api/outdated.go
chen-keinan: yes we want to save all its a small db which we take and embed in rego policy , so we need it in one peace :)

chen-keinan:
btw , I need to do a small change to data outdated api structure as well , so I'll ping you when I'm done

pkg/k8s/api/outdated.go
DmitriyLewen: Okay. It makes sense.
But maybe we need to separate `data sources` and `databases`. When we add several new sources, confusion can arise because `data sources` and `databases` will be saved in 1 bucket.
e.g.
```yaml
k8s:
    - outdated-api
    - outdatedapi-data-source
    - new-db
    - new-db-data-source
    - new-db-2
    - new-db-2-data-source
    ...
```
What if we create 2 buckets inside `k8s` for `data sources` and `db's`
like this:
```yaml
k8s:
    - db(need to choose the best name)
        - outdated-api
        - new-db
        - new-db-2
    - data-sources
        - outdatedapi-data-source
        - new-db-data-source
        - new-db-2-data-source
    ...
```

chen-keinan:
@DmitriyLewen all comments has been addressed

pkg/db/db.go
```diff
@@ -53,6 +53,10 @@ type Operation interface {
 	PutRedHatCPEs(tx *bolt.Tx, cpeIndex int, cpe string) (err error)
 	RedHatRepoToCPEs(repository string) (cpeIndices []int, err error)
 	RedHatNVRToCPEs(nvr string) (cpeIndices []int, err error)
+	//for K8s
+	GetK8sOutdatedAPI(key string) (outdatedapi types.OutDatedAPIData, err error)
+	PutK8sOutdatedAPI(tx *bolt.Tx, key string, apis interface{}) (err error)
```
DmitriyLewen, line 58:
```suggestion
	GetK8sDb(key string) (outdatedapi types.OutDatedAPIData, err error)
	PutK8sDb(tx *bolt.Tx, key string, apis interface{}) (err error)
```
What if we create "universal" name for these functions?
In the future, we will want to use these functions for other k8s databases.

pkg/k8s/api/outdated.go
```diff
@@ -0,0 +1,93 @@
...
```
DmitriyLewen, line 71:
```suggestion
		if err := vs.dbc.PutK8sDataSource(tx, dataType, source); err != nil {
```
Vulnerabilities use same bucket name for `datasources` and `DB's`.
I think we need to do same.

pkg/k8s/api/outdated.go
```diff
@@ -0,0 +1,93 @@
...
+type Advisory types.OutDatedAPIData
+
+var (
+	dataSource = "outdatedapi-data-source"
+	dataType   = "outdated-api"
+
+	source = types.DataSource{
+		ID:   K8sOutdatedAPI,
+		Name: "Kubernetes GitHub docs",
+		URL:  "https://github/kubernetes/kubernetes",
```
DmitriyLewen, line 29:
```suggestion
		URL:  "https://github.com/kubernetes/kubernetes",
```

DmitriyLewen (commented):
@chen-keinan I left some small notes. Take a look please.

The rest looks good!

chen-keinan:
@DmitriyLewen comments addressed

pkg/db/k8sdetail.go
```diff
@@ -0,0 +1,67 @@
...
+		if err := json.Unmarshal(value, &outdatedapi); err != nil {
+			return xerrors.Errorf("failed to unmarshal JSON: %w", err)
+		}
+		return nil
+	})
+	if err != nil {
+		return nil, xerrors.Errorf("failed to get the outdated-api %q: %w", key, err)
+	}
+	return outdatedapi, nil
+}
```
DmitriyLewen, line 67: can we use `outdatedapi`([types.OutDatedAPIData](https://github.com/aquasecurity/trivy-db/pull/233/files#diff-9aa20f1d30a7bc9bd5b07b0501cac0fe205afcb99c036c53d6a3bcf690ec4325R3)) for new databases?
If no - perhaps we need to return `value`(without unmarshal) and parse this into `Get() functions:
https://github.com/aquasecurity/trivy-db/blob/cf354a526ea44ec03e7a514249992a76ed76d792/pkg/k8s/api/outdated.go#L87-L96
if yes - I think we should choose a more "universal" name

chen-keinan: Not sure this one should be generic , it meant for outdated api datatype
if another k8s data will be added , we will have to add another file same as we do for vulns type we do not keep generic type for all.

DmitriyLewen: I am not sure that understood you correctly.
```
		if err := json.Unmarshal(value, &outdatedapi); err != nil {
			return xerrors.Errorf("failed to unmarshal JSON: %w", err)
		}
```
Will this work correctly with new k8s data?

chen-keinan: Yes ,as this file related to outdated api only, if a new k8d db will be added a new file should be introduced

DmitriyLewen: > as this file related to outdated api only

Then we need to move [this part](https://github.com/aquasecurity/trivy-db/blob/cf354a526ea44ec03e7a514249992a76ed76d792/pkg/db/k8sdetail.go#L58-L60) to [Get()](https://github.com/aquasecurity/trivy-db/blob/cf354a526ea44ec03e7a514249992a76ed76d792/pkg/k8s/api/outdated.go#L87-L96), because we will get error, when we try to unmarshal new data file(from `trivy-db-data`) with different structure.

chen-keinan: I see now what you mean I was redering to somthing else. Sorry for the confusion, ok I will dix it

chen-keinan:
@DmitriyLewen comments has been addressed

DmitriyLewen (approved):
Thank a lot for your work!
LGTM. Approved.

pkg/k8s/api/outdated.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/aquasecurity/trivy-db/pkg/types"
+	"github.com/aquasecurity/trivy-db/pkg/utils"
+)
+
+const (
+	k8sOutdatedAPiDir                = "api"
+	K8sOutdatedAPI    types.SourceID = "outdated-api"
+)
+
+type Advisory types.OutDatedAPIData
```
knqyf263, line 20: Is this an advisory?

Makefile
```diff
@@ -86,3 +86,8 @@ db-clean:
 db-fetch-vuln-list:
 	mkdir -p cache/vuln-list
 	wget -qO - https://github.com/aquasecurity/vuln-list/archive/main.tar.gz | tar xz -C cache/vuln-list --strip-components=1
+
+.PHONY: trivy-db-data
```
knqyf263, line 90: We probably should add a comment about the reason why we have vuln-list and trivy-db-data.

knqyf263 (commented):
I'm sorry for the late review, but merging PRs in trivy-db has a big impact because it is built and distributed to all Trivy users immediately. Keeping backward compatibility is critical in trivy-db and the first design is essential. I should have reviewed it before merging PR.

Fortunately, I didn't find a big issue, but I left some small comments. Could you take a look at them? @chen-keinan

pkg/k8s/api/outdated.go
chen-keinan: its a document dB not security , happy to change it

pkg/db/k8sdetail.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/aquasecurity/trivy-db/pkg/types"
+	bolt "go.etcd.io/bbolt"
+	"golang.org/x/xerrors"
+)
+
+const k8sRootBucket = "k8s"
+const DbBucket = "db"
+const DataSourceBucket = "data-source"
+
+func (dbc Config) PutK8sDataSource(tx *bolt.Tx, bktName string, source types.DataSource) (err error) {
```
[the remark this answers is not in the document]
chen-keinan: true it was prep for future , will remove it for now

PR #236, opened 2022-08-02
fix(redhat): add rejected vulnerability check for rhsa-xxx

DmitriyLewen:
## Description
 `RHSA-xxx` advisories don't checks to `reject`  when do optimizing because `NVD` only contains `CVE-xxx`.
Added check for RHSAs

## Related issues
- Close aquasecurity/trivy/issues/2623

### DIFF
--- a/pkg/vulnsrc/redhat-oval/redhat-oval.go
+++ b/pkg/vulnsrc/redhat-oval/redhat-oval.go
@@ -85,7 +85,7 @@ func (vs VulnSrc) Update(dir string) error {
 				continue
 			}
 
-			definitions, err := parseOVALStream(filepath.Join(versionDir, f.Name()), uniqCPEs)
+			definitions, err := vs.parseOVALStream(filepath.Join(versionDir, f.Name()), uniqCPEs)
 			if err != nil {
 				return xerrors.Errorf("failed to parse OVAL stream: %w", err)
 			}
@@ -288,7 +288,7 @@ func (vs VulnSrc) Get(pkgName string, repositories, nvrs []string) ([]types.Advi
 	return advisories, nil
 }
 
-func parseOVALStream(dir string, uniqCPEs CPEMap) (map[bucket]Definition, error) {
+func (vs VulnSrc) parseOVALStream(dir string, uniqCPEs CPEMap) (map[bucket]Definition, error) {
 	log.Printf("    Parsing %s", dir)
 
 	// Parse tests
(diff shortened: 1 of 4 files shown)


pkg/vulnsrc/redhat-oval/redhat-oval.go
```diff
@@ -342,6 +342,15 @@ func parseDefinitions(advisories []redhatOVAL, tests map[string]rpmInfoTest, uni
 
 			var cveEntries []CveEntry
 			for _, cve := range advisory.Metadata.Advisory.Cves {
+				// get details from NVD
+				details, err := vs.dbc.GetVulnerabilityDetail(cve.CveID)
```
knqyf263, line 346: VulnSrc should be order-agnostic. It means NVD must be inserted before Red Hat OVAL. I'd suggest adding CveIDs here.
https://github.com/aquasecurity/trivy-db/blob/2bd1364579ec652f8f595c4a61595fd9575e8496/pkg/types/types.go

And check the rejected status of the related CVE-IDs as well here.
https://github.com/aquasecurity/trivy-db/blob/930461748b6307590f4718daeab37cec51fe78b3/pkg/vulnsrc/vulnerability/vulnerability.go#L178-L189

pkg/vulnsrc/redhat-oval/types/types.go
```diff
@@ -1,65 +1,65 @@
-package redhatoval
+package types
```
knqyf263, line 1: Could you tell me the reason why we want to move this package?

pkg/db/advisory_detail.go
```diff
@@ -77,3 +79,65 @@ func (dbc Config) saveAdvisories(tx *bolt.Tx, bkt *bolt.Bucket, bktNames []strin
...
+	if redHatBucket == nil {
+		return nil
+	}
+
+	if err := dbc.saveRhsaAdvisories(tx, redHatBucket, []string{"Red Hat"}, vulnID, rejectedCve); err != nil {
+		return xerrors.Errorf("walk advisories error: %w", err)
+	}
+
+	return nil
+}
```
knqyf263, line 105: What is this function for...?

knqyf263 (commented):
This PR is too complicated. We should consider how to simplify it.

pkg/vulnsrc/redhat-oval/types/types.go
DmitriyLewen: we need to decode [Advisory](https://github.com/aquasecurity/trivy-db/blob/2c0339bb908551e6a6eb5fe5b22a4a004ce42333/pkg/vulnsrc/redhat-oval/types.go#L151-L176) structure here: https://github.com/aquasecurity/trivy-db/blob/03e5c844a77f00303a550476af124d753616575e/pkg/db/advisory_detail.go#L119-L128
this is necessary to remove rejected cves from `CveEntry`.
But we can't import `redhatoval` package from `db` package, because we will get `cycle import`.(`redhatoval` imported `db` for `PutAdvisoryDetail` func)

pkg/db/advisory_detail.go
DmitriyLewen: this function finds all `Red Hat` buckets, decodes them, removes rejected cves and saves to db as ` SaveAdvisoryDetails`.

PR #237, opened 2022-08-11
feat: add vendor status

DmitriyLewen:
## Description
Added structure with cve statuses for different vendors.
e.g.:
```
  "VendorStatus": {
    "nvd": "rejected",
    "redhat": "fixed",
    "alpine": "will not fix",
  }
```
There are next statuses:
-  "fixed"
- "affected"
- "will-not-fix"
- "deferred"
- "rejected"

### DIFF
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -9,6 +9,10 @@ type Severity int
 
 type VendorSeverity map[SourceID]Severity
 
+type Status string
+
+type VendorStatus map[SourceID]Status
+
 type CVSS struct {
 	V2Vector string  `json:"V2Vector,omitempty"`
 	V3Vector string  `json:"V3Vector,omitempty"`
@@ -29,6 +33,12 @@ const (
 	SeverityMedium
 	SeverityHigh
 	SeverityCritical
+
+	StatusFixed      Status = "fixed"
+	StatusAffected   Status = "affected"
+	StatusWillNotFix Status = "will-not-fix"
+	StatusDeferred   Status = "deferred"
+	StatusRejected   Status = "rejected"
 )
 
 var (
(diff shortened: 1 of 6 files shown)


pkg/types/types.go
```diff
@@ -77,6 +77,7 @@ type VulnerabilityDetail struct {
 	Description      string     `json:",omitempty"`
 	PublishedDate    *time.Time `json:",omitempty"` // Take from NVD
 	LastModifiedDate *time.Time `json:",omitempty"` // Take from NVD
+	Status           bool       `json:",omitempty"` // Rejected or not Rejected cve. Take from NVD
```
knqyf263, line 80: NVD actually has more statuses like DISPUTED. We currently support "REJECTED" or not, but I want to make it flexible for future enhancement.
https://nvd.nist.gov/general/cve-process

```suggestion
	Status           types.Status       `json:",omitempty"` // Take from NVD
```

pkg/vulnsrc/vulnerability/vulnerability.go
```diff
@@ -38,11 +38,12 @@ func (v Vulnerability) GetDetails(vulnID string) map[types.SourceID]types.Vulner
 	return details
 }
 
-func (Vulnerability) IsRejected(details map[types.SourceID]types.VulnerabilityDetail) bool {
-	return getRejectedStatus(details)
-}
-
 func (Vulnerability) Normalize(details map[types.SourceID]types.VulnerabilityDetail) types.Vulnerability {
+	if details[NVD].Status {
```
knqyf263, line 42:
```suggestion
	if details[NVD].Status == types.StatusRejected {
```

pkg/types/types.go
knqyf263: I mean we should define `type Status string` somewhere.

PR #241, opened 2022-08-25
feat: add Rust ecosystem from GHSA and disable from OSV

afdesk:
## Description
Trivy DB should contain data from GitHub Advisory Database for Rust ecosystem: https://github.com/advisories?query=type%3Areviewed+ecosystem%3Arust

## Related PRs
- [x] https://github.com/aquasecurity/vuln-list-update/pull/174


knqyf263 (commented):
Looks like you added a JSON file, but not the expected value in DB.
https://github.com/aquasecurity/trivy-db/blob/49b1399b9923280a6973c6c2953db0b6dd0e1b95/pkg/vulnsrc/ghsa/ghsa_test.go#L38-L56

knqyf263 (approved):

PR #243, opened 2022-08-28
support k8s outdated api db

chen-keinan:
support k8s outdated api db
Related: #233 231


pkg/k8s/api/outdated.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	return nil
+}
+
+func (vs Outdated) save(advisories []OutdatedApiDb) error {
+	err := vs.dbc.BatchUpdate(func(tx *bolt.Tx) error {
+		for _, adv := range advisories {
+			if err := vs.dbc.PutK8sDb(tx, dataType, adv); err != nil {
+				return xerrors.Errorf("failed to put data source: %w", err)
```
knqyf263, line 72: Is it a data source? Seems like just data about outdated API

pkg/k8s/api/outdated.go
```diff
@@ -0,0 +1,90 @@
...
+type OutdatedApiDb types.OutDatedAPIData
+
+var (
+	dataType = "outdated-api"
+
+	source = types.DataSource{
+		ID:   K8sOutdatedAPI,
+		Name: "Kubernetes GitHub docs",
+		URL:  "https://github.com/kubernetes/kubernetes",
+	}
```
knqyf263, line 29: Do we want to keep it? It is not used.

knqyf263 (commented):
Looks good! I left nit comments.

pkg/k8s/api/outdated.go
knqyf263:
```suggestion
				return xerrors.Errorf("failed to put outdated API data: %w", err)
```

pkg/k8s/api/outdated.go
chen-keinan: I will remove it

PR #251, opened 2022-09-15
Publish tag containing the build date to locate old DB

jdolitsky:
There does not currently appear to be any way to run Trivy using an older version of the DB.

Example use case: check how many vulnerabilities would have been detected by Trivy in a given image 30 days ago.

There is additional work to do in the CLI to change the tag to something other that "2" (for example `--db-tag=2022091512`). However, until these tags start to be published, there is no reliable way to obtain this data other than by scrubbing through the GitHub Actions logs, locating the digest published by oras, and following the air-gap instructions.

### DIFF
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -58,7 +58,7 @@ jobs:
       - name: Upload assets to GHCR
         run: |
           ./oras version
-          tags=(latest ${{ env.VERSION }})
+          tags=(latest ${{ env.VERSION }} $(date +"%Y%m%d%H"))
           for tag in ${tags[@]}; do
             ./oras push ghcr.io/${{ github.repository }}:${tag} \
               --manifest-config /dev/null:application/vnd.aquasec.trivy.config.v1+json \


itaysk:
Actually, `metadata.Metadata.UpdatedAt` is what we were looking for (sorry for the confusion). So what's left to do is just to tag the image with this date. In order to access it it's probably needed to add a cli command to the trivy-db cli that prints it.

jdolitsky:
@itaysk alright, we have implemented this in this PR! Please take a look when you have a moment

```
$ ./trivy-db timestamp
2022-11-17-18-09-07
```

knqyf263 (commented):
We're so sorry to make another suggestion again, but can't we load `metadata.json` rather than adding a new subcommand? It is simpler if it works.

You can read the file between `make db-compact` and `make db-compress` steps.
https://github.com/aquasecurity/trivy-db/blob/be9ad3ad92837eadf93f487d5e750c5f5b47744d/.github/workflows/cron.yml#L36-L40

I'd suggest as below.

```
      - name: Compact DB
        run: make db-compact

      - name: Get the UpdatedAt timestamp
        id: timestamp
        run: |
          echo "timestamp=$(jq .UpdatedAt assets/metadata.json | date "+%F-%H-%M-%S")" >> $GITHUB_OUTPUT

      - name: Compress assets
        run: make db-compress
```

PR #267, opened 2022-12-02
fix(govuln): don't save GHSA-IDs if CVE-IDs exist

DmitriyLewen:
## Details
If aliases contain both CVE-ID and GHSA-ID, we should save only CVE-ID, because GHSA-IDs is duplicates of CVE-IDs in this case.

We also save `go vulnerability details` for `ghsa`.
We will use this data for `govuln`


pkg/vulnsrc/ghsa/ghsa.go
```diff
@@ -136,9 +137,11 @@ func (vs VulnSrc) commit(tx *bolt.Tx, ecosystem types.Ecosystem, entries []Entry
 		}
 
 		pkgName := vulnerability.NormalizePkgName(ecosystem, entry.Package.Name)
-		err = vs.dbc.PutAdvisoryDetail(tx, vulnID, pkgName, []string{bucketName}, a)
-		if err != nil {
-			return xerrors.Errorf("failed to save GHSA: %w", err)
+		if ecosystem != vulnerability.Go { // we only keep vulnerability details for Go
```
knqyf263, line 140: We should mention we use `golang/vulndb` instead of GHSA and the reason why we don't use GHSA.
https://github.com/golang/vulndb

pkg/vulnsrc/ghsa/ghsa.go
```diff
@@ -164,6 +167,21 @@ func (vs VulnSrc) commit(tx *bolt.Tx, ecosystem types.Ecosystem, entries []Entry
...
+			vuln.ID = vulnID
+			if err = vs.dbc.PutVulnerabilityDetail(tx, vulnID, vulnerability.GHSA, vuln); err != nil {
+				return xerrors.Errorf("failed to save GHSA vulnerability detail: %w", err)
+			}
+
+			// for optimization
+			if err = vs.dbc.PutVulnerabilityID(tx, vulnID); err != nil {
+				return xerrors.Errorf("failed to save the vulnerability ID: %w", err)
+			}
+		}
```
knqyf263, line 184: Why do we need this block? If I understand correctly, it satisfies the conditions of the if statement in the case of CVE-ID. Do we need to store GHSA-IDs having CVE-ID? I thought we wanted to store GHSA-IDs without CVE-ID.

pkg/vulnsrc/ghsa/ghsa.go
DmitriyLewen: Done

pkg/vulnsrc/ghsa/ghsa.go
DmitriyLewen: NVD has some delay in updating information about vulnerabilities. 
We are keeping  details for CVE-xxx from ghsa and will use them to get severity.

DmitriyLewen: Done

knqyf263 (approved):

PR #277, opened 2023-02-08
Populate advisory state for RedHat OVALv2 data

jhebden-gl:
This is a fairly simple PR to add support for ingesting the "State" of an advisory for Red Hat, populated from the OVALv2 data which is already being ingested.

If an advisory (RHSA or CVE) defines a resolution state, such as "Out of support scope" or "Will not fix", then this data will be ingested into the "Red Hat" advisory bucket for the existing advisory entry under the "State" key.

I have also added a test to ensure this is properly ingested for CVE-2020-11879, a vulnerability which has no Red Hat errata (no RHSA) as it is designated "Will not fix" by Red Hat.

This data is incredibly useful for folks who are using trivy-db in other projects, and I could imagine it would be useful for adding additional data to Trivy container scans as well - however I have not added any functionality to Trivy at this stage to make sure of this field.


jhebden-gl:
Hello @knqyf263 - would you kindly be able to look at this? It is a very small change, I do not believe it will impact anything downstream. I'd really like to avoid maintaining a fork of the trivy-db repository if possible to be able to reference the Red Hat errata status. Thank you!

jhebden-gl:
I've just added one other small fix to the mergeAdvisories function to make sure that advisories for packages with different fix status do not get merged together, the existing behaviour was clobbering fix states. I've updated the merge functionality to treat state like fixVersion when it comes to merging, and this fixed a few edge cases parsing advisories such as CVE-2021-35937 where the packages have different fix states by package name and affected CPE.

adamcohen2:
Thanks for creating this @jhebden-gl, it would be great to have access to the state data!

@knqyf263 any chance you can take a look at this please? Thanks!

thiago-gitlab:
Hi, @krol3. 👋🏽 

Would you be able to help us find someone to review this PR? The author is a colleague of mine at GitLab.

jhebden-gl:
Hello @itaysk 👋🏻 Apologies for the direct ping, I was wondering if you could help to find someone who might be able to review this MR? It should be a relatively simple change & I have added tests to cover the modifications.

knqyf263 (approved):
We're sorry to be late. It looks good, but I'm feeling like `string` is too flexible for `State`. What if defining `type State string` or `type State int` like enum? We can merge this PR and start this implementation in another PR.

jhebden-gl:
> We're sorry to be late. It looks good, but I'm feeling like `string` is too flexible for `State`. What if defining `type State string` or `type State int` like enum? We can merge this PR and start this implementation in another PR.

No problem, thank you for reviewing!

I do agree setting known/acceptable values for `State` would be a sensible iteration to make on the implementation - `State` however was already present in the schema and I think it would be good to approach changing the type in a supplementary PR. This way the potential uses of this field could be audited and then the `State` enum/int could be properly defined to include all possible states used by different vulnerability sources - what do you think? I could see this being useful for other vulnerability advisory sources also.

knqyf263:
Thanks. I'll merge this PR, then. I just wanted to warn you that the type might change shortly. If you depend on this field, it would be breaking compatibility.

>This way the potential uses of this field could be audited

On the trivy side, this field is not used. We added that for future use.

jhebden-gl:
> Thanks. I'll merge this PR, then. I just wanted to warn you that the type might change shortly. If you depend on this field, it would be breaking compatibility.

Thank you - I'll definitely watch out for changes to this field & update accordingly. At the moment we use it in some automation but might extend our usage further. I'll add some more tests to our code to make sure the value is the expected format so I can catch the change.

> On the trivy side, this field is not used. We added that for future use.

That's really helpful to know, thank you!

DmitriyLewen:
Hello @jhebden-gl 
I tested these changes and found one case:

I scanned `centos` image with 2 DBs and got different results:

db from your branch:
```
centos (centos 8.4.2105)
Total: 465 (UNKNOWN: 0, LOW: 168, MEDIUM: 269, HIGH: 28, CRITICAL: 0)
```

db from main branch:
```
centos (centos 8.4.2105)
Total: 465 (UNKNOWN: 0, LOW: 165, MEDIUM: 272, HIGH: 28, CRITICAL: 0)
```

CVE-2020-21674 (libarchive), CVE-2021-3997(systemd) and CVE-2021-3997(systemd-libs) have different severity.

I am looking into this issue. I'll write to you when I get more information.

jhebden-gl:
> Hello @jhebden-gl I tested these changes and found one case:
> <...snip...>
> I am looking into this issue. I'll write to you when I get more information.

Hello @DmitriyLewen - thank you for testing out these changes. One thing I suspect might be happening, is that the [merging code for advisories had to be updated](https://github.com/aquasecurity/trivy-db/pull/277/files#diff-1768c801e6c1230a63e41a360b0a667eb9db1b5c3352eb8e23ac826be47d3ac2R149) so that the `State` field was considered as a point of difference between otherwise similar advisories. I'm not 100% sure, but it is possible some advisories were being merged previously which perhaps should not have been merged. Perhaps severity also needs to be added as a point of difference between advisories when merging entries? Comparing the advisories for severity during this merging process might give some clue as to what has changed. I'll also look at this tomorrow to see if I can spot anything.

knqyf263:
@jhebden-gl @DmitriyLewen Thanks for your investigation! It reminds me of a known issue.
https://github.com/aquasecurity/trivy/issues/749#issuecomment-775055450

It seems like Red Hat finally added different severities according to platforms (I'm not sure when it happened, though). It would be great to support platform severities in another PR.

PR #279, opened 2023-02-10
ci: remove unused workflow for Trivy DB v1 publishing

afdesk:
Trivy DB v1 isn't supported now, this workflow is deprecated too.


knqyf263 (approved):

knqyf263:
cf. https://github.com/aquasecurity/trivy-db/pull/278

PR #281, opened 2023-02-16
feat(chainguard): add support for Chainguard distro

luhring:
Adds support for incorporating Chainguard secdb data into Trivy DB for https://github.com/aquasecurity/trivy/issues/3630

### DIFF
--- a/pkg/vulnsrc/chainguard/chainguard.go
+++ b/pkg/vulnsrc/chainguard/chainguard.go
@@ -23,7 +23,7 @@ const (
 var (
 	source = types.DataSource{
 		ID:   vulnerability.Chainguard,
-		Name: "Chainguard Secdb",
+		Name: "Chainguard Security Data",
 		URL:  "https://packages.cgr.dev/chainguard/security.json",
 	}
 )
@@ -104,7 +104,7 @@ func (vs VulnSrc) saveSecFixes(tx *bolt.Tx, platform, pkgName string, secfixes m
 			}
 		}
 	}
-	
+
 	return nil
 }
 
--- a/pkg/vulnsrc/chainguard/chainguard.go
+++ b/pkg/vulnsrc/chainguard/chainguard.go
@@ -1,0 +1,121 @@
+package chainguard
+
+import (
+	"encoding/json"
+	"io"
+	"path/filepath"
+	"strings"
+
+	bolt "go.etcd.io/bbolt"
+	"golang.org/x/xerrors"
+
+	"github.com/aquasecurity/trivy-db/pkg/db"
+	"github.com/aquasecurity/trivy-db/pkg/types"
+	"github.com/aquasecurity/trivy-db/pkg/utils"
+	"github.com/aquasecurity/trivy-db/pkg/vulnsrc/vulnerability"
+)
+
+const (
+	chainguardDir = "chainguard"
+	distroName    = "chainguard"
+)
+
+var (
+	source = types.DataSource{
+		ID:   vulnerability.Chainguard,
+		Name: "Chainguard Secdb",
+		URL:  "https://packages.cgr.dev/chainguard/security.json",
+	}
+)
+
+type VulnSrc struct {
+	dbc db.Operation
+}
+
+func NewVulnSrc() VulnSrc {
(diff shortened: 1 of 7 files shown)


pkg/vulnsrc/chainguard/chainguard.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func (vs VulnSrc) saveSecFixes(tx *bolt.Tx, platform, pkgName string, secfixes map[string][]string) error {
+	for fixedVersion, vulnIDs := range secfixes {
+		advisory := types.Advisory{
+			FixedVersion: fixedVersion,
+		}
+		for _, vulnID := range vulnIDs {
+			// See https://gitlab.alpinelinux.org/alpine/infra/docker/secdb/-/issues/3
+			// e.g. CVE-2017-2616 (+ regression fix)
```
knqyf263, line 93: question: Is it relevant for Chainguard?

pkg/vulnsrc/chainguard/chainguard.go
```diff
@@ -0,0 +1,121 @@
...
+
+const (
+	chainguardDir = "chainguard"
+	distroName    = "chainguard"
+)
+
+var (
+	source = types.DataSource{
+		ID:   vulnerability.Chainguard,
+		Name: "Chainguard Secdb",
```
knqyf263, line 26: nit: It is called "Chainguard Security Data" here. Which is the correct name for Chainguard security advisories?

pkg/vulnsrc/chainguard/chainguard.go
```diff
@@ -0,0 +1,121 @@
...
+	for fixedVersion, vulnIDs := range secfixes {
+		advisory := types.Advisory{
+			FixedVersion: fixedVersion,
+		}
+		for _, vulnID := range vulnIDs {
+			// See https://gitlab.alpinelinux.org/alpine/infra/docker/secdb/-/issues/3
+			// e.g. CVE-2017-2616 (+ regression fix)
+			ids := strings.Fields(vulnID)
+			for _, cveID := range ids {
+				cveID = strings.ReplaceAll(cveID, "CVE_", "CVE-")
```
knqyf263, line 96: question: Also, it is for sanitizing Alpine secdb. Do you think it is needed in Chainguard?

luhring: That's a good point. We should be able to guarantee that we're not putting things like `CVE_` in our data. I can remove this 👍

luhring: Pushed a change to simplify this — let me know what you think

pkg/vulnsrc/chainguard/chainguard.go
luhring: I don't think it has to be — pushed a change that removes this.

pkg/vulnsrc/chainguard/chainguard.go
knqyf263: Sorry I forgot to put a link. We want to be consistent in what we call it across our projects.
https://github.com/aquasecurity/vuln-list-update/blob/c98ad9e9909ecb97de0f8777131c251023167c81/main.go#L245

luhring: Pushed an update!

knqyf263:
As I commented [here](https://github.com/aquasecurity/vuln-list-update/pull/190#issuecomment-1483065015), we are planning to cut v0.40.0 shortly.

knqyf263:
@luhring What does `0` mean? It will never be detected. Is it intended?
https://github.com/aquasecurity/vuln-list/blob/842558b6dcc7cfa5934bdf353212b212e6926b36/chainguard/chainguard/git.json#L4-L6

luhring:
> @luhring What does `0` mean? It will never be detected. Is it intended?

Yes, we put `0` in the secdb to NAK vulnerabilities that scanners find, to let scanners know that no versions of the given package should be matched to the vulnerabilities. This is the same behavior as Wolfi and as Alpine.

knqyf263:
:shipit:

Issue #312, opened 2023-05-23
fix(RedHat): compare advisories by severity 


DmitriyLewen:
## Description
There are cases, when advisory have different severity for different platforms.
It is also required to compare recommendations by severity.

See more [here](https://github.com/aquasecurity/trivy-db/pull/277#issuecomment-1558803205)

knqyf263:
fixed

PR #313, opened 2023-05-23
fix(RedHat): compare severity when merging advisories

DmitriyLewen:
## Description
See #312


jhebden-gl:
Thanks for looking at this so quickly! I thought I'd just add some thoughts & a small review given I've been fairly familiar with this code lately. Looking at the implementation, your fix makes sense and I can definitely see how previously records with different severity could be merged. It's great to get this fixed. I noticed that you did not have to change any existing test cases, so maybe the cases where advisories were improperly merged would be fairly rare as noted by your testing in #312. Adding the additional test cases which highlight this issue is an excellent idea & the test cases looked good to me.

One other thought I had was - Red Hat also has different severities for different packages within the same advisory, sometimes for different suites. I think the different severites based on package name is not a problem with the current implementation, and adjusting to handle different severties as a different should address the difference in suites. Given the bucket is named after the package & the processing is done based on this, I think it would not be possible to accidentally merge two different packages with different severties together, so I think this covers all cases now. LGTM!

DmitriyLewen:
Thanks for your quick review @jhebden-gl !

pkg/vulnsrc/redhat-oval/redhat-oval.go
```diff
@@ -485,3 +487,24 @@ func archesEqual(a, b []string) bool {
 	}
 	return true
 }
+
+func cvesEqual(a, b []CveEntry) bool {
```
knqyf263, line 491: I'm concerned one definition includes several CVE-IDs with different severities. Is it possible? I guess it may happen with RHSA, but we need to check it. Hopefully, RHSA-ID will not conflict across platforms.

```
      "Cves": [
        {
          "CveID": "CVE-2020-21674",
          "Cvss3": "7.5/CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H",
          "Cwe": "CWE-787-\u003eCWE-122",
          "Impact": "moderate",
          "Href": "https://access.redhat.com/security/cve/CVE-2020-21674",
          "Public": "20191228"
        },
        {
          "CveID": "CVE-2020-21675",
          "Cvss3": "7.5/CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H",
          "Cwe": "CWE-787-\u003eCWE-122",
          "Impact": "low",
          "Href": "https://access.redhat.com/security/cve/CVE-2020-21675",
          "Public": "20191228"
        }
      ],
```

Also, are the severity and impact always the same? Doesn't the following case happen?
- Severity: "Moderate"
- Impact: "low"

```
    "Advisory": {
      "From": "secalert@redhat.com",
      "Severity": "Moderate",
      "Cves": [
        {
          "CveID": "CVE-2020-21674",
          "Impact": "low",
        }
      ],
```

jhebden-gl: Per https://access.redhat.com/security/updates/classification & based on my understanding of the parsing code (which is not as good as either of yours) the overall severity for an advisory (as an RHSA can contain multiple vulnerabilities) would always be the highest severity out of the individual vulnerabilities included in the advisory. So, I think when parsing the OVALv2 data, it makes sense to parse impact for each CVE and correlate that with the CVE impact & fix availability information, and not use the advisory severity, for better accuracy for individual findings for individual packages.

knqyf263 (approved):

PR #314, opened 2023-05-29
feat: switch glad to ghsa for golang

DmitriyLewen:
## Description 
`Go vulndb` has delays to import `GHSA`.
Even though `GHSA` still has advisories with package name and not module name, we can still use `GHSA` for new advisories


pkg/vulnsrc/ghsa/ghsa.go
```diff
@@ -26,6 +26,10 @@ var (
 	sourceID   = vulnerability.GHSA
 	ecosystems = []types.Ecosystem{
 		vulnerability.Composer,
+		// ghsa for `Go` ecosystem can use package name instead of module name
+		// e.g. github.com/aws/aws-sdk-go/service/s3/s3crypto
+		// but `go-vuln` updates advisories with a delay
+		// this is why we still use ghsa for `Go`
```
knqyf263, line 32: I'd add a link to an example advisory.

```suggestion
		// GHSA for `Go` ecosystem can use package names instead of module names
		// e.g. https://github.com/advisories/GHSA-6jvc-q2x7-pchv
		//         => github.com/aws/aws-sdk-go/service/s3/s3crypto (package name)
		// but `go-vuln` updates advisories with a delay.
		// This is why we still use GHSA for `Go`.
```

knqyf263 (approved):

PR #318, opened 2023-06-13
fix(osv): skip withdrawn entries

nikpivkin:
## Description
See https://github.com/aquasecurity/trivy/issues/4621


pkg/vulnsrc/osv/osv.go
```diff
@@ -123,6 +123,11 @@ func (vs VulnSrc) save(eco ecosystem, entries []Entry) error {
 }
 
 func (vs VulnSrc) commit(tx *bolt.Tx, eco ecosystem, entry Entry) error {
+
+	if entry.Withdrawn != nil {
```
afdesk, line 127: maybe len() will be better?
```suggestion
	if len(entry.Withdrawn) > 0 {
```

nikpivkin: `Withdrawn` is a pointer to time.Time

afdesk: oh, ok.

knqyf263: Do we want to compare the date? The following code doesn't work probably, but I just wanted to explain my thought.

```suggestion
	if entry.Withdrawn != nil && entry.Withdrawn.Before(time.Now()){
```

pkg/vulnsrc/osv/testdata/happy/vuln-list/osv/python/withdrawn_after_now.json
```diff
@@ -0,0 +1,102 @@
+{
+  "id": "PYSEC-2019-156",
+  "modified": "2021-07-05T00:01:27.093286Z",
+  "published": "2019-11-04T20:15:00Z",
+  "withdrawn": "2024-06-06T10:37:00Z",
```
knqyf263, line 5: This test will fail next year. Can we change it 3000 or something?

knqyf263 (approved):

pkg/vulnsrc/osv/testdata/happy/vuln-list/osv/python/withdrawn_after_now.json
afdesk: let's wait for it.
I'm going to see it. while everything is going well ))

PR #321, opened 2023-06-21
fix(rocky): add multi-arch support

DmitriyLewen:
## Description
`Rocky Linux` supports `aarch64`.
These are cases when fixed version for `x86_64` != `aarch64` or there is no fixed version for `x86_64` architecture.

We need to add advisories for `aarch64` and separate architectures.
I add `Advisory.FixedVersions` struct for this case.
It will be used to separate architectures, versions, vendorIDs.
**`Advisory.FixedVersion` field is still used for backwards compatibility.**

db size before - 361.0 Mb
db size after - 361.1Mb

buckets struct before:
image

buckets after after:
image

## Related Issues
- https://github.com/aquasecurity/trivy/issues/4634


pkg/types/types.go
```diff
@@ -116,6 +116,9 @@ type Advisory struct {
 	FixedVersion    string `json:",omitempty"`
 	AffectedVersion string `json:",omitempty"` // Only for Arch Linux
 
+	// Advisory can contain different fixed versions for different architectures
+	FixedVersions FixedVersions `json:",omitempty"`
```
knqyf263, line 120: `Entries` is better to get aligned with Red Hat. I'm not sure if it is a good name, though.
https://github.com/aquasecurity/trivy-db/blob/c76cb31f21102a3d51f22c09e46dc6110200ce82/pkg/vulnsrc/redhat-oval/types.go#L160-L162

pkg/vulnsrc/rocky/rocky.go
```diff
@@ -143,22 +139,45 @@ func (vs *VulnSrc) put(errataVer map[string][]RLSA) error {
...
+				fixedVersion := types.FixedVersion{
 					FixedVersion: utils.ConstructVersion(pkg.Epoch, pkg.Version, pkg.Release),
+					Arch:         pkg.Arch,
+					VendorID:     erratum.ID,
+				}
+
+				// if the advisory for this package and CVE have been kept - just add the new architecture
+				if adv, ok := input.Advisories[pkg.Name]; ok {
+					// don't include duplicates
+					if !adv.FixedVersions.IsDuplicate(fixedVersion) {
```
knqyf263, line 167: I guess `slices.Contains` works.

```suggestion
					if !slices.Contains(adv.FixedVersions, fixedVersion) {
```

DmitriyLewen:
I see 1 case why we can't do that ( but i didn't see example for Rocky linux, but OracleLinux contains this case):
There is Vendor ID with different fixed versions for different architectures. 
e.g. https://linux.oracle.com/errata/ELSA-2022-4803.html :
`Oracle Linux 7 (aarch64) | rsyslog-8.24.0-57.0.4.el7_9.3.src.rpm`
`Oracle Linux 7 (x86_64) | rsyslog-8.24.0-57.0.1.el7_9.3.src.rpm`

Also i have 1 thought for new struct:
We can add `Flavors` for OracleLinux - https://github.com/aquasecurity/trivy-db/pull/221

pkg/types/types.go
DmitriyLewen: I thought about it and couldn't find good name.
But you are right. We use `Enties` in RedHat, let's also use `Entries` here.

Done in https://github.com/aquasecurity/trivy-db/pull/321/commits/14b1e69623e02e77ffd327040011f50da3997643

pkg/vulnsrc/rocky/rocky.go
DmitriyLewen: Good idea! 
Changed in https://github.com/aquasecurity/trivy-db/pull/321/commits/14b1e69623e02e77ffd327040011f50da3997643

knqyf263 (approved):

PR #328, opened 2023-07-02
feat: add vulnerability status

knqyf263:
## Description
This PR adds the new type `Status` and changes redhat and debian to use that. The detail is described [here](https://github.com/aquasecurity/trivy/issues/4763).

## Issue
https://github.com/aquasecurity/trivy/issues/4763


pkg/vulnsrc/debian/debian.go
```diff
@@ -660,3 +660,19 @@ func compareVersions(v1, v2 string) (int, error) {
 
 	return ver1.Compare(ver2), nil
 }
+
+func newStatus(vulnID, s string) types.Status {
```
DmitriyLewen, line 664: do you need `vulnID` here?

pkg/vulnsrc/redhat-oval/redhat-oval.go
```diff
@@ -482,3 +485,20 @@ func severityFromImpact(sev string) types.Severity {
...
+		return types.StatusAffected
+	case "under investigation":
+		return types.StatusUnderInvestigation
+	case "will not fix":
+		return types.StatusWillNotFix
+	case "out of support scope":
+		return types.StatusEndOfLife
+	}
+	if s != "" {
+		panic(fmt.Sprintf("unknown status: %s", s))
```
DmitriyLewen, line 501: Are you sure about `panic`?

knqyf263: Oh, it was for testing...

pkg/vulnsrc/debian/debian.go
knqyf263: No, it was for testing...

DmitriyLewen (approved):

knqyf263:
@jhebden-gl Just FYI: we changed state into status. We believe it will not break your system as it is a different field.

jhebden-gl:
> @jhebden-gl Just FYI: we changed state into status. We believe it will not break your system as it is a different field.

Thank you so much for letting me know, I really appreciate it 🙇🏻

PR #332, opened 2023-07-12
feat: add k8s vulns

chen-keinan:
## Description 
Feat: add k8s vulns

Related [#4029](https://github.com/aquasecurity/trivy/issues/4029)
Related https://github.com/aquasecurity/trivy/issues/5020
Depend on: https://github.com/aquasecurity/vuln-list-update/pull/239


pkg/vulnsrc/k8svulndb/k8svulndb.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	var patchedVersions, vulnerableVersions []string
+	for index, introduced := range item.AffectedVersion {
+		vulnerableVersions = append(vulnerableVersions, fmt.Sprintf("%s, <=%s", introduced.From, introduced.To))
+		if len(item.FixedVersion) > index {
+			patchedVersions = append(patchedVersions, item.FixedVersion[index].Fixed)
+		}
+	}
+	a := types.Advisory{
+		PatchedVersions:    patchedVersions,
```
knqyf263, line 107: Isn't it enough?

```suggestion
		PatchedVersions:   lo.Map(item.FixedVersion, func(v Version, _ int) string { return v.Fixed})
```

pkg/vulnsrc/k8svulndb/types.go
```diff
@@ -0,0 +1,21 @@
+package k8svulndb
+
+type K8sCVE struct {
+	ID              string    `json:"id,omitempty"`
+	CreatedAt       string    `json:"created_at,omitempty"`
+	Summary         string    `json:"summary,omitempty"`
+	Component       string    `json:"component,omitempty"`
+	Description     string    `json:"description,omitempty"`
+	AffectedVersion []Version `json:"affected_version,omitempty"`
```
knqyf263, line 9: nit

```suggestion
	AffectedVersions []Version `json:"affected_version,omitempty"`
```

pkg/vulnsrc/k8svulndb/types.go
```diff
@@ -0,0 +1,21 @@
...
+	FixedVersion    []Version `json:"fixed_version,omitempty"`
```
knqyf263, line 10:
```suggestion
	FixedVersions    []Version `json:"fixed_version,omitempty"`
```

pkg/vulnsrc/k8svulndb/k8svulndb.go
```diff
@@ -0,0 +1,142 @@
...
+		VulnerableVersions: vulnerableVersions,
+	}
+	err := vs.dbc.PutAdvisoryDetail(tx, item.ID, item.Component, []string{bucketName}, a)
+	if err != nil {
+		return xerrors.Errorf("failed to save k8s-vulndb advisory: %w", err)
+	}
+	severity, err := types.NewSeverity(strings.ToUpper(item.Severity))
+	if err != nil {
+		severity = types.SeverityLow
```
knqyf263, line 116: Shouldn't we use UNKNOWN this case?

knqyf263 (commented):
I have some feedback on vuln-list-k8s.

1. CVSS
https://github.com/aquasecurity/vuln-list-k8s/blob/fedb568387ba3d6258bde457201ba4b1a2b14139/k8s/cves/CVE-2023-2878.json#L22-L24

This should be as below:

```
	"cvssv3": {
		"vector"* "CVSS:3.1/AV:L/AC:L/PR:L/UI:N/S:C/C:H/I:N/A:N",
		"score": 6.5
	}
```

The score is calculated based on the vector. They should be put together. Also, CVSS v4 is coming soon.

2. Component name
It looks like `k8s.io` is not only the prefix.

https://github.com/aquasecurity/vuln-list-k8s/blob/fedb568387ba3d6258bde457201ba4b1a2b14139/k8s/cves/CVE-2023-2878.json#L5C1-L6C1

https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/fa9befecb289b9e1a9848a39a18b104278a394bd/go.mod#L1

3. Affected version
The following affected version is wrong.

https://github.com/aquasecurity/vuln-list-k8s/blob/fedb568387ba3d6258bde457201ba4b1a2b14139/k8s/cves/CVE-2023-2878.json#L7-L11

The advisory says `- secrets-store-csi-driver < 1.3.3`, but another advisory says `v1.27.0 - v1.27.1`, which means `1.27.0 <= k8s <= 1.27.1`. The k8s advisory format is not standardized, unfortunately. We must be aware of `<` or `<=`.

[OSV](https://ossf.github.io/osv-schema/) defines them differently.

```
			"events": [ {
				"introduced": string,
				"fixed": string,
				"last_affected": string,
				"limit": string
			} ],
```

`fixed` is `<`, and `last_affected` is `<=`. It might be a good idea to adopt the same structure.

JonZeolla:
I think https://github.com/aquasecurity/trivy-db/pull/338 attempts to add support for `last_affected`

chen-keinan:
@knqyf263 all comments has been addressed, please have another look.

chen-keinan:
@knqyf263 lets wait with the PR approval , I need to review with you issue I found with upstream name

pkg/vulnsrc/k8svulndb/k8svulndb.go
```diff
@@ -0,0 +1,162 @@
...
+
+const (
+	vulnListDir = "vuln-list-k8s"
+	apiDir      = "k8s/cves"
+)
+
+var (
+	source = types.DataSource{
+		ID:   vulnerability.K8sVulnDB,
+		Name: "The k8s Vulnerability Database",
```
knqyf263, line 30: nit: They call it "Official CVE Feed" or "Official Kubernetes CVE Feed".
https://kubernetes.io/docs/reference/issues-security/official-cve-feed/
https://kubernetes.io/blog/2022/09/12/k8s-cve-feed-alpha/

```suggestion
		Name: "Official Kubernetes CVE Feed",
```

knqyf263 (commented):
Looks like we need to improve advisories more.

No patched version.
https://github.com/aquasecurity/vuln-list-k8s/blob/main/k8s/cves/CVE-2018-1002102.json

The version range seems wrong.
https://github.com/aquasecurity/vuln-list-k8s/blob/7a48561f95d894417cd493f6c7366d9de206efec/k8s/cves/CVE-2017-1002102.json#L8-L67
https://github.com/aquasecurity/vuln-list-k8s/blob/7a48561f95d894417cd493f6c7366d9de206efec/k8s/cves/CVE-2018-1002105.json#L8-L156

No last affected or fixed
https://github.com/aquasecurity/vuln-list-k8s/blob/main/k8s/cves/CVE-2019-11244.json
https://github.com/aquasecurity/vuln-list-k8s/blob/main/k8s/cves/CVE-2023-2728.json#L11-L15

Wrong introduced
https://github.com/aquasecurity/vuln-list-k8s/blob/main/k8s/cves/CVE-2020-10749.json

chen-keinan:
@knqyf263 I have fixed the collector and validated all CVEs content.
please review again

knqyf263:
@chen-keinan It seems to be challenging to correctly parse the official CVE feed. How did you improve that?

chen-keinan:
> @chen-keinan It seems to be challenging to correctly parse the official CVE feed. How did you improve that?

What I did was to pick to vulnerable versions and cve data from a more structure external vuln source (as appear in k8s db) `[https://cveawg.mitre.org/api/cve/<CVE-ID>]` and the only the component name, ref and summary was taken from k8s source vulndb

knqyf263:
Can I review the script first? The vulnerability data is essential.

chen-keinan:
> Can I review the script first? Where is it maintained? The vulnerability data is essential.

sure , the collector start at this [point](https://github.com/aquasecurity/k8s-db-collector/blob/main/collectors/cvedb/cve/collector.go#L23)

knqyf263:
Thanks. I'll take a look after completing all the tasks for v0.45.0.

pkg/vulnsrc/vulnerability/const.go
```diff
@@ -47,4 +48,5 @@ const (
 	Swift     types.Ecosystem = "swift"
 	Cocoapods types.Ecosystem = "cocoapods"
 	Bitnami   types.Ecosystem = "bitnami"
+	K8s       types.Ecosystem = "k8s"
```
knqyf263, line 51: It's recommended to use the same value.

```suggestion
	K8s       types.Ecosystem = "kubernetes"
```

pkg/vulnsrc/osv/osv.go
```diff
@@ -350,6 +350,8 @@ func convertEcosystem(eco Ecosystem) types.Ecosystem {
 		return vulnerability.Swift
 	case "bitnami":
 		return vulnerability.Bitnami
+	case "k8s":
```
knqyf263, line 353: It must be the same value used in OSV advisories.

```suggestion
	case "kubernetes":
```

pkg/vulnsrc/k8svulndb/k8svulndb.go
```diff
@@ -0,0 +1,47 @@
...
+type transformer struct{}
+
+func (t *transformer) TransformAdvisories(advs []osv.Advisory, entry osv.Entry) ([]osv.Advisory, error) {
+	if len(entry.Affected) > 0 {
+		if len(entry.Affected[0].Severities) > 0 {
+			severity := types.SeverityUnknown
+			vector := entry.Affected[0].Severities[0].Type
+			bm, err := metric.NewBase().Decode(vector)
+			if err == nil {
+				severity = types.Severity(bm.Severity())
```
knqyf263, line 37: FYI: Their severity is different from ours. Our iota is 0-4, but their iota is 0-5.

```
%!v(PANIC=String method: runtime error: index out of range [5] with length 5)
```

 We can remove this transformer, so we don't have to care about it, though.

pkg/vulnsrc/k8svulndb/k8svulndb.go
```diff
@@ -0,0 +1,47 @@
...
+			}
+			for i := range advs {
+				advs[i].Severity = severity
+				advs[i].CVSSScoreV3 = bm.Score()
+				advs[i].CVSSVectorV3 = vector
+			}
+		}
+	}
+	return advs, nil
+}
```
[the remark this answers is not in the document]
knqyf263: FYI: It doesn't work now due to this bug. I confirmed the severity was filled without this transformer after the bug gets fixed.
https://github.com/aquasecurity/vuln-list-update/pull/239#discussion_r1337073319

chen-keinan:
@knqyf263 comments has been addressed

knqyf263:
I had this branch on my local computer, but force-push broke it. Any benefits on force-push?

Actually, I also used to force-push when rebasing the main branch so that a merge commit won't be created, but I stopped doing that after talking with Simar because we currently use "squash and merge". The merge commit doesn't matter. It is more important to keep the commit history in PR, but please let me know if you have a specific reason.

pkg/vulnsrc/k8svulndb/k8svulndb.go
```diff
@@ -0,0 +1,30 @@
...
+		},
+	}
+	return osv.New(k8sDir, vulnerability.K8sVulnDB, sources, &defaultTransformer{})
+}
+
+type defaultTransformer struct{}
+
+func (t *defaultTransformer) TransformAdvisories(advs []osv.Advisory, entry osv.Entry) ([]osv.Advisory, error) {
+	return advs, nil
+}
```
knqyf263, line 30: You can pass nil, then the default transformer will be used.

```suggestion
	return osv.New(k8sDir, vulnerability.K8sVulnDB, sources, nil)
}
```

chen-keinan:
> I had this branch on my local computer, but force-push broke it. Any benefits on force-push?
> 
> Actually, I also used to force-push when rebasing the main branch so that a merge commit won't be created, but I stopped doing that after talking with Simar because we currently use "squash and merge". The merge commit doesn't matter. It is more important to keep the commit history in PR, but please let me know if you have a specific reason.

no specific reason , I had rebased with upstream and then made some changes so it required me to do force (following to rebase)

knqyf263:
@chen-keinan So, would you run `git merge main` instead of `git rebase main`? Then, you don't need force-pushing.

pkg/vulnsrc/k8svulndb/testdata/happy-fixed/vuln-list-k8s/upstream/CVE-2023-2727.json
```diff
@@ -0,0 +1,38 @@
...
+  "details": "Users may be able to launch containers using images that are restricted by ImagePolicyWebhook when using ephemeral containers. Kubernetes clusters are only affected if the ImagePolicyWebhook admission plugin is used together with ephemeral containers.",
+  "affected": [
+    {
+      "package": {
+        "ecosystem": "kubernetes",
+        "name": "k8s.io/apiserver"
+      },
+      "severity": [
+        {
+          "type": "CVSS:3.1",
```
DmitriyLewen, line 15: Here we are using `CVSS_V3` to decode severity - https://github.com/aquasecurity/trivy-db/blob/0155e8b2d9e3077da1e72db5e800c7138ae8926a/pkg/vulnsrc/osv/osv.go#L318

If i understand correctly - this is a typo (https://github.com/aquasecurity/vuln-list-k8s/pull/92/files#diff-bfaecf1529a98db9d6d59c17ddff6f603038d988190ae591c51fd58a494ace17R15). Can you change this to see severity in tests?

pkg/vulnsrc/k8svulndb/testdata/happy/vuln-list-k8s/upstream/CVE-2023-2728.json
```diff
@@ -0,0 +1,38 @@
...
+  "details": "Users may be able to launch containers using images that are restricted by ImagePolicyWebhook when using ephemeral containers. Kubernetes clusters are only affected if the ImagePolicyWebhook admission plugin is used together with ephemeral containers.\n\n",
+  "affected": [
+    {
+      "package": {
+        "ecosystem": "kubernetes",
+        "name": "k8s.io/apiserver"
+      },
+      "severity": [
+        {
+          "type": "CVSS:3.1",
```
DmitriyLewen, line 15: same

DmitriyLewen (commented):
LGTM. 
Left 1 comment about tests.

chen-keinan:
@DmitriyLewen thanks for the comment test data has been fixed

DmitriyLewen (approved):
Thanks!
Approved.

knqyf263:
We'll review the k8s advisories and merge the PR.
https://github.com/aquasecurity/vuln-list-k8s/pulls

knqyf263:
@chen-keinan What if we skip storing `k8s.io/kubectl` and `k8s.io/client-go`? They are not detected in Trivy k8s scanning, right? We will not use these buckets even if we store those vulnerabilities. Or am I missing something? Is KBOM able to detect `k8s.io/kubectl`?

FYI: `k8s.io/client-go` is a go dependency, so the vulnerabilities in `pkg:golang/k8s.io/client-go` are covered by GHSA.

chen-keinan:
> @chen-keinan What if we skip storing `k8s.io/kubectl` and `k8s.io/client-go`? They are not detected in Trivy k8s scanning, right? We will not use these buckets even if we store those vulnerabilities. Or am I missing something? Is KBOM able to detect `k8s.io/kubectl`?
> 
> FYI: `k8s.io/client-go` is a go dependency, so the vulnerabilities in `pkg:golang/k8s.io/client-go` are covered by GHSA.

I prefer to keep `kubectl` as I'm still looking at a way to find it. regarding `client-go` we can exclude it

chen-keinan:
> > @chen-keinan What if we skip storing `k8s.io/kubectl` and `k8s.io/client-go`? They are not detected in Trivy k8s scanning, right? We will not use these buckets even if we store those vulnerabilities. Or am I missing something? Is KBOM able to detect `k8s.io/kubectl`?
> > FYI: `k8s.io/client-go` is a go dependency, so the vulnerabilities in `pkg:golang/k8s.io/client-go` are covered by GHSA.
> 
> I prefer to keep `kubectl` as I'm still looking at simple ways to find it.

I also think that we can fallback `client-go` to k8s.io/kubernetes` wdyt ?

knqyf263:
>I prefer to keep kubectl as I'm still looking at simple ways to find it.

Do you mean you have an idea to find `kubectl` in `trivy k8s`?

knqyf263:
>I also think that we can fallback client-go to k8s.io/kubernetes where its not clear, wdyt ? otherwise lets exclude client-go

I did that in vuln-list-k8s. If the advisory says `client-go` is affected, it means the library `client-go` is affected, and doesn't affect other components.

knqyf263:
>I have an idea , but I do not like it , thinking on better way still. anyway I would like to keep the kubectl cves for now

Got it.

knqyf263:
Oh, my local branch is broken again. @chen-keinan Did you force-push again for some reason?

knqyf263 (approved):

PR #334, opened 2023-07-25
chore: fall back into vuln-list-reserve

knqyf263:
See https://github.com/aquasecurity/vuln-list-update/pull/229

Also, this PR drops Arch Linux since Trivy doesn't support it.


knqyf263:
worked
https://github.com/aquasecurity/trivy-db/actions/runs/5654264373/job/15316989713

DmitriyLewen (approved):

PR #338, opened 2023-08-01
feat(osv): support `last_affected` field

nikpivkin:
Added support for `last_affected` field for OSV.

Related PRs:
- [x] https://github.com/aquasecurity/vuln-list-update/pull/232

### DIFF
--- a/pkg/vulnsrc/osv/osv.go
+++ b/pkg/vulnsrc/osv/osv.go
@@ -164,6 +164,7 @@ func (vs VulnSrc) commit(tx *bolt.Tx, eco ecosystem, entry Entry) error {
 					}
 					vulnerable = fmt.Sprintf(">=%s", event.Introduced)
 				// Entries in the events array can contain either last_affected or fixed events, but not both
+				// cf. https://ossf.github.io/osv-schema/#requirements
 				case event.Fixed != "":
 					// patched versions
 					patchedVersions = append(patchedVersions, event.Fixed)


pkg/vulnsrc/osv/osv.go
```diff
@@ -169,6 +169,9 @@ func (vs VulnSrc) commit(tx *bolt.Tx, eco ecosystem, entry Entry) error {
 
 					// e.g. {"introduced": "1.2.0}, {"fixed": "1.2.5}
 					vulnerable = fmt.Sprintf("%s, <%s", vulnerable, event.Fixed)
+				case event.LastAffected != "":
```
knqyf263, line 172: We should add a comment about the following spec defined [here](https://ossf.github.io/osv-schema/#requirements).

>Entries in the events array can contain either last_affected or fixed events, but not both

pkg/vulnsrc/osv/osv.go
```diff
@@ -169,6 +169,9 @@ func (vs VulnSrc) commit(tx *bolt.Tx, eco ecosystem, entry Entry) error {
...
+					unaffectedVersions = append(unaffectedVersions, fmt.Sprintf(">%s", event.LastAffected))
```
knqyf263, line 173: Do we need `unaffectedVersions`? For example, `VulnerableVersions: []string{">=0, <=3.8.4"}` already describes `>3.8.5` is not affected.

pkg/vulnsrc/osv/osv.go
```diff
@@ -148,7 +148,7 @@ func (vs VulnSrc) commit(tx *bolt.Tx, eco ecosystem, entry Entry) error {
 
 	for _, affected := range entry.Affected {
 		pkgName := vulnerability.NormalizePkgName(eco.name, affected.Package.Name)
-		var patchedVersions, vulnerableVersions []string
+		var patchedVersions, vulnerableVersions, unaffectedVersions []string
```
knqyf263, line 151: We no longer need this variable, right?

nikpivkin: fixed

knqyf263 (approved):

PR #341, opened 2023-08-11
BugFix|Assigned erratum to custom put input

nikithaguduru:
erratum value is not being sent to custom "Put" input anymore which is resulting in tracker failure. This started happening post the recent changes in rocky tracker. 

Added back the erratum assignment to custom Put input

### DIFF
--- a/pkg/vulnsrc/rocky/rocky.go
+++ b/pkg/vulnsrc/rocky/rocky.go
@@ -222,6 +222,7 @@ func (vs *VulnSrc) commit(tx *bolt.Tx, platformName string, errata []RLSA) error
 			input.PlatformName = platformName
 			input.CveID = cveID
 			input.Vuln = vuln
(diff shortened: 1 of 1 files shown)


pkg/vulnsrc/rocky/rocky.go
```diff
@@ -222,6 +222,7 @@ func (vs *VulnSrc) commit(tx *bolt.Tx, platformName string, errata []RLSA) error
 			input.PlatformName = platformName
 			input.CveID = cveID
 			input.Vuln = vuln
+			input.Erratum = erratum
```
knqyf263, line 225: Can you please add a comment?

```suggestion
			input.Erratum = erratum // For Trivy Premium
```

knqyf263:
In favor of https://github.com/aquasecurity/trivy-db/pull/370

PR #343, opened 2023-08-21
fix: if break condition

testwill:


pkg/db/db.go
```diff
@@ -212,7 +212,7 @@ func (dbc Config) forEach(bktNames []string) (map[string]Value, error) {
 			bkt := root
 			for _, nestedBkt := range nestedBuckets {
 				bkt = bkt.Bucket([]byte(nestedBkt))
-				if bkt == nil {
+				if bkt != nil {
```
knqyf263, line 215: `==` is correct here

knqyf263 (changes requested):

PR #352, opened 2023-09-04
fix: OSV affected versions parsing

juan131:
This PR ensures the OSV parsing is compatible with OSV files that describe affected versions using both `affected.[]ranges` and `affected.[]versions` fields, see:

- https://ossf.github.io/osv-schema/#affected-fields

> (...) A version is considered affected if it lies within any one of the ranges or is listed in the versions list (...)

The current implementation only takes into account `affected.[]versions` as an alternative to affected.[]ranges`. However, both fields can coexist.


knqyf263:
>If the affected versions can be accurately summarized by one or more non-overlapping ranges, they may be encoded using the ranges field with an appropriate type (see below). In this case, the ranges act as a kind of compact form of a larger versions list.

If I understand correctly, we don't need to parse `affected.[]versions ` when `affected.[]ranges` exists.

knqyf263:
>A version is considered affected if it lies within any one of the ranges or is listed in the versions list

Or is it possible that they point to different versions?

go.mod
```diff
@@ -22,6 +22,8 @@ require (
 	k8s.io/utils v0.0.0-20201110183641-67b214c5f920
 )
 
+require github.com/Masterminds/semver v1.5.0
```
knqyf263, line 25: We have a better library.
https://github.com/aquasecurity/go-version

juan131:
> If I understand correctly, we don't need to parse `affected.[]versions` when `affected.[]ranges` exists. Or is it possible that they point to different versions?

We need to. Imagine a software has the following releases:

- 0.1.0 🐞 
- 0.2.0 🐞
- 0.3.0 🐞
- 1.0.0 ✅ 
- 1.1.0 🐞
- 1.1.1 ✅ 

There's an issue affecting every release on `0.` major branch that was fixed on `1.0.0` release. However, on `1.1.0` there was a regression and the issue was reintroduced, fixing it definitely on `1.1.1`. An OSV such as the one below would reflect that information perfectly:

```json
  (...)
  "affected": [
      (...)
      "ranges": [
        {
          "type": "SEMVER",
          "events": [
            {
              "introduced": "0"
            },
            {
              "fixed": "1.1.0"
            }
          ]
        }
      ]
      "versions": [
        "1.1.0"
      ]
```

go.mod
juan131: Fixed at https://github.com/aquasecurity/trivy-db/pull/352/commits/603ed00e23a9b5997ae6c02006495a716c962297
That said, I see a problem with `github.com/aquasecurity/go-version/semver` vs `github.com/Masterminds/semver`. Given an incomplete version such as `0.1` (it should be `0.1.0`), `semver.Parse` returns an error (aquasecurity) while `semver.NewVersion("0.1")` (Masterminds) doesn't and fills the version.
See https://go.dev/play/p/YrZDE3Gyypz

juan131:
@knqyf263 did you have a chance to take a look to my comments?

knqyf263:
No, I didn't.
I've set my status to busy. I'm just wondering if GitHub didn't notify you.

juan131:
No worries!! All the best to you for a happy and restful paternity leave

carrodher:
Hi all! Since @knqyf263 is on leave, we'd greatly appreciate it if someone could take a look and provide feedback. Your assistance would help keep things moving forward.

cc @dmitriylewen

knqyf263:
@DmitriyLewen is also on vacation. He will be back next week.

knqyf263:
I still have less capacity, but I'm gradually going back to work.

@juan131 In your example, the following advisory is better, no?
https://github.com/aquasecurity/trivy-db/pull/352#issuecomment-1706102184

```
  (...)
  "affected": [
      (...)
      "ranges": [
        {
          "type": "SEMVER",
          "events": [
            {
              "introduced": "0.1.0"
            },
            {
              "fixed": "1.0.0"
            }
          ]
        }
        {
          "type": "SEMVER",
          "events": [
            {
              "introduced": "1.1.0"
            },
            {
              "fixed": "1.1.1"
            }
          ]
        }
      ]
      "versions": [
        "0.1.0",
        "0.2.0", 
        "0.3.0",
        "1.1.0"
      ]
```

knqyf263:
I guess it is technically allowed to have different versions from ranges (non-overlapped), but I don't see such use cases so far.

As written in the documentation, ranges should be a summary of affected versions if possible.

>If the affected versions can be accurately summarized by one or more non-overlapping ranges, they may be encoded using the ranges field with an appropriate type (see below).

I'm open to supporting this case as it's probably not violating the spec, but I'd recommend following the convention in the Bitnami database.

>In this case, the ranges act as a kind of compact form of a larger versions list.

juan131:
IMHO listing `0.1.0`, `0.2.0` & `0.3.0` in `versions` array when you have the range below is redundant. I mean, it's adding an information that was already present. That said, the main problem is not redundancy but obtaining the list of versions when the information obtained from NVD is a range (we'll have to go to the project's release page and obtain that information and that can overcomplicate the OSV generation a lot).

```json
{
  "type": "SEMVER",
  "events": [{
    "introduced": "0.1.0"
   }, {
     "fixed": "1.0.0"
   }]
}
```

Regarding the second range you mentioned, it's true that I'd rather have it rather than having the `1.1.0` version listed in the `versions` array. However, the problem is that it's not easy to extract the information about when it was fixed from NVD when it's listed as a "single affected version". Why? Again because you don't know what's the version that came after the vulnerable one unless you extract that info from the project's release page.

knqyf263:
>I mean, it's adding an information that was already present

Yes, but again, it is recommended in OSV. Please refer to the document.
https://ossf.github.io/osv-schema/#affected-fields

>However, the problem is that it's not easy to extract the information about when it was fixed from NVD when it's listed as a "single affected version".

I think that is exactly why `last_affected` is defined in the spec. The following section mentions false negatives.
https://ossf.github.io/osv-schema/#last_affected-vs-fixed-example

knqyf263:
In short, there are four patterns:

1. ranges == versions
2. ranges: empty, versions: cover all affected versions
3. ranges: cover all affected ranges, versions: empty
4. ranges and versions are different

OSV recommends `#1`, but I understand it is sometimes challenging. In that case, `#2` or `#3` is better than `#4`, which is a bit confusing IMO. Considering the NVD situation, `#2` with `last_affected` looks the best for the Bitnami database.

juan131:
So.. you're suggesting sth like this, right?

```json
  (...)
  "affected": [
      (...)
      "ranges": [
        {
          "type": "SEMVER",
          "events": [
            {
              "introduced": "0.1.0"
            },
            {
              "fixed": "1.0.0"
            }
          ]
        },
        {
          "type": "SEMVER",
          "events": [
            {
              "introduced": "1.1.0"
            },
            {
              "last_affected": "1.1.0"
            }
          ]
        }        
      ]
      "versions": [] <- ONLY FILL THIS IF EVERY AFFECTED VERSION CAN BE RETRIEVED 
```

knqyf263:
I said `versions: empty`, but I actually mean either `versions` or `ranges` should cover all cases if possible. If `ranges` covers all affected ranges, the `versions` field doesn't matter, like empty, partially enumerated, etc.

Anyway, I'm open to `#4` if you'd go with that. It is your database. Rather than covering all cases by combining versions and changes (`#4`), I just wondered whether it would be more straightforward if both covered all cases (#1) even though it is redundant as recommended by OSV, or if either versions or changes covered all cases (#2 or #3).

gongomgra:
Hi @knqyf263,

Thanks for your messages. We will go with option `#3` in our database and we will try to revisit option `#1` in the future. I will keep you posted in the [bitnami/vulndb/issue](https://github.com/bitnami/vulndb/issues/133) when we have updated the database.

juan131:
Is there anything blocking this PR to be merged?

knqyf263:
Yes, we should care about ecosystem versioning. Many languages such as Python, Ruby and Java don't use semver. I implemented ecosystem-aware version comparison, but forgot to push it and left my computer in the office. I'll update this PR tomorrow.

knqyf263:
> Yes, we should care about ecosystem versioning

Otherwise, it always returns false, and a list of versions is always inserted into DB.
https://github.com/aquasecurity/trivy-db/blob/9f0210916dc28b3b53115b925aa303bcae39f990/pkg/vulnsrc/osv/osv.go#L289

knqyf263:
Looks good to me now.
@DmitriyLewen Can you please take a look?

DmitriyLewen (approved):
LGTM.

knqyf263:
Thanks, @DmitriyLewen . There are two problems.

1. NVD doesn't have the correct fixed versions and version ranges ([GHSA has them](https://github.com/discourse/discourse/security/advisories/GHSA-gr5h-hm62-jr3j), though).
2. The version ranges look redundant (a lot of ranges).

We might need to combine NVD and GHSA or manually review advisories for the 1st problem.

The second problem actually doesn't matter for trivy-db as the stored data will be the same as a list of versions. But just to clarify, I've just suggested using either/both ranges or versions rather than mixing them. I didn't mean we should always use either ranges or versions. To clean up the advisory, we might want to make a list of versions when no ranges are detected in the advisory (but again, we don't mind about versions or ranges).

e.g. mixed versions and ranges 
- versions:
  - 1.1
  - 4.1
- ranges: 
  - >=2.1, <2.3
  - >=3.1, <3.2

Additionally, the above case is also covered after this PR gets merged. So, we're ok that you'll go with the mixed approach. I'm just trying to sort out the situation, and you can decide it 👍

knqyf263 (approved):

juan131:
Thanks @knqyf263 @DmitriyLewen

PR #345, opened 2023-08-25
feat: use the GHSA repository

knqyf263:
## Description
This PR GHSA from its API to [the repository](https://github.com/github/advisory-database).

## TODO
- [x] Drop OSV PyPI
    - GHSA imports [Python Packaging Advisory database](https://github.com/pypa/advisory-database). We no longer need to fetch the OSV database.
    - https://docs.github.com/en/code-security/security-advisories/global-security-advisories/about-the-github-advisory-database
- [x] Make OSV implementation more generic
    - Change the OSV implementation to be more universally usable, since many advisories (e.g. [Bitnami](https://github.com/bitnami/vulndb)) have recently adopted the OSV format
- [x] Migrate GHSA from API to the repository
- [x] Write tests

PR #355, opened 2023-09-21
fix(ghsa): return Rust advisories 

DmitriyLewen:
## Description
After #345 changes we addув check of ecosystems for OSV - https://github.com/aquasecurity/trivy-db/blob/d5388c99ca492bf0c8822b27a2e5190794543bb5/pkg/vulnsrc/osv/osv.go#L143-L146 
GHSA uses `Rust`ecosystem -https://github.com/aquasecurity/trivy-db/blob/d5388c99ca492bf0c8822b27a2e5190794543bb5/pkg/vulnsrc/ghsa/ghsa.go#L32
OSV uses `Cargo` ecosystem - https://github.com/aquasecurity/trivy-db/blob/d5388c99ca492bf0c8822b27a2e5190794543bb5/pkg/vulnsrc/osv/osv.go#L345-L346

That is why we don't save Rust advisories.

Result DB:
изображение

изображение

### DIFF
--- a/pkg/vulnsrc/ghsa/ghsa.go
+++ b/pkg/vulnsrc/ghsa/ghsa.go
@@ -58,6 +58,10 @@ func (GHSA) Update(root string) error {
 			Name: fmt.Sprintf(platformFormat, cases.Title(language.English).String(string(ecosystem))),
 			URL:  fmt.Sprintf("https://github.com/advisories?query=type%%3Areviewed+ecosystem%%3A%s", ecosystem),
 		}
+		// Trivy uses Cargo for Rust
+		if ecosystem == vulnerability.Rust {
+			ecosystem = vulnerability.Cargo
+		}
 		dataSources[ecosystem] = src
 
 		// CocoaPods' vulnerability detection uses the Swift advisories.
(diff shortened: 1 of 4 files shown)


knqyf263:
@DmitriyLewen I've removed the `Rust` const so we will not make this kind of mistake. If I understand correctly, it is not used outside of GHSA. Please correct me if I'm wrong.

DmitriyLewen:
>  If I understand correctly, it is not used outside of GHSA

You are right. Trivy and Trivy-db don't use this constant. We can remove it.

knqyf263:
Thanks for confirming. And I explicitly split GHSA ecosystems and Trivy ecosystems as it was confusing. If it looks good to you, I'll merge this PR.

knqyf263 (approved):

DmitriyLewen:
> I explicitly split GHSA ecosystems and Trivy ecosystems as it was confusing

This is a better way than mine. Thanks for the help!
Let's merge this PR.

knqyf263:
Kicked
https://github.com/aquasecurity/trivy-db/actions/runs/6258596006

PR #358, opened 2023-09-26
refactor: use types.Ecosystem in bucket.Name()

knqyf263:
Using `types.Ecosystem` prevents misuse of the first argument of bucket.Name.

### DIFF
--- a/pkg/vulnsrc/glad/glad.go
+++ b/pkg/vulnsrc/glad/glad.go
@@ -8,6 +8,8 @@ import (
 	"strings"
 
 	bolt "go.etcd.io/bbolt"
+	"golang.org/x/text/cases"
+	"golang.org/x/text/language"
 	"golang.org/x/xerrors"
 
 	"github.com/aquasecurity/trivy-db/pkg/db"
@@ -21,8 +23,9 @@ const (
 	// GitLab Advisory Database
 	gladDir = "glad"
 
-	Conan packageType = "Conan"
-	Maven packageType = "Maven"
+	// cf. https://gitlab.com/gitlab-org/security-products/gemnasium-db/-/tree/e4176fff52c027165ae5a79f5b1193090e2fbef0#package-slug-and-package-name
+	Conan packageType = "conan"
+	Maven packageType = "maven"
 )
 
 var (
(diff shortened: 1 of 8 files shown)


pkg/vulnsrc/glad/glad.go
```diff
@@ -21,20 +21,21 @@ const (
...
+	supportedIDPrefixes = []string{
+		"CVE",
+		"GHSA",
+		"GMS",
+	}
+
+	ecosystems = map[packageType]types.Ecosystem{
+		Maven: vulnerability.Maven,
+		Conan: vulnerability.Conan,
+	}
```
DmitriyLewen, line 38: What if we just use slice of ecosystems? It should avoid confusion between ecosystems and packageTypes.
```suggestion
	ecosystems = []types.Ecosystem{
		vulnerability.Maven,
		vulnerability.Conan,
	} 
```

DmitriyLewen (commented):
@knqyf263 left small comment

pkg/vulnsrc/glad/glad.go
knqyf263: They can be different values because types.Ecosystem is defined by us. We anyway need a mapping between GitLab package types and our ecosystem values, right? Do you have any ideas?

DmitriyLewen: When i checked code - i found that we only use `packageType` for path to advisories.
But after your comment i understand that we use glad types in `vuln-list-update` - https://github.com/aquasecurity/vuln-list-update/blob/26dae1a5f7eccb8091db8288b679d215a8b11d88/glad/glad.go#L163
And if we remove `packageType` - we can get error in case, when ecosystem name != GitLab package types (as is case with Rust and Cargo for ghsa).

Now i think your solution is okay.
But can you add link to GitLab package types - https://gitlab.com/gitlab-org/security-products/gemnasium-db#package-slug-and-package-name

knqyf263: Good idea. Pushed.
https://github.com/aquasecurity/trivy-db/pull/358/commits/a73bf41682e2f44e2ed9c038b616d80642a42bc6

DmitriyLewen (approved):

PR #365, opened 2023-10-19
feat(ubuntu): support Ubuntu 23.10

ptodev:
Ubuntu 23.10's code name is Mantic Minotaur - see https://wiki.ubuntu.com/Releases.


knqyf263:
The test is failing, but it looks irrelevant to this change.
@DmitriyLewen Can you please take a look?

DmitriyLewen:
@knqyf263 problem in random order of GLAD ecosystem map.
I created fix for this to avoid this error next time - #366.

I also checked db with this change.
DB looks correct

изображение


You can merge this PR.

knqyf263 (approved):

knqyf263:
@ptodev Thanks

PR #366, opened 2023-10-20
chore: exclude Maven for GLAD

DmitriyLewen:
## Description
Exclude `Maven` ecosystem for GLAD.
See more information below.


knqyf263:
We can drop Maven in GLAD.

DmitriyLewen:
Do you want to use only GHSA for Maven?

DmitriyLewen:
I am not sure that i understood you.
Do you want to remove Maven from GLAD?

knqyf263:
Yes. Are there any unique vulnerabilities in GLAD from GHSA? People likely refer to GHSA and GLAD sometimes results in confusion. If GHSA has almost all vulnerabilities, we can remove GLAD Maven. What do you think?

DmitriyLewen:
I agree to remove `Maven` from `GLAD`.
I am more likely to create problems for updating` GLAD` advisories than` GHSA` advisories.
You're also right - users are more likely to use `GHSA` advisories.

Let me check the unique CVEs for `GLAD`.
If this does not have a big impact, we will remove `Maven` for `GLAD`.

knqyf263:
It is not necessarily identical. It is acceptable if the difference is small enough.

DmitriyLewen:
I saved unique identifiers (CVE, GHSA):
ghsa - 3924 unique identifiers
rad - 4229 uniqIDs

their comparison - https://gist.github.com/DmitriyLewen/945827ce78d1aed00e6a7f47b87510c7
Some of the missing CVEs by ghsa are old CVEs.

wdyt? Do you want me to investigate further?

knqyf263:
Can you give it another shot? For example, CVE-2023-40743 exists in GHSA.

DmitriyLewen:
Sorry, i forgot to update https://github.com/github/advisory-database.
ghsa has 3999 unique identifiers

Also i updated diff file.

Also i found cases when one DB uses GHSA and another uses CVE
e.g. `CVE-2023-30537` for GHSA and `GHSA-vrr8-fp7c-7qgp` for GLAD.

knqyf263:
I'm reviewing some advisories. Most of the missing advisories are GitHub-unreviewed and Jenkins plugins.
https://github.com/advisories/GHSA-3vcr-579j-4x48

DmitriyLewen:
> GitHub-unreviewed

We can hope that GitHub will add these guidelines.

my opinion is that we can exclude Maven from GitLab.

I made the changes - https://github.com/aquasecurity/trivy-db/pull/366/commits/ad4ab6ad66286261d6a597997f8cf8e6680fddc3

pkg/vulnsrc/glad/glad.go
```diff
@@ -35,8 +34,8 @@ var (
 		"GMS",
 	}
 
+	// Mapping between GLAD slug and Trivy ecosystem
 	ecosystems = map[packageType]types.Ecosystem{
```
DmitriyLewen, line 38: I didn't delete this map because of 2 points:
- don't forget that we need to compare the trivy-db ecosystem and the GLAD pool.
- keep the easy way to add a new ecosystem if needed.

knqyf263 (approved):

wagde-orca:
hi @knqyf263 and @DmitriyLewen 
I think removing gitlab causes problems in detecting some CVEs... and problems in detecting fixed versions of other CVEs...
i saw many changes between gitlab and github like in fixed versions of CVE-2019-12384 which is more accurate in gitlab... and many missing CVEs in github...
I think this removal is drastic and should have been done more carefully... 
I guess, that at least you should give the user an option to keep gitlab or not use it when downloading the DB...
Thanx
I can share more info
 if needed

wagde-orca:
more missing data in github is CVEs for
  [org.jenkins-ci.main:jenkins-core]
for example CVE-2020-2251
https://www.jenkins.io/security/advisory/2020-09-01/

knqyf263:
@wagde-orca You can raise PRs on GitHub.

Issue #369, opened 2023-11-02
Rocky multi-arch support broke backward compatibility


Moniseeta:
https://github.com/aquasecurity/trivy-db/pull/321 changed the structures for storing rocky advisories. 
https://github.com/aquasecurity/trivy-db/pull/321/files#diff-8e49337a1417a39a62ebf427e84fb2faa566c96e37ae627ac484c05dbe65c483R132 - Here, Custom field is missing
https://github.com/aquasecurity/trivy-db/pull/321/files#diff-4e9d5bb9d02520fbc8f26be7aa6e6eaebbc5c1eb8a320f3f681b3e747a306ab6R225 - Here erratum field is not being passed to PutInput 

This is causing break in backward compatibility

PR #370, opened 2023-11-02
fix rocky: multi-arch support broke backward compatibility

Moniseeta:
Added handling for Custom and Erratum fields

Resolves: #369


knqyf263 (approved):

knqyf263:
@Moniseeta You need to sign the CLA.

PR #384, opened 2024-02-15
ci: use maximize-build-space

### LINKED ISSUE #383: Publishing Trivy DB is failing
Hey, just wanted to raise this in case it's not known yet:

The scheduled Trivy DB Github Action has started failing persistently since Feb 13 ([GH Actions](https://github.com/aquasecurity/trivy-db/actions)).

It's erroring out with ([example run](https://github.com/aquasecurity/trivy-db/actions/runs/7911759389)):

```
System.IO.IOException: No space left on device : '/home/runner/runners/2.313.0/_diag/Worker_20240213-180155-utc.log'
```

This is blocking new releases of the Trivy DB from getting published.

According to [Github docs](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories) Github-hosted runners only come with 14GB disk space guaranteed. Not sure how much disk space we can easily safe while building the Trivy DB, but maybe reducing log volume could help as a temporary fix?

DmitriyLewen:
## Description
cache dir size is 12GB:
```
➜ du -sh ./cache 
12G    ./cache
➜ make db-build
...
➜ du -sh ./cache/db 
2.7G	./cache/db
➜ du -sh ./cache   
16G	./cache
```
So we need move space to build, compact and compress db.
Add `maximize-build-space`.

## Related Issues
- [ ] Close #383

### DIFF
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -16,7 +16,7 @@ jobs:
       - name: Maximize build space
         uses: easimon/maximize-build-space@v10
         with:
-          root-reserve-mb: 32768 # vuln-list dirs + language repositores use more that 12GB of storage
+          root-reserve-mb: 32768 # vuln-list dirs + language repositories use more than 12GB of storage
           remove-android: "true"
           remove-docker-images: "true"
           remove-dotnet: "true"
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -13,6 +13,15 @@ jobs:
     name: Build DB
     runs-on: ubuntu-latest
     steps:
+      - name: Maximize build space
+        uses: easimon/maximize-build-space@v10
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/cron.yml
```diff
@@ -13,6 +13,15 @@ jobs:
     name: Build DB
     runs-on: ubuntu-latest
     steps:
+      - name: Maximize build space
+        uses: easimon/maximize-build-space@v10
+        with:
+          root-reserve-mb: 32768 # vuln-list dirs + language repositores use more that 12GB of storage
```
knqyf263, line 19: typo?

```suggestion
          root-reserve-mb: 32768 # vuln-list dirs + language repositories use more than 12GB of storage
```

DmitriyLewen: Thanks! fixed.

knqyf263 (approved):

PR #407, opened 2024-06-04
feat: Expose Ubuntu fix status for downstream consumption

### LINKED ISSUE #408: feat: Expose Ubuntu fix status for downstream consumtion.
### Summary

Trivy Ubuntu advisories provide a FixedVersion when there is a released fix for a package, Affected can be inferred when the advisory is present without a FixedVersion, it does not expose any other status that Canonical use to indicate the status of a fix. This is insufficient to infer a status of ignored, pending or needed as these status show the package is affected and in the process of getting to a fixed version or not if the status is ignored.

PR is #407

skahn007gl:
## Summary
resolves #408 

Trivy Ubuntu advisories provide a `FixedVersion` when there is a released fix for a package, `Affected` can be inferred when the advisory is present without a `FixedVersion`, it does not expose any other status that Canonical use to indicate the status of a fix. This is insufficient to infer a status of `ignored`, `pending` or `needed` as these status show the package is affected and in the process of getting to a fixed version or not if the status is `ignored`.

This change exposes the Status provided in launchpad advisories without changed existing behaviour to populate the notes field in an advisory with `Status:$status`

## Changes

- `gitignore:21` :: name change of `cache`, `assets` to `_cache`, `_assets` to reflect directory name change to support gopls lint ignore declaration of `_` prefix on directories.

- `MakeFile` :: Adjusting naming convention of `cache` & `_assets` to avoid gopls linter causing max open file errors on MacOS.

- `pkg/types/status.go` :: New status added to support all potential status on launchpad advisories. 

- `pkg/vulnsrc/ubuntu/ubuntu.go` :: Expanded target status to capture all provided status's. Adjusted logic to use these status. Added ubuntu status to trivy status normalisation function.

- `pkg/vulnsrc/ubuntu/ubuntu_test.go` :: Added test for status normalisation.


.gitignore
```diff
@@ -17,8 +17,10 @@ coverage.txt
 
 **/.DS_Store
 
-assets
-cache
+
+#added _ to exclude from gopls linting, specifically helps when working on MacOS
+_assets
+_cache
```
jhebden-gl, line 23: Perhaps it would be better to add the `_assets` and `_cache` rename to a separate PR, to make this one easier to review and merge? I think the changes are not necessary for the Ubuntu fix state functionality, but more of a development nice-to-have if you're using an editor which is helped by this change.

pkg/types/status.go
```diff
@@ -20,18 +23,21 @@ var (
...
+		"pending",
 	}
 )
 
 const (
 	StatusUnknown Status = iota
 	StatusNotAffected
 	StatusAffected
-	StatusFixed
+	StatusFixed // Ubuntu "Released" Mapped to Fixed
```
jhebden-gl, line 35: I think this comment can be removed, as this is a global file and this is an Ubuntu specific note, this state is used by multiple providers.

pkg/types/status.go
```diff
@@ -20,18 +23,21 @@ var (
...
 	StatusUnderInvestigation
-	StatusWillNotFix // Red Hat specific
+	StatusWillNotFix // Red Hat specific // Ubuntu Specific "ignored" Superseded || End of life || End of Kernal support
```
jhebden-gl, line 37:
```suggestion
	StatusWillNotFix
```
I think that now that this is not specific to a single provider, it makes sense to just remove the comment.

pkg/types/status.go
```diff
@@ -20,18 +23,21 @@ var (
...
 	StatusFixDeferred
 	StatusEndOfLife
+	StatusPending //Ubuntu Specific "pending","needed"
```
jhebden-gl, line 40:
```suggestion
	StatusPending
```
I've suggested above this state can just be used as `StatusFixDeferred`, but if there is a reason we need extra states here, as this is also used by Red Hat IIRC, it makes sense to just remove the comment, as this state can be used by multiple providers.

pkg/types/status.go
```diff
@@ -20,18 +23,21 @@ var (
 		"will_not_fix",
 		"fix_deferred",
 		"end_of_life",
+		"needed",
+		"pending",
```
jhebden-gl, line 27: Does it make sense to simply normalise these cases to `fix_deferred`? These all seem to imply the fix is not available _yet_ but will be available in the _future_, as opposed to `will_not_fix`?

pkg/vulnsrc/ubuntu/ubuntu.go
```diff
@@ -213,3 +214,20 @@ func SeverityFromPriority(priority string) types.Severity {
...
+
+// StatusFromUbuntuStatus normalises Ubuntu status into common Trivy Types
+func StatusFromUbuntuStatus(status string) types.Status {
+	switch status {
+	case "not-affected", "DNE", "not-vulnerable", "ignored":
+		return types.StatusWillNotFix
+	case "deffered":
+		return types.StatusFixDeferred
+	case "needed", "pending":
+		return types.StatusPending
```
jhebden-gl, line 226:
```suggestion
	case "deferred", "needed", "pending":
		return types.StatusFixDeferred
```
Corrected what appears to be a typo in the `case`, and suggesting these are treated as the same case to avoid needing to extend the supported state list for Trivy DB.

pkg/vulnsrc/ubuntu/ubuntu_test.go
```diff
@@ -60,3 +62,65 @@ func TestVulnSrc_Update(t *testing.T) {
...
+		{
+			name:     "needed",
+			status:   "needed",
+			expected: types.StatusPending,
+		},
+		{
+			name:     "pending",
+			status:   "pending",
+			expected: types.StatusPending,
+		},
```
jhebden-gl, line 107:
```suggestion
			name:     "needed",
			status:   "needed",
			expected: types.StatusFixDeferred,
		},
		{
			name:     "pending",
			status:   "pending",
			expected: types.StatusFixDeferred,
		},
```
In line with above comments

jhebden-gl (changes requested):
Whilst I'm not a maintainer here, I am aware of the scope of this work and did work on the RHEL changes for fix status, so thought I'd add some review comments to try and help to save some time for the reviewers.

.gitignore
skahn007gl: I agree. I will remove these changes from this MR

pkg/vulnsrc/ubuntu/ubuntu.go
skahn007gl: Good catch on the typo

I have compressed StatusPending into StatusFixDeferred

skahn007gl:
> Whilst I'm not a maintainer here, I am aware of the scope of this work and did work on the RHEL changes for fix status, so thought I'd add some review comments to try and help to save some time for the reviewers.

@jhebden-gl Thanks or the feedback. I have pushed a new change set with your suggested changes

jhebden-gl:
Hi @knqyf263 👋🏻 - I was wondering if you might be able to assist with a review on this PR? We are hoping to use this data downstream. Thank you!

.gitignore
```diff
@@ -17,8 +17,10 @@ coverage.txt
 
 **/.DS_Store
 
+
+#added _ to exclude from gopls linting, specifically helps when working on MacOS
```
knqyf263, line 21: Forgot to revert?

.gitignore
```diff
@@ -17,8 +17,10 @@ coverage.txt
...
 assets
-cache
+cache
```
knqyf263, line 23:
```suggestion
cache
```

pkg/vulnsrc/ubuntu/ubuntu.go
```diff
@@ -213,3 +214,18 @@ func SeverityFromPriority(priority string) types.Severity {
 		return types.SeverityUnknown
 	}
 }
+
+// StatusFromUbuntuStatus normalises Ubuntu status into common Trivy Types
+func StatusFromUbuntuStatus(status string) types.Status {
+	switch status {
+	case "not-affected", "DNE", "not-vulnerable", "ignored":
```
knqyf263, line 221: I don't think `not-affected`, `DNE` and `not-vulnerable` are "Will not fix". They're "Not affected", and we don't want to save them in the database as it increases the size.

.gitignore
skahn007gl: I did miss it when reverting

pkg/vulnsrc/ubuntu/ubuntu.go
skahn007gl: I agree.
I will push a change set for this today.

knqyf263: `ignored` makes sense to me as it's considered "Will not fix".

skahn007gl: @knqyf263 I have pushed all the tidy up changes. Please re-review when you have a chance.
Thanks

pkg/vulnsrc/ubuntu/ubuntu.go
```diff
@@ -170,7 +170,8 @@ func defaultPut(dbc db.Operation, tx *bolt.Tx, advisory interface{}) error {
 			}
 
 			adv := types.Advisory{}
-			if status.Status == "released" {
+			adv.Status = StatusFromUbuntuStatus(status.Status)
+			if adv.Status == types.StatusFixed {
 				adv.FixedVersion = status.Note
 			}
```
knqyf263, line 176: We don't want to store the "fixed" status in DB so that we can keep the DB size as small as possible.

```suggestion
			status = StatusFromUbuntuStatus(status.Status)
			if status == types.StatusFixed {
				adv.FixedVersion = status.Note
			} else {
				// Store the status only if it's unfixed
				adv.Status = status
			}
```

jhebden-gl: Ah, this makes a lot of sense. Thanks for the suggested change @knqyf263! We already have code on our side which treats advisories with `FixedVersion` set as status fixed (we use this for Red Hat advisories) so this approach will be consistent too.

pkg/vulnsrc/ubuntu/ubuntu.go
```diff
@@ -170,8 +170,12 @@ func defaultPut(dbc db.Operation, tx *bolt.Tx, advisory interface{}) error {
...
 
 			adv := types.Advisory{}
-			if status.Status == "released" {
+			status = StatusFromUbuntuStatus(status.Status)
```
DmitriyLewen, line 173: resolve type mismatch

pkg/vulnsrc/ubuntu/ubuntu.go
```diff
@@ -23,7 +23,7 @@ const (
 )
 
 var (
-	targetStatuses        = []string{"needed", "deferred", "released"}
+	targetStatuses        = []string{"ignored", "needed", "pending", "deferred", "released"}
```
DmitriyLewen, line 26: I checked all posible statuses:
`[not-affected DNE released ignored needs-triage needed deferred pending]`

we need to add `needs-triage` (as `Unknown` status (see https://github.com/aquasecurity/trivy-db/blob/11edc670d14bcb450e2f02738a54eaae8a2744bd/pkg/vulnsrc/ubuntu/ubuntu.go#L231-L232))

But i think we can just skip `not-affected` and `DNE` statuses to avoid skipping new statuses (if ubuntu will add them)
@knqyf263 wdyt?

pkg/vulnsrc/ubuntu/ubuntu_test.go
```diff
@@ -33,6 +34,7 @@ func TestVulnSrc_Update(t *testing.T) {
 					Key: []string{"advisory-detail", "CVE-2020-1234", "ubuntu 18.04", "xen"},
 					Value: types.Advisory{
 						FixedVersion: "1.2.3",
+						Status:       types.StatusFixed,
```
DmitriyLewen, line 37: it should be empty, no?
https://github.com/aquasecurity/trivy-db/pull/407#discussion_r1648346360

pkg/vulnsrc/ubuntu/ubuntu.go
knqyf263: Thanks for understanding. Since this DB is frequently downloaded, the size is essential for us.

pkg/vulnsrc/ubuntu/ubuntu.go
skahn007gl: Thanks for spotting this.
type conflict resolved with var rename.

status was a conflicting var name.

pkg/vulnsrc/ubuntu/ubuntu.go
skahn007gl: Status `needs-triage` has been added into targetStatuses.

pkg/vulnsrc/ubuntu/ubuntu_test.go
skahn007gl: Test fixed.
Status should be empty with logic change to include it only when status is not fixed.

DmitriyLewen (commented):
fix linter error, please
https://github.com/aquasecurity/trivy-db/actions/runs/9608945100/job/26502649422?pr=407#step:4:26

pkg/vulnsrc/ubuntu/ubuntu.go
knqyf263: @DmitriyLewen `needs-triage` is "Under investigation". Trivy DB doesn't save such advisories.

DmitriyLewen: oh... right. comment about `needs-triage` confused me.
@skahn007gl can you exclude `needs-triage` from `targetStatuses` and remove [this comment](https://github.com/aquasecurity/trivy-db/blob/f6a5b03dc85c2147eb8195c8945177cd161ea2c5/pkg/vulnsrc/ubuntu/ubuntu.go#L231) to avoid confusing.

skahn007gl: @DmitriyLewen Done, test updated to.

DmitriyLewen: Thanks! fix linter error please

skahn007gl:
> fix linter error, please https://github.com/aquasecurity/trivy-db/actions/runs/9608945100/job/26502649422?pr=407#step:4:26

Done :)

DmitriyLewen (approved):
LGTM

@knqyf263 take a look, when you have time

knqyf263:
@skahn007gl Thanks for updating. At last, could you add a test with a non-fixed status? Adding a new advisory or updating [the existing one](https://github.com/aquasecurity/trivy-db/blob/b17897be045dd39c5fdfe053f8fd8ea643827202/pkg/vulnsrc/ubuntu/testdata/vuln-list/ubuntu/CVE-2020-1234.json).

knqyf263:
@skahn007gl Also, could you open a PR with this change in Trivy? It's better to test this change with Trivy before merging this PR.

- Add `Status` here and update unit tests
https://github.com/aquasecurity/trivy/blob/4369a19af771f81df141530bacdc8680e7120ac7/pkg/detector/ospkg/ubuntu/ubuntu.go#L100-L110
- Update integration tests
https://github.com/aquasecurity/trivy/blob/bbaf5952bc8059c537401bd4364e019eecd16c9a/integration/standalone_tar_test.go#L217-L233
- Update documentation in two places
https://github.com/aquasecurity/trivy/blob/648ead9553eb2cbeac90e3ef7330a70c352255ac/docs/docs/configuration/filtering.md#by-status
https://github.com/aquasecurity/trivy/blob/648ead9553eb2cbeac90e3ef7330a70c352255ac/docs/docs/coverage/os/ubuntu.md#status

Tips: Use [replace](https://go.dev/doc/modules/gomod-ref#:~:text=Use%20the%20replace%20directive%20to,to%20use%20the%20replacement%20path.) until this PR gets merged.

pkg/vulnsrc/ubuntu/testdata/vuln-list/ubuntu/CVE-2020-1234.json
```diff
@@ -19,10 +19,11 @@
       "bionic": {
         "Status": "released",
         "Note": "1.2.3"
-      },
-      "focal": {
-        "Status": "needs-triage",
-        "Note": ""
+      }
+    },
+    "trusty": {
```
knqyf263, line 24: The key must be a package name. Isn't `trusty` a codename?

skahn007gl: Ah yes it should be. I read `xen` as `xenial`, Will change the test data and get it pushed

skahn007gl:
> @skahn007gl Also, could you open a PR with this change in Trivy? It's better to test this change with Trivy before merging this PR.
> 
> * Add `Status` here and update unit tests
>   https://github.com/aquasecurity/trivy/blob/4369a19af771f81df141530bacdc8680e7120ac7/pkg/detector/ospkg/ubuntu/ubuntu.go#L100-L110
> * Update integration tests
>   https://github.com/aquasecurity/trivy/blob/bbaf5952bc8059c537401bd4364e019eecd16c9a/integration/standalone_tar_test.go#L217-L233
> * Update documentation in two places
>   https://github.com/aquasecurity/trivy/blob/648ead9553eb2cbeac90e3ef7330a70c352255ac/docs/docs/configuration/filtering.md#by-status
>   https://github.com/aquasecurity/trivy/blob/648ead9553eb2cbeac90e3ef7330a70c352255ac/docs/docs/coverage/os/ubuntu.md#status
> 
> Tips: Use [replace](https://go.dev/doc/modules/gomod-ref#:~:text=Use%20the%20replace%20directive%20to,to%20use%20the%20replacement%20path.) until this PR gets merged.

Hey @knqyf263 I am unsure what needs to change in the existing integration tests.
Is my understanding of the tests correct that it tests 2 scenarios?
 
1. Include all findings 
2. Include findings with fixes only

This includes the status field in the .golden test data.

I have made the other changes and raised a PR to Trivy :) 
[Trivy PR7020](https://github.com/aquasecurity/trivy/pull/7020)

pkg/vulnsrc/ubuntu/testdata/vuln-list/ubuntu/CVE-2020-1234.json
skahn007gl: test data updated and pushed.

knqyf263:
There is a DB fixture here.
https://github.com/aquasecurity/trivy/blob/8d618e48a2f1b60c2e4c49cdd9deb8eb45c972b0/integration/testdata/fixtures/db/ubuntu.yaml

We need to add a status here and ensure the status appears in the result.

skahn007gl:
> There is a DB fixture here. https://github.com/aquasecurity/trivy/blob/8d618e48a2f1b60c2e4c49cdd9deb8eb45c972b0/integration/testdata/fixtures/db/ubuntu.yaml
> 
> We need to add a status here and ensure the status appears in the result.
> 
> This bucket must also be updated if you want to add a new vulnerability. https://github.com/aquasecurity/trivy/blob/8d618e48a2f1b60c2e4c49cdd9deb8eb45c972b0/integration/testdata/fixtures/db/vulnerability.yaml
> 
> When adding a new bucket for Ubuntu, that needs to be added here. https://github.com/aquasecurity/trivy/blob/8d618e48a2f1b60c2e4c49cdd9deb8eb45c972b0/integration/testdata/fixtures/db/data-source.yaml#L137-L141

@knqyf263 How are the .golden files generated/created?

I have updated the DB fixture with an existing CVE for ubuntu 18.04 that would have a non fixed status, but i'm unable to trigger a failure on the existing ubuntu18.04 integration tests.

```yaml
- bucket: ubuntu 18.04
  pairs:
    - bucket: libspring-java
      pairs:
        - key: CVE-2022-22965
          value:
            Status: deferred
            ```

**Comment by knqyf263** [comment]
`deferred` is not our status.
https://github.com/aquasecurity/trivy-db/blob/9bdfe07de664866c0d889b2d1f8d6ecc11b640ad/pkg/types/status.go#L17-L26

**Comment by skahn007gl** [comment]
> `deferred` is not our status.
> 
> https://github.com/aquasecurity/trivy-db/blob/9bdfe07de664866c0d889b2d1f8d6ecc11b640ad/pkg/types/status.go#L17-L26

Corrected to `fix_deferred`, but still no failure in the test run.

I am assuming I need to update `ubuntu-1804.json.golden` to update the expected results for the tests linked https://github.com/aquasecurity/trivy/blob/bbaf5952bc8059c537401bd4364e019eecd16c9a/integration/standalone_tar_test.go#L217-L233

If this is the case, How are the .golden files generated/created?

Thanks for the help and guidance on this as I work through it.

**Comment by knqyf263** [comment]
>I am assuming I need to update ubuntu-1804.json.golden to update the expected results for the tests linked https://github.com/aquasecurity/trivy/blob/bbaf5952bc8059c537401bd4364e019eecd16c9a/integration/standalone_tar_test.go#L217-L233

No, the test should fail due to a mismatch with the golden file. You need to update the golden file once you confirm the test fails. Did you update all the buckets I listed above? It may be easier for us to debug if you push the change to your PR.

**Comment by skahn007gl** [comment]
> > I am assuming I need to update ubuntu-1804.json.golden to update the expected results for the tests linked https://github.com/aquasecurity/trivy/blob/bbaf5952bc8059c537401bd4364e019eecd16c9a/integration/standalone_tar_test.go#L217-L233
> 
> No, the test should fail due to a mismatch with the golden file. You need to update the golden file once you confirm the test fails. Did you update all the buckets I listed above? It may be easier for us to debug if you push the change to your PR in Trivy.

Thanks for confirming, I agree it will be easier to debug if i share my changes. I will update the PR shortly.

**Comment by skahn007gl** [comment]
@knqyf263 I have pushed the updated DB fixture and OS bucket.
I used an existing vulnerability.

Note the changes do not include the use of `replace` in `go.mod`, but i do use it locally.
`e.g: replace github.com/aquasecurity/trivy-db => /xxx/yyy/zzz/trivy-db`

**Comment by knqyf263** [comment]
You added Ubuntu 21.10, but the scanned image is not Ubuntu 21.10. The data is not used.

**Comment by skahn007gl** [comment]
> You added Ubuntu 21.10, but the scanned image is not Ubuntu 21.10. The data is not used.

Adjusted to ubuntu 16.04 as there is an image under `integration/testdata/fixtures/images`
Added test for `ubuntu 16.04` into `integration/standalone_tar_test.go`
note: Test does not use the correct `json.golden` as there is no existing one for `ubuntu 16.04`

Successfully generated a failed test result.

I am running the tests via `mage test:integration`(mage is new to me) or `go test -timeout 15m -v -tags=integration ./integration/...`
But i do not see any test artefacts generated, only pass/fail stat's.

Is there a way to generate the output from the comparison?.

I have pushed my updates.

**Comment by skahn007gl** [comment]
Hey @knqyf263 
Sorry for the delays on my side, I've had some personal life stuff crop up.

I have simplified the integration test to use as much as existing as possible to understand how these tests hang together, I believe I am still missing something here.

- Added a CVE to the db fixture
- Added the CVE to the vulnerability fixture
- As i re-used ubuntu 18.04, I did not add a new bucket to [data-source](https://github.com/aquasecurity/trivy/blob/8d618e48a2f1b60c2e4c49cdd9deb8eb45c972b0/integration/testdata/fixtures/db/data-source.yaml#L137-L141) 

When i run the integration test's, there is no test failure for `ubuntu-1804`

I added some dummy data to to `ubuntu-1804.json.golden` This induces an error, but the error indicates it did not expect the new vulnerability.

```
	-   Vulnerabilities: ([]types.DetectedVulnerability) (len=6) {
	+   Vulnerabilities: ([]types.DetectedVulnerability) (len=5) {
```

File: pkg/vulnsrc/ubuntu/ubuntu.go [hunk 16]
```diff
@@ -170,8 +170,12 @@ func defaultPut(dbc db.Operation, tx *bolt.Tx, advisory interface{}) error {
 			}
 
 			adv := types.Advisory{}
-			if status.Status == "released" {
+			normalised_status := StatusFromUbuntuStatus(status.Status)
```
DmitriyLewen, line 173: https://go.dev/doc/effective_go#mixed-caps
```suggestion
			normalisedStatus := StatusFromUbuntuStatus(status.Status)
```

DmitriyLewen: also you need to update this variable below :)

knqyf263: Yes, need to fix

DmitriyLewen (approved):
LGTM

pkg/vulnsrc/ubuntu/ubuntu.go
```diff
@@ -213,3 +217,17 @@ func SeverityFromPriority(priority string) types.Severity {
 		return types.SeverityUnknown
 	}
 }
+
+// StatusFromUbuntuStatus normalises Ubuntu status into common Trivy Types
+func StatusFromUbuntuStatus(status string) types.Status {
+	switch status {
+	case "ignored":
+		return types.StatusWillNotFix
```
knqyf263, line 225: I took another read of the document.
https://askubuntu.com/questions/1509705/cve-questions-about-status/1509706#1509706

>"ignored" support has ended for one reason or another/ end-of-life.

It looks like Ubuntu doesn't determine if the package is affected. So, if I understand correctly, we should not show this type of vulnerability.

skahn007gl: We consider this an intentional decision that is `vendor will not fix`, this status is important in our downstream workflow and was included for this reason.

knqyf263: "Will not fix" means "this package is affected by this flaw on this platform, but there is currently no intention to fix it." It's a Red Hat definition, but we follow it.
https://access.redhat.com/blogs/product-security/posts/2066793

skahn007gl: @knqyf263 Are you happy to proceed with this?
What do you need from me to get this merged?

jhebden-gl: @skahn007gl - it seems that treating ignored as wontfix is the issue here, as ignored/EOL findings are typically not ingested to reduce Trivy DB size. I think we can just skip ingesting ignored, we still have the "notaffected" status to determine if the outcome of analysis is that Ubuntu decided not to fix it. For EOL stuff, we can work that out another way on our end, and treat findings for EOL releases as wontfix if they don't have another status.

knqyf263: Trivy is designed to minimize false positives as much as possible. EOL means unknown and is possibly a false positive, which is why we currently do not detect it.

skahn007gl: While i see the reasoning behind this and do not disagree, I am happy to follow guidance and remove the status for ubuntu.

@knqyf263 let me know if you feel strongly on removing `ignored` for Ubuntu and i will make the changes.

I would like to highlight the `ignored` is ingested for Debian in `pkg/vulnsrc/debian.go:682`
```
func newStatus(s string) types.Status {
	switch strings.ToLower(s) {
	// "end-of-life" is considered as vulnerable
	// e.g. https://security-tracker.debian.org/tracker/CVE-2022-1488
	case "no-dsa", "unfixed":
		return types.StatusAffected
	case "ignored":
		return types.StatusWillNotFix
	case "postponed":
		return types.StatusFixDeferred
	case "end-of-life":
		return types.StatusEndOfLife
	}
	return types.StatusUnknown
} 
```

knqyf263: >I would like to highlight the ignored is ingested for Debian in pkg/vulnsrc/debian.go:682

We believe "ignored" in Debian differs from "ignored" in Ubuntu. They use the same term by chance. Please let us know if you find a document in which Debian also uses "ignored" for EOL.

skahn007gl: Hey @knqyf263, @jhebden-gl 
I have removed `ignored` status as we can make an assumption on `wontfix` by presence of data (Affected), Deferred status, No fix version.

knqyf263 (approved):

skahn007gl:
Hey @DmitriyLewen @knqyf263 Thank you for time and efforts on this work. We have sourced this data another way and no longer require these PR's.

PR #412, opened 2024-06-28
refactor(metadata): pass db dir

knqyf263:
## Current Situation
Currently, the [db](https://github.com/aquasecurity/trivy-db/blob/27fa703f5ffa3893729ffc84a9b096f3406eebf2/pkg/db/db.go#L89) and [metadata](https://github.com/aquasecurity/trivy-db/blob/27fa703f5ffa3893729ffc84a9b096f3406eebf2/pkg/metadata/metadata.go#L38) packages are passed a cache directory, within which they create subdirectories like `$CACHE_DIR/db`. This results in database and metadata files being stored at paths such as:

- `$CACHE_DIR/db/trivy.db`
- `$CACHE_DIR/db/metadata.json`

## Problem
While the metadata implementation and other aspects are common and potentially reusable for Java DB and other databases, the hard-coded `db/` subdirectory makes it difficult to reuse these packages effectively.

## Proposed Solution
To improve reusability, I propose to remove the hard-coded `db/` subdirectory and instead allow the DB directory to be passed from outside. This change will result in file paths like:

- `$DB_DIR/trivy.db`
- `$DB_DIR/metadata.json`

Where `$DB_DIR` can be set to any desired path, including but not limited to the current `$CACHE_DIR/db`. 

As a CLI flag, the value passed as `--output-dir` is passed directly to $DB_DIR.

- `$OUTPUT_DIR/trivy.db`
- `$OUTPUT_DIR/metadata.json`

## Note
We need to update Trivy as it also affects Trivy.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,8 +17,9 @@ coverage.txt
 
 **/.DS_Store
 
-assets
-cache
+/assets
+/cache
+/out
 
 # trivy-db Outputs
 trivy-db
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,8 @@
 SHELL=/bin/bash
 LDFLAGS=-ldflags "-s -w"
+CACHE_DIR=cache
+OUT_DIR=out
+ASSET_DIR=assets
 
 GOPATH=$(shell go env GOPATH)
 GOBIN=$(GOPATH)/bin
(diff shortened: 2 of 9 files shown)


DmitriyLewen (approved):
LGTM

1 question:
in source code you use `OUT_DIR`, but in PR description `DB_DIR`.
Is this the same thing?

knqyf263:
>in source code you use OUT_DIR, but in PR description DB_DIR.
>Is this the same thing?

Yes. [DB dir](https://github.com/aquasecurity/trivy-db/blob/01829d77758c5634c0825254f7d2ef19be56fb65/pkg/db/db.go#L58) is used as an argument because it is a DB directory, but I thought [output dir](https://github.com/aquasecurity/trivy-db/blob/01829d77758c5634c0825254f7d2ef19be56fb65/pkg/app.go#L42-L46) would be easier to understand as a CLI flag of `trivy-db build`, so that is what I did. I don't mind changing it to the same name.

BTW, I'm afraid of breaking something. I built a database locally and confirmed it worked as expected, but could you double-check it?

knqyf263:
>OUT_DIR is good for me. But I suggest using OUT_DIR in the PR description so that we don't get confused if we look at that PR later.

Updated.

>I have already build/compress/compact db locally. It works correctly for me too.

Awesome. Thanks!

PR #413, opened 2024-06-28
pkg: replace `utils/strings/InSlice` func with `slices.Contains`

### LINKED ISSUE #245: replace `utils/strings/InSlice` to `slices.Contains`
Now we don't need this deprecated function `InSlice`:

https://github.com/aquasecurity/trivy-db/blob/main/pkg/utils/strings/strings.go#L28-L35

mohamedawnallah:
## Description

In this commit, we replace `utils/strings/InSlice` function with the built-in `slices.Contains` function which is supported in go 1.18+ and it would work with the current project go version `1.19` as appears in the `go.mod` file.

Closes #245

## How has this been tested?

I've run the test cases using `make test` and all of them passed.

### DIFF
--- a/pkg/utils/strings/strings.go
+++ b/pkg/utils/strings/strings.go
@@ -25,15 +25,6 @@ func IsInt(s string) bool {
 	return err == nil
 }
 
-func InSlice(a string, list []string) bool {
-	for _, b := range list {
-		if b == a {
-			return true
-		}
-	}
-	return false
-}
-
 func Merge(a, b []string) []string {
 	uniq := map[string]struct{}{}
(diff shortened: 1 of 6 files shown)


knqyf263 (approved):

knqyf263:
I think `slices` was added to the standard package in Go 1.21.

mohamedawnallah:
Ah sorry my bad. Your are right:
https://pkg.go.dev/slices?tab=versions

knqyf263:
opened https://github.com/aquasecurity/trivy-db/pull/415

knqyf263:
Thanks

PR #420, opened 2024-07-11
feat: fetch k8s cves from kubernetes-sigs/cve-feed-osv

chen-keinan:
## Description

The purpose of this change is to collect kubernetes cves data from kubernetes-sigs/cve-feed-osv which is  relable and has osv format, instead of the existing k8s collector

- Related https://github.com/kubernetes-sigs/cve-feed-osv/issues/1


DmitriyLewen (commented):
I thought about `vuln-list-k8s` repository.
After merging this PR, we may want to delete/stop updating this repository.
@chen-keinan @knqyf263 wdyt?

knqyf263:
Yes, we should stop vuln-list-k8s and archive it.

knqyf263 (approved):
LGTM

DmitriyLewen (approved):
LGTM

PR #428, opened 2024-08-23
refactor(redhat-oval): return an error if the CPE is not found

DmitriyLewen:
## Description
`Get()` function returns an error if the CPE is not found

### DIFF
--- a/pkg/vulnsrc/redhat-oval/redhat-oval.go
+++ b/pkg/vulnsrc/redhat-oval/redhat-oval.go
@@ -248,6 +248,10 @@ func (vs VulnSrc) Get(pkgName string, repositories, nvrs []string) ([]types.Advi
 		return nil, xerrors.Errorf("CPE convert error: %w", err)
 	}
 
+	if len(cpeIndices) == 0 {
+		return nil, xerrors.Errorf("No CPE indices found")
+	}
+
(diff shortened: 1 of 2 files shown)


pkg/vulnsrc/redhat-oval/redhat-oval.go
```diff
@@ -248,6 +248,10 @@ func (vs VulnSrc) Get(pkgName string, repositories, nvrs []string) ([]types.Advi
 		return nil, xerrors.Errorf("CPE convert error: %w", err)
 	}
 
+	if len(cpeIndices) == 0 {
+		return nil, xerrors.Errorf("No CPE indices found")
```
knqyf263, line 252:
```suggestion
		return nil, xerrors.Errorf("no CPE indices found")
```

knqyf263: And this error message is unclear. What do you think adding this issue in the error message?
https://github.com/aquasecurity/trivy-db/issues/435

DmitriyLewen: It make sense.
Updated in https://github.com/aquasecurity/trivy-db/pull/428/commits/83716e83d10894e13a41236165c2a389e42c47c3

knqyf263 (approved):

Issue #429, opened 2024-08-25
add bolt options to db.Init to allow multiple readers on the same db


orizerah:
Receive `bbolt.Options` in `db.Init` function to allow Readonly instances

PR #430, opened 2024-08-25
feat(bolt-options): add bolt options to db.Init

orizerah:
https://github.com/aquasecurity/trivy-db/issues/429

add bbolt.Options to db.Init arguments for a more customizable db

### DIFF
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -55,7 +55,7 @@ type Operation interface {
 type Config struct {
 }
 
-func Init(dbDir string) (err error) {
+func Init(dbDir string, dbOpts *bolt.Options) (err error) {
 	if err = os.MkdirAll(dbDir, 0700); err != nil {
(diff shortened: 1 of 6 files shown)


pkg/db/db.go
```diff
@@ -55,7 +55,7 @@ type Operation interface {
 type Config struct {
 }
 
-func Init(dbDir string) (err error) {
+func Init(dbDir string, dbOpts *bolt.Options) (err error) {
```
knqyf263, line 58: Trivy doesn't need `bolt.Options`. What if using Functional Option Pattern?
e.g. https://github.com/aquasecurity/trivy-db/blob/ed690fa5e757e97982e86d44230841e6e95fd793/pkg/vulndb/db.go#L34-L38

DmitriyLewen (approved):
LGTM

knqyf263 (approved):

PR #436, opened 2024-09-10
fix: do not use a pointer receiver for Status.MarshalJSON

knqyf263:
## Description
In MarshalJSON, using a pointer receiver was not supposed to be marshalled correctly, but in Trivy, the output was correct (I did not investigate the reason in detail), and furthermore, to suppress the following warning, I used a pointer.

>Struct Status has methods on both value and pointer receivers. Such usage is not recommended by the Go Documentation. 

However, it seems that `Status` was only correctly marshalled by chance because [DetectedVulnerability](https://github.com/aquasecurity/trivy/blob/7aea79dd93cfb61453766dbbb2e3fc0fbd317852/pkg/types/report.go#L114) was a slice. We actually found a case where `DetectedVulnerability` was [not correctly marshalled](https://github.com/aquasecurity/trivy/pull/7463) when `DetectedVulnerability` was not a slice, so this PR will correct this.

Translated with DeepL.com (free version)

### DIFF
--- a/pkg/types/status.go
+++ b/pkg/types/status.go
@@ -58,7 +58,7 @@ func (s *Status) Index() int {
 	return int(*s)
 }
 
-func (s *Status) MarshalJSON() ([]byte, error) {
+func (s Status) MarshalJSON() ([]byte, error) {
 	return json.Marshal(s.String())
 }
 


DmitriyLewen (approved):
I tested this change in https://github.com/aquasecurity/trivy/pull/7463
`status` field in `ExperimentalModifiedFindings` is marshaled correctly

pkg/types/status.go
```diff
@@ -58,7 +58,7 @@ func (s *Status) Index() int {
 	return int(*s)
 }
 
-func (s *Status) MarshalJSON() ([]byte, error) {
+func (s Status) MarshalJSON() ([]byte, error) {
```
afdesk, line 61: Now struct Status has methods on both value (`UnmarshalJSON`) and pointer (`String`, `Index` and `UnmarshalJSON`) receivers.

Should we change receivers for other functions too?

DmitriyLewen: But in this case we can't set Status in `UnmarshalJSON`, can we?

afdesk: it seems it's a common practice to use `value` receiver for UnmarshalJSON and `pointer` for other methods:
https://github.com/CycloneDX/cyclonedx-go/blob/6f53207eded10e9e2a362772b253621840e7ad94/cyclonedx_json.go

afdesk: in the docs
https://pkg.go.dev/encoding/json#example-package-CustomMarshalJSON

knqyf263: Yes, MarshalJSON and UnmarshalJSON should be excluded from the linter rule.
https://youtrack.jetbrains.com/issue/GO-13587

PR #440, opened 2024-09-24
ci: add publishing into Public ECR

afdesk:
This PR adds publishing the DB to ECR public registry as well as GHCR.


knqyf263 (approved):

knqyf263:
We're carefully assessing quotas in ECR Public now.
https://docs.aws.amazon.com/AmazonECR/latest/public/public-service-quotas.html

PR #444, opened 2024-10-01
ci: set to download DB daily

knqyf263:
This PR changes to have clients download the DB once a day as we are facing [GHCR rate limits](https://github.com/aquasecurity/trivy/discussions/7538). However, the DB will be built [once every 6 hours](https://github.com/aquasecurity/trivy-db/blob/cfa337a1088bbcee598ab93656c83fe6b9acb946/.github/workflows/cron.yml#L5).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,7 +77,7 @@ db-fetch-langs:
 
 .PHONY: db-build
 db-build: trivy-db
-	./trivy-db build --cache-dir ./$(CACHE_DIR) --output-dir ./$(OUT_DIR) --update-interval 6h
+	./trivy-db build --cache-dir ./$(CACHE_DIR) --output-dir ./$(OUT_DIR) --update-interval 24h
 
 .PHONY: db-compact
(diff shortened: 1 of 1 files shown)


DmitriyLewen (approved):

knqyf263:
I'm testing this change in https://github.com/aquasecurity/trivy-db/actions/runs/11138880131.

DmitriyLewen:
I checked this locally:
it works correctly:
```bash
➜  cat metadata.json 
{"Version":2,"NextUpdate":"2024-10-03T04:43:04.211896Z","UpdatedAt":"2024-10-02T04:43:04.211897Z","DownloadedAt":"0001-01-01T00:00:00Z"}
```

knqyf263:
Great. Since this change impacts all users, I'd like to re-test it on GitHub Actions just in case.

knqyf263:
I locally confirmed the new DB worked with Trivy.

PR #450, opened 2024-10-15
fix(cron): pin `ubuntu-24.04` runner and use installed (bump to v1.2.0) `oras` version

DmitriyLewen:
## Description
`ubuntu-24.04` runner is now `latest` runner.

But `ubuntu-24.04` runner doesn't have preinstalled `oras` - https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#cli-tools

Before using `ubuntu-24.04` - we used `ubuntu-22.04`.
`ubuntu-22.04` contains [oras v1.2.0](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#cli-tools).

Changes of this PR:
- pin `ubuntu-24.04` runner (to avoid problem with next update `latest` runner).
- bump installed `oras` version to `v1.2.0` (as in `ubuntu-22.04` runner)
- use installed `oras`.

test run - https://github.com/DmitriyLewen/trivy-db/actions/runs/11339374413/job/31534396989

### DIFF
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -106,7 +106,7 @@ jobs:
             for tag in "${tags[@]}"; do
               echo "Pushing artifact with tag: ${tag}"
           
-              if oras push --artifact-type application/vnd.aquasec.trivy.config.v1+json \
+              if ./oras push --artifact-type application/vnd.aquasec.trivy.config.v1+json \
                 "${full_registry_url}:${tag}" \
                 db.tar.gz:application/vnd.aquasec.trivy.db.layer.v1.tar+gzip; then
                 echo "Successfully pushed to ${full_registry_url}:${tag}"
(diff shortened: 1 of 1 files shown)


knqyf263:
We were not using `oras` locally installed accidentally, and the error started occurring as ubuntu-24.04 deleted `oras` from the installed tools, right?

DmitriyLewen:
> We were not using oras locally installed accidentally

Right.

> error started occurring as ubuntu-24.04 deleted oras from the installed tools, right?

Almost right. `ubuntu-22.04` contains `oras`. But `ubuntu-24.04` doesn't contain `oras`.
GitHub didn't remove `oras`. GitHub started using `ubuntu-24.04` as `latest`.

knqyf263:
Yes, I meant GitHub deleted oras starting from ubuntu-24.04. So, this PR title looks like it just updates the oras version, but it actually fixes some problems. Can you update the title and the description to clarify that?

DmitriyLewen:
@knqyf263 Done

knqyf263 (approved):

PR #452, opened 2024-10-15
fix(oracle): split version flavors

DmitriyLewen:
## Description
Oracle Linux uses `normal`, `fips` (e.g. `3.6.16-4.0.1.el8_fips`) and `ksplice` (e.g. `2:2.28-151.0.1.ksplice2.el8`) version flavors.
We need to separate these versions so that we only compare version with same flavors.
See https://github.com/aquasecurity/trivy-db/issues/220 for more details.

### PR info
We are currently unable to change the database schema, so this PR is workaround to keep multiple versions for single CVE.
We currently use [FixedVersion](https://github.com/aquasecurity/trivy-db/blob/358e1d6974bba8e4031580fd490b82f7cac5080c/pkg/types/types.go#L119-L121) for OS advisories.
But in this case [PatchedVersions](https://github.com/aquasecurity/trivy-db/blob/358e1d6974bba8e4031580fd490b82f7cac5080c/pkg/types/types.go#L126) was used.

It adds options to save all flavors for single advisory.

To save backward compatibility - we only keep version with `normal` flavor in `FixedVersion`.

### DB sizes:
before:
```bash
-rw-r--r--@ 1 dmitriy  staff    53M Oct  8 16:51 db.tar.gz
-rw-------@ 1 dmitriy  staff   507M Oct  8 16:50 trivy.db

```

after
```bash
-rw-r--r--@ 1 dmitriy  staff    52M Oct 14 17:48 db.tar.gz
-rw-------@ 1 dmitriy  staff   523M Oct 14 17:47 trivy.db
```

## Related PRs
- https://github.com/aquasecurity/trivy-db/pull/221
- https://github.com/aquasecurity/trivy-db/pull/222

### DIFF
--- a/pkg/vulnsrc/oracle-oval/oracle-oval_test.go
+++ b/pkg/vulnsrc/oracle-oval/oracle-oval_test.go
@@ -421,6 +421,9 @@ func TestVulnSrc_Get(t *testing.T) {
 				{
 					VulnerabilityID: "ELSA-2019-1145",
 					FixedVersion:    "32:9.11.4-17.P2.el8_0",
+					PatchedVersions: []string{
+						"32:9.11.4-17.P2.el8_0",
+					},
 				},
 			},
 		},
--- a/pkg/vulnsrc/oracle-oval/oracle-oval_test.go
+++ b/pkg/vulnsrc/oracle-oval/oracle-oval_test.go
@@ -39,24 +39,36 @@ func TestVulnSrc_Update(t *testing.T) {
 					Key: []string{"advisory-detail", "CVE-2007-0493", "Oracle Linux 5", "bind-devel"},
 					Value: types.Advisory{
 						FixedVersion: "30:9.3.3-8.el5",
+						PatchedVersions: []string{
+							"30:9.3.3-8.el5",
+						},
 					},
 				},
 				{
 					Key: []string{"advisory-detail", "CVE-2007-0494", "Oracle Linux 5", "bind-devel"},
 					Value: types.Advisory{
 						FixedVersion: "30:9.3.3-8.el5",
+						PatchedVersions: []string{
+							"30:9.3.3-8.el5",
+						},
 					},
 				},
 				{
 					Key: []string{"advisory-detail", "CVE-2007-0493", "Oracle Linux 5", "bind-sdb"},
 					Value: types.Advisory{
 						FixedVersion: "30:9.3.3-8.el5",
+						PatchedVersions: []string{
+							"30:9.3.3-8.el5",
+						},
 					},
 				},
 				{
 					Key: []string{"advisory-detail", "CVE-2007-0494", "Oracle Linux 5", "bind-sdb"},
 					Value: types.Advisory{
 						FixedVersion: "30:9.3.3-8.el5",
+						PatchedVersions: []string{
+							"30:9.3.3-8.el5",
+						},
 					},
 				},
 				{
@@ -117,48 +129,72 @@ func TestVulnSrc_Update(t *testing.T) {
 					Key: []string{"advisory-detail", "CVE-2018-1094", "Oracle Linux 6", "kernel-uek-doc"},
 					Value: types.Advisory{
 						FixedVersion: "4.1.12-124.24.3.el6uek",
+						PatchedVersions: []string{
+							"4.1.12-124.24.3.el6uek",
+						},
 					},
 				},
 				{
 					Key: []string{"advisory-detail", "CVE-2018-19824", "Oracle Linux 6", "kernel-uek-doc"},
 					Value: types.Advisory{
 						FixedVersion: "4.1.12-124.24.3.el6uek",
+						PatchedVersions: []string{
+							"4.1.12-124.24.3.el6uek",
+						},
 					},
 				},
 				{
 					Key: []string{"advisory-detail", "CVE-2018-1094", "Oracle Linux 6", "kernel-uek-firmware"},
 					Value: types.Advisory{
 						FixedVersion: "4.1.12-124.24.3.el6uek",
+						PatchedVersions: []string{
+							"4.1.12-124.24.3.el6uek",
+						},
 					},
 				},
 				{
 					Key: []string{"advisory-detail", "CVE-2018-19824", "Oracle Linux 6", "kernel-uek-firmware"},
 					Value: types.Advisory{
 						FixedVersion: "4.1.12-124.24.3.el6uek",
+						PatchedVersions: []string{
+							"4.1.12-124.24.3.el6uek",
+						},
 					},
 				},
 				{
 					Key: []string{"advisory-detail", "CVE-2018-1094", "Oracle Linux 7", "kernel-uek-doc"},
 					Value: types.Advisory{
 						FixedVersion: "4.1.12-124.24.3.el7uek",
+						PatchedVersions: []string{
+							"4.1.12-124.24.3.el7uek",
+						},
 					},
 				},
 				{
 					Key: []string{"advisory-detail", "CVE-2018-19824", "Oracle Linux 7", "kernel-uek-doc"},
 					Value: types.Advisory{
 						FixedVersion: "4.1.12-124.24.3.el7uek",
+						PatchedVersions: []string{
+							"4.1.12-124.24.3.el7uek",
+						},
 					},
 				},
 				{
 					Key: []string{"advisory-detail", "CVE-2018-1094", "Oracle Linux 7", "kernel-uek-firmware"},
 					Value: types.Advisory{
 						FixedVersion: "4.1.12-124.24.3.el7uek",
+						PatchedVersions: []string{
+							"4.1.12-124.24.3.el7uek",
+						},
 					},
 				},
 				{
 					Key: []string{"advisory-detail", "CVE-2018-19824", "Oracle Linux 7", "kernel-uek-firmware"},
 					Value: types.Advisory{
 						FixedVersion: "4.1.12-124.24.3.el7uek",
+						PatchedVersions: []string{
+							"4.1.12-124.24.3.el7uek",
+						},
 					},
 				},
 				{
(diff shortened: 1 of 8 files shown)


knqyf263:
Have you considered using ELSA-IDs for keys, which would require a modification on the Trivy side to also display CVE-IDs, but I don't think it would be that bad since the old Trivy only displays ELSA-IDs instead of CVE-IDs.

knqyf263:
I suggested the same before 😄 
https://github.com/aquasecurity/trivy-db/pull/221#issuecomment-1167366007

DmitriyLewen:
I'm not sure about that.
1. ~~I can check, but I'm not sure we'll gain in DB size for this solution, because we'll need to create separate `Entry` for each CVE-ID (or use VendorIDs to merge CVEs with the same versions).~~
We can try to use [RedHat advisory](https://github.com/aquasecurity/trivy-db/blob/c423540993103ecc4618ccd9939297b9d22de39f/pkg/vulnsrc/redhat-oval/types.go#L173-L183) with `fixedVersion` as array.
If it is okay for you - i can try this and detect DB size
2. We'll need to add logic to remove 2 (or more) fixed versions (e.g. https://linux.oracle.com/cve/CVE-2014-7169.html contains 2 ELSA for `bash` (OL 5) => 2 fixed versions). 
So Trivy should solve this problem (which means Trivy will waste resources for such cases).

knqyf263:
>We can try to use [RedHat advisory](https://github.com/aquasecurity/trivy-db/blob/c423540993103ecc4618ccd9939297b9d22de39f/pkg/vulnsrc/redhat-oval/types.go#L173-L183) with fixedVersion as array.

Please let me think about it.

I'm now confused about FIPS. [This adviosry](https://linux.oracle.com/errata/ELSA-2022-6854.html) seems relevant to FIPS according to the description, but the version doesn't contain `_fips`.

```
gnutls
[3.7.6-12]
- fips: mark PBKDF2 with short key and output sizes non-approved
- fips: only mark HMAC as approved in PBKDF2
- fips: mark gnutls_key_generate with short key sizes non-approved
- fips: fix checking on hash algorithm used in ECDSA
- fips: preserve operation context around FIPS selftests API
```

Do you have any ideas?

DmitriyLewen:
hm... i couldn't to think explanation for that...
This is a changelog, so there may be typos...

[ELSA-2022-9221](https://linux.oracle.com/errata/ELSA-2022-9221.html) has `fips` version, but doesn't have `fips` prefies:
```
[3.6.16-4.0.1_fips]
- Allow RSA keygen with modulus sizes bigger than 3072 bits and validate the seed length
as defined in FIPS 186-4 section B.3.2 [Orabug: 33200526]
- Allow bigger known RSA modulus sizes when calling
rsa_generate_fips186_4_keypair directly [Orabug: 33200526]
- Change Epoch from 1 to 10
```

knqyf263:
[This advisory](https://linux.oracle.com/errata/ELSA-2024-1784.html) is mixing up normal and fips flavors.

DmitriyLewen:
> [This vulnerability](https://linux.oracle.com/cve/CVE-2022-2509.html) has no ELSA for FIPS. Does it mean FIPS gnutls is not affected?
It looks like the FIPS gnutls shares the code with the normal gnutls.

Take a look - https://github.com/aquasecurity/trivy/issues/1967#issuecomment-1092987400

> Does it mean FIPS gnutls is not affected?

IIUC you are right (or there is no fix for FIPS  gnutls)

pkg/vulnsrc/oracle-oval/oracle-oval_test.go
```diff
@@ -39,24 +39,36 @@ func TestVulnSrc_Update(t *testing.T) {
 					Key: []string{"advisory-detail", "CVE-2007-0493", "Oracle Linux 5", "bind-devel"},
 					Value: types.Advisory{
 						FixedVersion: "30:9.3.3-8.el5",
+						PatchedVersions: []string{
+							"30:9.3.3-8.el5",
```
knqyf263, line 43: We can use `FixedVersion` for normal flavor. Do we need to store this version in `PatchedVersions`? It helps reduce the size.

DmitriyLewen: hm... after adding arches (#453) this logic will be confusing (because we will only keep  normal version for `x86_64` arch in `FixedVersion`) 
But i will check your solution to understand how much space we can save.

knqyf263: Actually, I'm wondering how much using `PatchedVersions` helps from the size perspective after migrating to `Advisories` in #453.

knqyf263: I'll write an example. Just a moment.

DmitriyLewen: But I used `PatchedVersions` in #453 to avoid creating a new `Entry` for each flavor.

knqyf263: 1. `PatchedVersions`

<details><summary>PatchedVersions</summary>
<p>

```
types.Advisories{
	FixedVersion: "3.6.16-4.el8",
	Entries: []types.Advisory{
		{
			Arches: []string{
				"aarch64",
				"x86_64",
			},
			PatchedVersions: []string{
				"10:3.6.16-4.0.1.el8_fips",
				"3.6.16-4.el8",
			},
		},
	},
},
```

</p>
</details> 

2. `FixedVersion`

<details><summary>FixedVersion</summary>
<p>

```
types.Advisories{
	FixedVersion: "3.6.16-4.el8",
	Entries: []types.Advisory{
		{
			FixedVersion:	"10:3.6.16-4.0.1.el8_fips",
			Arches: []string{
				"aarch64",
				"x86_64",
			},
		},
		{
			FixedVersion:	"3.6.16-4.el8",
			Arches: []string{
				"aarch64",
				"x86_64",
			},
		},
	},
},
```

</p>
</details>  


One advantage of Option 2 is its simplicity of implementation. In the case of Option 1, the implementation is a little more complex as the advisory needs to be separated for different arches and put together for equal arches.

<details><summary>PatchedVersions with different arches</summary>
<p>

```
types.Advisories{
	FixedVersion: "3.6.16-4.el8",
	Entries: []types.Advisory{
		{
			Arches: []string{
				"aarch64",
				"x86_64",
			},
			PatchedVersions: []string{
				"10:3.6.16-4.0.1.el8_fips",
			},
		},
		{
			Arches: []string{
				"x86_64",
			},
			PatchedVersions: []string{
				"3.6.16-4.el8",
			},
		},
	},
},
```

</p>
</details> 

In the above example of `FixedVersion`, the arch is clearly redundant, but this could be reduced by leaving it empty if the advisory applies to all arches.

<details><summary>FixedVersion with omitted arches</summary>
<p>

```
types.Advisories{
	FixedVersion: "3.6.16-4.el8",
	Entries: []types.Advisory{
		{
			FixedVersion:	"10:3.6.16-4.0.1.el8_fips",
		},
		{
			FixedVersion:	"3.6.16-4.el8",
			Arches: []string{
				"aarch64",
				"x86_64",
			},
		},
	},
},
```

</p>
</details>  

If the size is still too large without using `PatchedVersions`, then go that way. I also experimented with various ways to simplify the implementation when adding support for Red Hat OVALv2, but ended up with a complex implementation because of the huge difference in database size.

DmitriyLewen: >  if the advisory applies to all arches.

I am not sure that there are cases when ELSA contains all arches:
```bash
➜ grep -rh "Oracle Linux arch is" | sed -e 's/^[ \t]*//' | sort | uniq
"Comment": "Oracle Linux arch is aarch64"
"Comment": "Oracle Linux arch is i386"
"Comment": "Oracle Linux arch is i686"
"Comment": "Oracle Linux arch is ia64"
"Comment": "Oracle Linux arch is sparc64"
"Comment": "Oracle Linux arch is x86_64"
```

> If the size is still too large without using PatchedVersions, then go that way.

I got your point.
I'll check it and write to you.

knqyf263: I saw only aarch64 and x86_64 in recent advisories. I guess others used to be supported in older versions.

DmitriyLewen: 2024 year also contains `i686`:
```
➜  2024 grep -rh "Oracle Linux arch is" | sed -e 's/^[ \t]*//' | sort | uniq
"Comment": "Oracle Linux arch is aarch64"
"Comment": "Oracle Linux arch is i686"
"Comment": "Oracle Linux arch is x86_64"
```

knqyf263: The intention is to reduce unnecessary data storage by leaving the most common cases empty and only adding values for special cases, so for example, empty could mean ‘aarch64’ and ‘x86_64’. This is a bit confusing, but I think it can be solved by writing it clearly in the comments.

knqyf263: Hmm. I have no idea why it can happen. Could you pick up some key/value pairs and compare them? The `FixedVersion` approach should be larger...

DmitriyLewen: I've done this before, but it seems to be correct.
I can only see one explanation:
most advisories contain only 1 `fixedVersion`.
so for these advisories DB contains only one row instead of an array with one element
e.g.:
```
{"Entries":[{"Arches":["aarch64","x86_64"],"FixedVersion":"2.1.3-4.el9_1"}],"FixedVersion":"2.1.3-4.el9_1"}
{"Entries":[{"Arches":["aarch64","x86_64"],"PatchedVersions":["2.1.3-4.el9_1"]}],"FixedVersion":"2.1.3-4.el9_1"}
```

we get a smaller size for `patchedVersions` only if there are multiple versions:
```
{"Entries":[{"Arches":["aarch64","x86_64"],"FixedVersion":"2.34-60.0.3.el9_2.7"},{"Arches":["aarch64","x86_64"],"FixedVersion":"2:2.34-60.0.3.ksplice1.el9_2.7"}],"FixedVersion":"2.34-60.0.3.el9_2.7"} 
{"Entries":[{"Arches":["aarch64","x86_64"],"PatchedVersions":["2.34-60.0.3.el9_2.7","2:2.34-60.0.3.ksplice1.el9_2.7"]}],"FixedVersion":"2.34-60.0.3.el9_2.7"}
```
but it seems to save more using fixed versions

knqyf263: It is counter-intuitive, but if that is the result, we have to accept it. It means that measurement is important.

DmitriyLewen: > It is counter-intuitive,

I wrote a code to compare CVE from 2 databases.
so I will list all CVE for both databases and recheck them manually.

DmitriyLewen: I saved all CVE values ​​from 2 databases and checked them.

The number of `ksplice` and `fips` versions is small (these values ​​are for OL 9):
```bash
➜ cat versions.txt | wc -l                         
   71781
➜ cat versions.txt | grep -E 'ksplice|fips' | wc -l
    4136
```

example for normal version and ksplice:
```
Oracle Linux 9/glib2-tests/CVE-2024-34397 - {"Entries":[{"Arches":["aarch64","x86_64"],"PatchedVersions":["2.68.4-14.el9_4.1"]}],"FixedVersion":"2.68.4-14.el9_4.1"}
Oracle Linux 9/glib2-tests/CVE-2024-34397 - {"Entries":[{"Arches":["aarch64","x86_64"],"FixedVersion":"2.68.4-14.el9_4.1"}],"FixedVersion":"2.68.4-14.el9_4.1"}
Oracle Linux 9/glibc/CVE-2023-4527 - {"Entries":[{"Arches":["aarch64","x86_64"],"PatchedVersions":["2.34-60.0.3.el9_2.7","2:2.34-60.0.3.ksplice1.el9_2.7"]}],"FixedVersion":"2.34-60.0.3.el9_2.7"}
Oracle Linux 9/glibc/CVE-2023-4527 - {"Entries":[{"Arches":["aarch64","x86_64"],"FixedVersion":"2.34-60.0.3.el9_2.7"},{"Arches":["aarch64","x86_64"],"FixedVersion":"2:2.34-60.0.3.ksplice1.el9_2.7"}],"FixedVersion":"2.34-60.0.3.el9_2.7"}
```

For each entry with only one version, we save a few bytes. 
This saves more than using an array for the `ksplice` and `fips` versions.

knqyf263: Great research! So, we would go for `FixedVersion`, right?

knqyf263: The `PatchedVersions` method needs a more complicated implementation as it needs to aggregate entries, but the data size will be bigger. There is no reason to go with that.

knqyf263: OK, so are you going to update this PR?

DmitriyLewen: I think we don't need to add flavors and arches in one PR.
So I will update this PR, after merging this PR we will add arches.

knqyf263: Agree. We should add changes step by step.

DmitriyLewen:
close this PR in favor #221

PR #471, opened 2024-11-18
fix: set trivy.db file to be world-readable

lyoung-confluent:
Currently the `trivy.db` file in the OCI artifact is set to be readable only by the `runner` user that created the file, unlike the `metadata.json` file:
```
$ crane export public.ecr.aws/aquasecurity/trivy-db:2 > trivy-db.tar
$ tar -tvf trivy-db.tar
-rw-------  0 runner docker 564006912 Nov 18 10:17 trivy.db
-rw-r--r--  0 runner docker       143 Nov 18 10:17 metadata.json
```
This is different from the `trivy-java.db` file:
```
$ crane export public.ecr.aws/aquasecurity/trivy-java-db:1 > trivy-java-db.tar
$ tar -tvf trivy-java-db.tar
-rw-r--r--  0 runner docker 1122304000 Nov 17 19:41 trivy-java.db
-rw-r--r--  0 runner docker        143 Nov 17 19:41 metadata.json
```
This PR modifies the default mode to be `0644` when built via this repository/GitHub actions runner, matching the Java DB behavior.

This doesn't have any impact during normal usage via `trivy` because it changes the ownership/mode to match the user `trivy` is running as (via `hashicorp/go-getter`). However, it is useful to change this behavior because it allows the DB OCI artifact layers can be directly appended to a trivy Docker image in some specialized use-cases without needing to download, decompress, modify, recompress, then upload as new layers (ex: when using AWS Lambda).

### DIFF
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -3,7 +3,6 @@ package db
 import (
 	"bytes"
 	"encoding/json"
-	"io/fs"
 	"os"
 	"path/filepath"
 	"runtime/debug"
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -3,6 +3,7 @@ package db
 import (
 	"bytes"
 	"encoding/json"
+	"io/fs"
 	"os"
 	"path/filepath"
 	"runtime/debug"
@@ -58,6 +59,13 @@ type Option func(*Options)
 
 type Options struct {
 	boltOptions *bolt.Options
+	mode        fs.FileMode
+}
+
+func WithMode(mode fs.FileMode) Option {
+	return func(opts *Options) {
+		opts.mode = mode
+	}
 }
 
 func WithBoltOptions(boltOpts *bolt.Options) Option {
@@ -67,7 +75,9 @@ func WithBoltOptions(boltOpts *bolt.Options) Option {
 }
 
 func Init(dbDir string, opts ...Option) (err error) {
-	dbOptions := &Options{}
+	dbOptions := &Options{
+		mode: 0600,
+	}
 	for _, opt := range opts {
 		opt(dbOptions)
 	}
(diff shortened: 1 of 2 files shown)


pkg/db/db.go
```diff
@@ -90,7 +100,7 @@ func Init(dbDir string, opts ...Option) (err error) {
 		debug.SetPanicOnFault(false)
 	}()
 
-	db, err = bolt.Open(dbPath, 0600, dbOptions.boltOptions)
+	db, err = bolt.Open(dbPath, dbOptions.mode, dbOptions.boltOptions)
```
knqyf263, line 103: We don't have a case where we want to configure the permission. What if we just pass 0644 here?

lyoung-confluent: This `Init` method [is invoked by trivy](https://github.com/aquasecurity/trivy/blob/516e7cb906acb44d0d85f4e7bb123b17aab352a8/pkg/db/db.go#L36) so I wanted to avoid modifying the default behavior unless I had to. However, looking into it this wouldn't even result in the creation of a new `trivy.db` file, it's only used to open an existing (downloaded) one so the mode is irrelevant. If you'd prefer the simpler 1-line change to `bolt.Open` I can update the PR

knqyf263: Yes, as you pointed out, Trivy doesn't create any new files. It's all done in GitHub Actions of trivy-db.
@DmitriyLewen Any concerns about permission updates?

DmitriyLewen: hm... I can't think of any issues for these changes.

lyoung-confluent: I've updated the PR to just directly modify this line to use `0644` by default

knqyf263 (approved):

pkg/db/db.go
DmitriyLewen: didn't have time to write (you already merged this PR):
IIUC we also need to change perms here:
https://github.com/aquasecurity/trivy-db/blob/a2f579001fbd352f5882990872e386fa39a2484d/pkg/db/db.go#L88

lyoung-confluent: https://github.com/aquasecurity/trivy-db/pull/472

PR #480, opened 2024-12-23
ci/cd: Upload artifacts to Aqua registry

tikolsky:
In order for internal and external users not to hit dockerhub rate limits, we need to upload opensource artifacts to aquasec as well upon releases.


knqyf263:
Looks like a mistake

PR #483, opened 2025-01-15
feat(redhat): migrate OVALv2 from CSAF VEX

knqyf263:


knqyf263:
## Update on Blocker Status

The blocker mentioned in the PR description has been resolved.

### Background
[SECDATA-1097](https://issues.redhat.com/browse/SECDATA-1097) ("Add binary package names to CSAF VEX files for unpatched vulnerabilities") has been **closed as Done**.

### What Changed
Previously, there was an asymmetry in package name handling:

| Format | Patched | Unpatched |
|--------|---------|-----------|
| OVALv2 | binary only | source + binary |
| CSAF VEX (old) | source + binary | **source only** ← problem |
| CSAF VEX (new) | source + binary | **source + binary** ✅ |

### Verification
Looking at [CVE-2023-5344](https://security.access.redhat.com/data/csaf/v2/vex/2023/cve-2023-5344.json), unpatched vulnerabilities now include both binary and source packages:

```json
"product_ids": [
  "red_hat_enterprise_linux_9:vim-X11",       // binary
  "red_hat_enterprise_linux_9:vim-common",    // binary
  "red_hat_enterprise_linux_9:vim-enhanced",  // binary
  "red_hat_enterprise_linux_9:vim-minimal",   // binary
  "red_hat_enterprise_linux_9:vim.src"        // source
]
```

Source packages can be identified by:
- `.src` suffix in product_id
- `?arch=src` in PURL

### Impact
Since binary package names are now included for unpatched vulnerabilities, older Trivy clients should be able to detect them correctly. This removes the breaking change concern.

knqyf263:
New problem with kernel packages
https://issues.redhat.com/browse/SECDATA-1174

knqyf263:
New critical problems
- https://issues.redhat.com/browse/SECDATA-1176
- https://issues.redhat.com/browse/SECDATA-1177 (more critical: no unfixed vulns will be detected)

pkg/vulnsrc/redhat-csaf/set.go
```diff
@@ -0,0 +1,66 @@
...
+
+import (
+	"sort"
+
+	"golang.org/x/exp/constraints"
+)
+
+// -----------------------------------------
+// Generic Set implementation (thread-unsafe
+// -----------------------------------------
```
DmitriyLewen: nit:
```suggestion
// ------------------------------------------
// Generic Set implementation (thread-unsafe)
// ------------------------------------------
```

pkg/set/set.go
DmitriyLewen, line 1: I think we can move set implementation into pkg/set dir.

pkg/vulnsrc/redhat-csaf/types.go
```diff
@@ -0,0 +1,122 @@
...
+	return json.Marshal(&struct {
+		Status int `json:"Status,omitempty"`
+		Alias
+	}{
+		Status: int(e.Status),
+		Alias:  Alias(e),
+	})
+}
+
+func (e *Entry) UnmarshalJSON(data []byte) error {
```
DmitriyLewen: nit:
```suggestion
// UnmarshalJSON unmarshals JSON data into Entry and validates + converts Status to types.Status.
func (e *Entry) UnmarshalJSON(data []byte) error {
```

knqyf263:
https://issues.redhat.com/browse/SECDATA-1176 got closed

pkg/vulnsrc/redhat-csaf/advisory.go
```diff
@@ -0,0 +1,216 @@
...
+	if rels := pt.RelationShips; rels != nil {
+		for _, rel := range *rels {
+			if rel == nil {
+				continue
+			}
+			if fpn := rel.FullProductName; fpn != nil && lo.FromPtr(fpn.ProductID) == id {
+				return fpn.ProductIdentificationHelper, true
+			}
+		}
+	}
```
DmitriyLewen: Is this possible?
Can you share example?

pkg/vulnsrc/redhat-csaf/advisory.go
```diff
@@ -0,0 +1,216 @@
...
+		return "", "", eb.Wrapf(errUnexpectedRecord, "module purl not found")
+	}
+	eb = eb.With("module_purl", *moduleRef.PURL)
+
+	purl, err := packageurl.FromString(string(*moduleRef.PURL))
+	if err != nil {
+		return "", "", eb.Wrapf(err, "invalid purl")
+	} else if purl.Type != "rpmmod" { // Must be "rpmmod" for modular streams
+		return "", "", eb.With("purl_type", purl.Type).Wrapf(errUnexpectedRecord, "unexpected purl type")
+	}
```
DmitriyLewen: IIUC we shouldn't return error for `rpm` type:
https://github.com/knqyf263/trivy-db/blob/e581cfc764b0a1b874cd30616a991e55aac2b344/pkg/vulnsrc/redhat-csaf/advisory.go#L50-L62

also we can move this code here.

pkg/vulnsrc/redhat-csaf/advisory.go
```diff
@@ -0,0 +1,216 @@
...
+	if module == "" {
+		if rpmmod := purl.Qualifiers.Map()["rpmmod"]; rpmmod != "" {
+			// rpmmod = "389-ds:1.4:8100020240613122040:25e700aa"
+			// module = "389-ds:1.4"
+			parts := strings.SplitN(rpmmod, ":", 3)
+			if len(parts) >= 2 {
+				module = parts[0] + ":" + parts[1]
+			}
+		}
+	}
```
DmitriyLewen, line 140: Can you add testcase for this?

pkg/vulnsrc/redhat-csaf/parse.go
```diff
@@ -0,0 +1,490 @@
...
+		// - Product ID: 6Workstation-RHSCL-2.0:rh-mariadb100-mariadb-1:10.0.20-1.el6.src
+		//    -> PURL: pkg:rpm/redhat/rh-mariadb100-mariadb@10.0.20-1.el6?arch=src\u0026epoch=1
+		//       CPE: cpe:/a:redhat:rhel_software_collections:2::el6
+		//
+		// In this case, we only need to keep one record since they represent the same affected component.
+		type uniqKey struct {
+			Package
+			RawEntry
+		}
+		uniq := NewSet[uniqKey]()
```
DmitriyLewen: Perhaps I'm missing something:
Do we actually need such detailed deduplication here?
Currently:
If one of the fields in another entry differs, we will still overwrite the entire RawEntry in this map:
https://github.com/knqyf263/trivy-db/blob/e581cfc764b0a1b874cd30616a991e55aac2b344/pkg/vulnsrc/redhat-csaf/parse.go#L391-L397
And if not, there won't be an overwrite.
But if we remove RawEntry, we'll simply always overwrite the map.
It seems like this is an excessive check.

pkg/vulnsrc/redhat-csaf/advisory.go
DmitriyLewen, line 1: I'm not sure, but this might be a "simplification" of the code/logic:
Currently, we iterate through the relationships/product_tree (with nested branches) arrays every time.
What if we collect them into maps:
map[full_product_name.product_id]relationship
map[product_id]branch
and so on
Then we can easily find the needed element by product_id.
This might make the logic a bit simpler + during debugging you'll immediately be able to see all elements in the debugger.

pkg/vulnsrc/redhat-csaf/parse.go
```diff
@@ -0,0 +1,490 @@
...
+			}
+			uniq.Append(key)
+
+			p.addRawEntry(pkg, vulnID, rawEntry)
+		}
+	}
+	return nil
+}
+
+func (p *Parser) detectStatus(remediation *csaf.Remediation) types.Status {
```
DmitriyLewen, line 343: what about `workaround` status?

It seems that `affected` is more logical status - https://github.com/aquasecurity/vuln-list-redhat/blob/b4fbd07781fe3397b4647e2c5e01327cd7c39989/csaf-vex/2025/cve-2025-4574.json#L2817-L2818
wdyt?

pkg/vulnsrc/redhat-csaf/aggregate.go
```diff
@@ -0,0 +1,231 @@
...
+			AffectedCPEList: []string{key.CPE},
+			Arches:          archSet.Values(), // Get sorted slice of architectures
+		})
+	}
+
+	return result
+}
+
+// aggregateCPEs merges entries that only differ in their CPEs
+func (a *Aggregator) aggregateCPEs(entries Entries) Entries {
```
DmitriyLewen, line 143: Interestingly, CSAF contains significantly fewer CPEs.
Example for `389-ds-base`:
OVAL:
изображение

CSAF:
изображение


However, I checked the VEX file, and the presence of 2 CPEs in this case is absolutely correct.

knqyf263: There are countless differences like this between OVALv2 and CSAF. The problem is that we have no way of knowing which one is correct, so we have to ask for clarification each time. 

According to the page below, there are obviously more than two products, so I’m wondering whether OVALv2 is the correct one.
https://access.redhat.com/errata/RHSA-2025:7395

knqyf263: Reported
https://issues.redhat.com/projects/SECDATA/issues/SECDATA-1181

pkg/vulnsrc/redhat-csaf/parse.go
knqyf263: Good question. I investigated this further and found an interesting issue with the CSAF VEX data.

Looking at [CVE-2025-4574](https://github.com/aquasecurity/vuln-list-redhat/blob/b4fbd07781fe3397b4647e2c5e01327cd7c39989/csaf-vex/2025/cve-2025-4574.json), the `workaround` remediation includes **both** affected and not-affected products:

**Products in workaround: 141**
- `known_affected`: 79 products
- `known_not_affected`: 62 products

The `known_not_affected` products are explicitly marked with [`vulnerable_code_not_present`](https://github.com/aquasecurity/vuln-list-redhat/blob/b4fbd07781fe3397b4647e2c5e01327cd7c39989/csaf-vex/2025/cve-2025-4574.json#L2549-L2552) flag:

```json
"flags": [
  {
    "label": "vulnerable_code_not_present",
    "product_ids": [
      "red_hat_directory_server_11:redhat-ds:11/389-ds-base",
      "red_hat_directory_server_12:redhat-ds:12/389-ds-base",
      ...
    ]
  }
]
```

Yet these same products appear in the [workaround remediation](https://github.com/aquasecurity/vuln-list-redhat/blob/b4fbd07781fe3397b4647e2c5e01327cd7c39989/csaf-vex/2025/cve-2025-4574.json#L2817-L2818).

This seems logically inconsistent - why would a product with "vulnerable code not present" need a workaround?

I'm not sure if this is intentional on Red Hat's side or a data quality issue. We may need to filter out `known_not_affected` products when processing workaround remediations, or report this to Red Hat.

knqyf263: Since I feel like it's their problem, I reported it.
https://issues.redhat.com/browse/SECDATA-1182

pkg/set/set.go
knqyf263: Moved to `pkg/set`. Also renamed:
- `OrderedSet` → `set.Ordered`
- `NewOrderedSet` → `set.NewOrdered`
- `NewSet` → `set.New`

630281d

pkg/vulnsrc/redhat-csaf/types.go
knqyf263: Added a comment explaining why `MarshalJSON` serializes `Status` as an int instead of a string.

`UnmarshalJSON` is not currently needed because `Get()` delegates to `redhat-oval`. It's commented out with a TODO to uncomment when `redhat-oval` is removed.

07ceeb0

pkg/vulnsrc/redhat-csaf/advisory.go
knqyf263: Based on my investigation of Red Hat's CSAF VEX data in vuln-list-redhat, I could not find any example where `product_identification_helper` is defined in relationships - it's always in branches.

For example, even for modular packages like `firefox-x11::firefox:flatpak`, the `product_reference` is defined in branches with the PURL:
```json
{
  "product_id": "firefox-x11::firefox:flatpak",
  "product_identification_helper": {
    "purl": "pkg:rpm/redhat/firefox-x11?rpmmod=firefox:flatpak"
  }
}
```

While the CSAF spec allows `full_product_name` in relationships to have `product_identification_helper`, Red Hat does not use this pattern. This code path appears to be defensive for:
1. Potential future changes in Red Hat's data format
2. Compatibility with other CSAF producers

Should I remove this code path or keep it as defensive code?

knqyf263: I initially thought `pkg:rpmmod` was defined in relationships, but after investigation, I found that it is actually defined in branches. Red Hat's CSAF VEX data always defines `product_identification_helper` (containing PURL/CPE) in branches, not in relationships. This code path was never triggered for Red Hat data, so I have removed it.

I also added a test case for the new `?rpmmod=` qualifier format (e.g., `pkg:rpm/redhat/firefox?rpmmod=firefox:flatpak`) which replaced the old `pkg:rpmmod/...` format.

knqyf263: Fixed in 9445f4e26ad829da98deddf4041e4d0b14ea9c1a

pkg/vulnsrc/redhat-csaf/advisory.go
knqyf263: Added in https://github.com/aquasecurity/trivy-db/commit/9445f4e26ad829da98deddf4041e4d0b14ea9c1a

pkg/vulnsrc/redhat-csaf/parse.go
knqyf263: I may be misunderstanding your point, but when you say "overwrite", are you referring to [this line](https://github.com/knqyf263/trivy-db/blob/e581cfc764b0a1b874cd30616a991e55aac2b344/pkg/vulnsrc/redhat-csaf/parse.go#L396)? If so, this is an `append`, not an overwrite.

There are many entries associated with a pair of the same package and vulnerability ID. For example, cases where only the CPE or arch differs. Therefore, this is not overwriting a map but appending to a slice to be aggregated later. This means that if there are duplicate entries, the slice will contain duplicate elements. That's why this deduplication is necessary.

DmitriyLewen: I was reviewing the logic for iterating through rawEntries.
I don't know why I thought about overwriting...

It seems I messed up a bit.
Thanks for the clarification and sorry for confusing.

pkg/vulnsrc/redhat-csaf/advisory.go
knqyf263: I'm not sure how much this would simplify the code. This lookup processing is needed anyway - the difference is just whether we do it incrementally while iterating through remediations, or all at once upfront. Managing the maps would add some lines.

Considering that VEX files contain many product IDs that are NOT affected by the vulnerability, parsing everything upfront would be wasteful in terms of both CPU and memory.

On the other hand, if the point is about modularizing/separating the processing to make debugging easier for each step, that might make sense.

pkg/vulnsrc/redhat-csaf/parse.go
knqyf263: I investigated this further. The workaround remediation includes known_not_affected products, which seems like a data inconsistency (as I reported it).
                                                                                                                                           
However, from a vulnerability detection perspective, this is not a problem. I verified that all known_affected products are always covered by other remediations (vendor_fix, no_fix_planned, or none_available). So, ignoring the workaround does not cause any affected products to be missed.

pkg/vulnsrc/redhat-csaf/advisory.go
DmitriyLewen: Yes, I assumed this wouldn't be optimal in terms of resources.
But VEX files still contain many errors and inaccuracies, so it seemed to me that this way it would be easier to debug the code and find them. However, you've studied these files in more detail (and wrote this code), so you know better.
But I'm not against the current implementation, so we can leave everything as is.

pkg/vulnsrc/redhat-csaf/parse.go
DmitriyLewen: Cool! Thanks for the investigation.

pkg/vulnsrc/redhat-csaf/advisory.go
knqyf263: I've implemented it like this - what do you think?
https://github.com/aquasecurity/trivy-db/pull/483/commits/777be589ff518e3e1d740237c836a463f11b3a01

Most bugs are in the data, not our script, so when I find issues I always look at the original VEX files first. I haven't found the current implementation hard to debug, but I don't have a strong preference either way, so I'm fine with the map-based approach.

pkg/vulnsrc/redhat-csaf/advisory.go
```diff
@@ -0,0 +1,163 @@
...
+	if helper := productHelperMap[productID]; helper != nil && helper.CPE != nil {
+		return *helper.CPE
+	}
+	return ""
+}
+
+// LookUpProduct returns the pre-computed Product for a given ProductID.
+func (a CSAFAdvisory) LookUpProduct(productID csaf.ProductID) (*Product, error) {
+	return a.productMap[productID], nil
+}
```
DmitriyLewen: nit:
Looks like the error is unneeded now.

DmitriyLewen (approved):
LGTM

pkg/vulnsrc/redhat-csaf/advisory.go
knqyf263: Thanks. https://github.com/aquasecurity/trivy-db/pull/483/commits/72456b88feca2146e94009d964d15529f8f7bd8c

PR #484, opened 2025-01-19
feat: handling of same fix version with different elsa ids and include elsa ids for aqua storage,remove them before saving into oss db

santhosh1729:
- Modify Oracle tracker to include logic for saving advisory IDs (ELSA IDs) in the database.


pkg/vulnsrc/oracle-oval/oracle-oval.go
```diff
@@ -216,32 +217,40 @@ type archFlavor struct {
...
 	latestVersions := selectLatestVersions(advisories)
 
-	// Step 2: Aggregate arches by their chosen version
-	versionToArches := make(map[string][]string)
+	// Step 2: Aggregate architectures by their chosen version
+	versionToArches := make(map[string]types.Advisory)
 	for k, v := range latestVersions {
-		versionToArches[v] = append(versionToArches[v], k.Arch)
+		adv := versionToArches[v.FixedVersion]
+		adv.VendorIDs = v.VendorIDs
```
DmitriyLewen, line 238: It can be wrong for case when 2 advisories with same fixedVersion got from 2 different ELSA-IDs
e.g.
- ELSA-xxxx-0001 - fixedVersion == 0.0.1, arch == amd64
- ELSA-xxxx-0002 - fixedVersion == 0.0.1, arch == aarch64
for this case entry will include only one ELSA-xxxx-0002 (or 0001)

DmitriyLewen: perhaps we need to create 2 different entries for this case

DmitriyLewen (commented):
Hello @santhosh1729 
Thanks for your work!

We are worried about size of trivy-db.
Can you check and write actual and new sizes for DBs?

Also i left 1 comment, take a look, please.

pkg/vulnsrc/oracle-oval/oracle-oval.go
santhosh1729: @DmitriyLewen  Thanks for the review.

As trivy doesn't need this vendor ids data in db, removed them before storing in db so there is no impact on db size.

Addressed same fix version with different elsa-ids and architectures case.

Can you please review this? Thank you!!

DmitriyLewen: Hello @santhosh1729 
I am currently bezy on new Trivy release.

I will check your PR after release.

DmitriyLewen:
@santhosh1729 

I see a little differently how we can work with vendorIDs.
Can you take a look [these changes](https://github.com/dmitriylewen/trivy-db/compare/d74594e72b61b7a76401def7a5037790e2311a03..e0e80d62def9373ae07ddd1e765bd29662e5717e)? (these changes still require refactoring (update comments, etc.). but you will see the main logic)

santhosh1729:
> @santhosh1729
> 
> I see a little differently how we can work with vendorIDs. Can you take a look [these changes](https://github.com/dmitriylewen/trivy-db/compare/d74594e72b61b7a76401def7a5037790e2311a03..e0e80d62def9373ae07ddd1e765bd29662e5717e)? (these changes still require refactoring (update comments, etc.). but you will see the main logic)

DmitriyLewen:
Hi @santhosh1729
Did you close the PR by mistake?

DmitriyLewen:
But these advisories are for different versions of Oracle Linux.
ELSA-2024-2758 for OL 9
ELSA-2024-3138 for OL 8.

So these advisories will be saved in different containers.
[OSVer from Package](https://github.com/DmitriyLewen/trivy-db/blob/8c398f13db0ed9be333fe1b9ddab158ab7262967/pkg/vulnsrc/oracle-oval/types.go#L40) should handle your case.

Or am I missing something?

santhosh1729:
Sorry, I overlooked the os version.

DmitriyLewen:
No problem. Can you update your PR based on these changes?

santhosh1729:
Sure, I will take care of that

The latest Trivy-DB includes log changes, and Aqua consumes these changes. we encounter an issue because the logger is a private variable.

Can I update the logger to be public everywhere?

https://github.com/aquasecurity/trivy-db/commit/388c617890ac1c513c3fdfec5c8d5fc78d41e58e#diff-4e5beea79657fce401854eb1e0e4558528a32279f61876579816724a01b38d2cR51

DmitriyLewen:
hm... 
Do you use something like this?
```Go
type WrappedVulnSrc struct {
	newField string
	oracleoval.VulnSrc
}

func (w WrappedVulnSrc) testFunc() {
	// you need to get logger here
}
```

Or do you have another case?

Anyway i think it is better to create another PR for these changes (we can review new PR the first)

santhosh1729:
We use OSS tracker by creating the object for struct
```
trivyLog "github.com/aquasecurity/trivy-db/pkg/log"

func NewVulnSrc() *alma.VulnSrc {
	dbc := aquaDB.AquaConfig{}
	return &alma.VulnSrc{
		DB: &AquaAlma{
			Operation: dbc,
		},
		Logger: trivyLog.WithPrefix("alma"),
	}
}
```

DmitriyLewen:
Got it!
Maybe we can use `Functional Options Pattern` (`WithLogger`, etc. functions).
But we are using `DB` field as exported field, so this looks like redundancy for trivy-db, and we can just make `logger` exported field.
@knqyf263 wdyt?

knqyf263:
I'm on vacation now. I'll take a look tomorrow.

DmitriyLewen (approved):
@santhosh1729 thanks for your work!
LGTM

@knqyf263 tell me if you will watch this PR. If not - I will merge it.

knqyf263:
I'm reviewing this PR now.

knqyf263:
@santhosh1729 Doesn't this work?

```
func NewVulnSrc() *alma.VulnSrc {
	vc := alma.NewVulnSrc()
	vc.DB = &AquaAlma{
		Operation: aquaDB.AquaConfig{},
	}
}

**Comment by santhosh1729** [comment]
```
type VulnSrc struct {
	DB // Those who want to customize Trivy DB can override put/get methods.
	DB     // Those who want to customize Trivy DB can override put/get methods.
	logger *log.Logger
}
```


https://github.com/aquasecurity/trivy-db/pull/486/files#diff-4e5beea79657fce401854eb1e0e4558528a32279f61876579816724a01b38d2cR51

The logger needs to be initialized; otherwise, an error will occur at `vs.logger.Warn("Invalid path", log.FilePath(path))`
https://github.com/aquasecurity/trivy-db/pull/486/files#diff-4e5beea79657fce401854eb1e0e4558528a32279f61876579816724a01b38d2cR94

**Comment by knqyf263** [comment]
It's initialized here through NewVulnSrc.
https://github.com/knqyf263/trivy-db/blob/518f5629d9bef210dac3c3ac1d7ed08acaf7783b/pkg/vulnsrc/alma/alma.go#L62

**Comment by santhosh1729** [comment]
Aqua uses this way by overriding the Put function on Aqua's end
```

type AquaAlma struct {
	aquaDB.Operation
}

func NewVulnSrc() *alma.VulnSrc {
	dbc := aquaDB.AquaConfig{}
	return &alma.VulnSrc{
		DB: &AquaAlma{
			Operation: dbc,
		},
		Logger: trivyLog.WithPrefix("alma"),
	}
}
```

**Comment by knqyf263** [comment]
Hmm. Why doesn't https://github.com/aquasecurity/trivy-db/pull/484#issuecomment-2652735212 work?

**Comment by santhosh1729** [comment]
Sorry for the late reply. Yes, it works for alma case.
There are other trackers in which dbc field is not exported so unable override with aqua db config

image


```
amazon - https://github.com/aquasecurity/trivy-db/pull/486/files#diff-d86629b57d4a89d5021578bb53ee01854b72bd648dc6e00c17bb05cb3699f868R37
debian - https://github.com/aquasecurity/trivy-db/pull/486/files#diff-51ceff92c4f1b9d7cce4380f0a43352791b7850ffd07caeb7081e00cc20483b4R67
redhat - https://github.com/aquasecurity/trivy-db/pull/486/files#diff-1768c801e6c1230a63e41a360b0a667eb9db1b5c3352eb8e23ac826be47d3ac2R45
suse   - https://github.com/aquasecurity/trivy-db/pull/486/files#diff-ecaeb113322d8591d4e43a37dfbe2515da3d8025d62756bb2a1ce839809d08dcR48
ubuntu - https://github.com/aquasecurity/trivy-db/pull/486/files#diff-ee12b114717f396f628bd99eb7f5f77e6b9769ee148b1d7be9326eb888aa0859R77
```

**Comment by knqyf263** [comment]
>There are other trackers in which dbc field is not exported so unable override with aqua db config

Yes, can you please elaborate on what the problem is? I understand you override the behavior of Alma, Oracle, etc, by passing DB, while you call the OSS NewVulnSrc for Amazon, Debian, etc., as we have changed the implementation to be overrideable as you requested. If you call NewVulnSrc, the logger will be correctly initialized.

**Comment by santhosh1729** [comment]
Sure, I'm trying to explain that in the new approach, we are using amazon.NewVulnSrc() to initialize the vulnerability source. However, we cannot override dbc with aquadb.AquaConfig{} because dbc is an unexported variable in the amazon package

Trivy OSS amazon vulnsrc struct
```

func NewVulnSrc() VulnSrc {
    return VulnSrc{
        dbc:        db.Config{},
        logger:     log.WithPrefix("amazon"),
        advisories: map[string][]ALAS{},
    }
}
```

Existing aqua implementation to use OSS amazon tracker

```
func NewVulnSrc() VulnSrc {
    // Existing code
    return VulnSrc{
        dbc:        aquadb.AquaConfig{},
        advisories: map[string][]ALAS{},
    }
}
```
Suggested implementation

func NewVulnSrc() VulnSrc {
    vc := amazon.NewVulnSrc() // Trivy DB Amazon source
    vc.dbc = aquadb.AquaConfig{} // ❌ Cannot override because `dbc` is unexported
    ``
return vc
  }
  
  The dbc field in amazon.NewVulnSrc() is not exported, meaning we cannot override it outside the package.

**Comment by DmitriyLewen** [comment]
It seems like the best way would be to make `dbc` an exported field, as is implemented for other sources.

**Comment by knqyf263** [comment]
>The dbc field in amazon.NewVulnSrc() is not exported, meaning we cannot override it outside the package.

No, I know that. I meant you didn't need to override it, and that's why it is not exported. And this PR is about Oracle Linux. Why are you talking about Amazon? Do you also need to override the behavior in Amazon?

**Comment by knqyf263** [comment]
I'm still wondering why your current code doesn't work. You can continue to use the following code. Do you need any change?

```
func NewVulnSrc() VulnSrc {
    // Existing code
    return VulnSrc{
        dbc:        aquadb.AquaConfig{},
        advisories: map[string][]ALAS{},
    }
}
```

**Comment by DmitriyLewen** [comment]
these are 2 different PRs 😄 
this PR is for Oracle Linux

and this discussion (about logs) is for another PR

**Comment by knqyf263** [comment]
In the case of Amazon Linux, they seem to use their own VulnSrc, unlike Oracle Linux,  so they don't need to initialize a logger, right? For Oracle Linux, they use the OSS VulnSrc and need to initialize the logger somehow, and I suggested how to do that. However, I'm still unsure why they want to initialize the logger in Amazon Linux.
Or are we talking about extending Amazon Linux as an entirely new requirement?

**Comment by DmitriyLewen** [comment]
> Or are we talking about extending Amazon Linux as an entirely new requirement?

IIUC your are right.

This PR is required that `PutInputs` includes VendorID
and as for the logger and VulnSrc - this is a conversation about changes for another PR

@santhosh1729 correct me - if you want to add these changes into this PR.

**Comment by santhosh1729** [comment]
For the Amazon tracker case, Aqua first calls OSS Amazon and then Aqua Amazon Tracker to append advisory details. This is why logger initialization is necessary.

I will initiate a separate thread for the logger changes. If the Oracle changes look good, please proceed with the merge.

**Comment by knqyf263** [comment]
>Aqua first calls OSS Amazon and then Aqua Amazon Tracker to append advisory details

Why can't you call `amazon.NewVulnSrc` in OSS?

**Comment by knqyf263** [comment]
>For the Amazon tracker case, Aqua first calls OSS Amazon and then Aqua Amazon Tracker to append advisory details.

But you said Aqua first calls OSS Amazon tracker. Then, you can use NewVulnSrc.

**Comment by knqyf263** [comment]
@santhosh1729 I think you are mixing up the current implementation with the implementation you want to achieve. Currently, `dbc` is not exported, so Aqua can't overwrite it today. Therefore, the `logger` should not be an issue.
https://github.com/aquasecurity/trivy-db/blob/main/pkg/vulnsrc/amazon/amazon.go#L67-L72

You showed this code:

```
func NewVulnSrc() VulnSrc {
    // Existing code
    return VulnSrc{
        dbc:        aquadb.AquaConfig{},
        advisories: map[string][]ALAS{},
    }
}
```

This `VulnSrc` is `aquaamazon.VulnSrc`, right? How are you using `ossamazon.VulnSrc`?

If you now have a *new* need to overwrite `dbc`, then the logger might become a problem, but that is an entirely different matter. Could you clarify once again what exactly you want to do?

File: pkg/vulnsrc/oracle-oval/oracle-oval.go [hunk 2]
```diff
@@ -335,6 +356,30 @@ func (o *Oracle) Put(tx *bolt.Tx, input PutInput) error {
 	return nil
 }
 
+// removeVendorIDs removes VendorID from advisories + merges arches for advisories (by fixedVersion).
+// This is needed to save space in OSS trivy-db.
+// But Aqua storage requires this information.
+func removeVendorIDs(advs types.Advisories) types.Advisories {
+	versionToArches := make(map[string][]string) // fixed version -> arches
+	for _, entry := range advs.Entries {
+		versionToArches[entry.FixedVersion] = append(versionToArches[entry.FixedVersion], entry.Arches...)
```
knqyf263, line 365: Isn't it possible that arches are duplicated?

pkg/vulnsrc/oracle-oval/oracle-oval.go
```diff
@@ -335,6 +356,30 @@ func (o *Oracle) Put(tx *bolt.Tx, input PutInput) error {
...
+	}
+
+	entries := lo.MapToSlice(versionToArches, func(ver string, arches []string) types.Advisory {
+		sort.Strings(arches)
```
knqyf263, line 369: In other words, don't we need deduplication?

```suggestion
		sort.Strings(lo.Uniq(arches))
```

pkg/vulnsrc/oracle-oval/oracle-oval.go
DmitriyLewen: we can get duplicates only if we have 2 records with the same `fixed version` and `arch`.
But it is impossible because we store uniq arch+flavor:
https://github.com/aquasecurity/trivy-db/blob/47d4c6ca71d4261e331332b1209a09d780316765/pkg/vulnsrc/oracle-oval/oracle-oval.go#L268-L292

But if you found this case - give me an example and I will check.

knqyf263: I'm probably missing something, but the logic you shared doesn't appear to help for me.

image

DmitriyLewen: selectLatestVersions returns only 2 values:
- x86_64 + normal flavors - v1.0.0 + ELSA-0001
- arm64 + normal flavors - v1.0.0 + ELSA-0001

https://github.com/aquasecurity/trivy-db/blob/47d4c6ca71d4261e331332b1209a09d780316765/pkg/vulnsrc/oracle-oval/oracle-oval.go#L277-L289


This PR does not preserve all ELSA-ids. IIUC @santhosh1729 - he only needs one ELSA-ID (for the latest fixed version).

knqyf263: I see. It's too complicated that different ELSA-IDs are used per arch. I'm sure it will cause bugs.

image

knqyf263: Could we simplify implementation by treating the ELSA-ID as the primary ID, thereby eliminating the need for aggregation? It might be necessary to upgrade to Trivy DB v3, as the current Trivy client does not support it.

DmitriyLewen: > It's too complicated that different ELSA-IDs are used per arch. I'm sure it will cause bugs.

it all depends on how aqua plans to use it. With the right approach - I don't think there will be any problems with it. And this confusion is due to confusion in the oracle advisories.

> Could we simplify implementation by treating the ELSA-ID as the primary ID

this will simplify the work for us.
But on the other hand, users will have to handle the nuances of Oracle advisories themselves (e.g., when several ELSA-IDs fix one CVE-ID).

DmitriyLewen: >  It might be necessary to upgrade to Trivy DB v3

I don't mind. We already talked about switching to v3.

knqyf263: >But on the other hand, users will have to handle the nuances of Oracle advisories themselves (e.g., when several ELSA-IDs fix one CVE-ID).

I think Trivy can convert the detected ELSA-ID into a CVE-ID and retrieve only the latest version if there are duplicate CVE-IDs. This way, we wouldn't have to worry about arch or flavor. Also, we can keep showing users the same information (CVE-IDs). What do you think?

DmitriyLewen: We do something like that for RedHat:
https://github.com/aquasecurity/trivy/blob/4202c4ba0d8fcff4b89499fe03050ef4efd37330/pkg/detector/ospkg/redhat/redhat.go#L117-L118

So we can use similar logic.

Main point is that we need to change the Trivy code.
And I also don't know how to change the size of DB.

knqyf263: >Main point is that we need to change the Trivy code.

Yes, so we cannot apply the change now.

>And I also don't know how to change the size of DB.

Also, true. We need to check it.

Anyway, I'll merge this PR and think about v3 later.

knqyf263 (approved):

PR #485, opened 2025-01-21
refactor: add `AllSourceIDs`

DmitriyLewen:
## Description
Add `AllSourceIDs` with list of all sourceIDs.

This is required for aquasecurity/trivy/issues/8180

### DIFF
--- a/pkg/vulnsrc/vulnerability/const.go
+++ b/pkg/vulnsrc/vulnerability/const.go
@@ -73,7 +73,9 @@ var Ecosystems = []types.Ecosystem{
 	Kubernetes,
 }
 
-// AllSourceIDs is list of all supported sourceID
+// AllSourceIDs is a list of all supported sourceIDs.
+// This order affects the choice of `Severity`, `Title`, `Description` and `CweIDs`.
+// e.g. if `NVD` has no severity - we will keep the severity of `RedHat` (if it exists), etc.
 var AllSourceIDs = []types.SourceID{
 	NVD,
 	RedHat,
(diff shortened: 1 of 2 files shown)


pkg/vulnsrc/vulnerability/const.go
```diff
@@ -70,3 +72,34 @@ var Ecosystems = []types.Ecosystem{
 	Bitnami,
 	Kubernetes,
 }
+
+// AllSourceIDs is list of all supported sourceID
```
knqyf263, line 76: We may want to document that the order affects the severity selection.

pkg/vulnsrc/vulnerability/const.go
```diff
@@ -70,3 +72,36 @@ var Ecosystems = []types.Ecosystem{
 	Bitnami,
 	Kubernetes,
 }
+
+// AllSourceIDs is a list of all supported sourceIDs.
+// This order affects the choice of `Severity`, `Title`, `Description` and `CweIDs`.
+// e.g. if `NVD` has no severity - we will keep the severity of `RedHat` (if it exists), etc.
```
knqyf263, line 78: nit

```suggestion
// AllSourceIDs lists all supported vulnerability source IDs in order of precedence.
// When searching for vulnerability details (Severity, Title, Description, and CWE-IDs),
// the sources are checked in this order until valid data is found.
// For example, if severity data is missing in NVD, it will check Red Hat next,
// continuing through the list until it finds a valid severity value.```

PR #486, opened 2025-01-21
refactor: use slog

knqyf263:
## Description
This PR replaces `log` with `log/slog` and standardizes the logger implementation across vulnerability source packages by introducing a consistent logging pattern with source-specific prefixes.

## Key changes
- Replace direct usage of `log` with `github.com/aquasecurity/trivy-db/pkg/log`
- Embed `*log.Logger` in `VulnSrc` structs
- Implement source-specific prefixes for each vulnerability source using `log.WithPrefix()`
- Standardize logging method calls (e.g., `vs.Info()`, `vs.Warn()`) across all sources

## Benefits
1. Improved log clarity:
   - Each log message is clearly tagged with its source
   - Easier to identify and debug source-specific issues
   - Better log filtering capabilities

2. Enhanced maintainability:
   - Consistent logging patterns across all vulnerability sources
   - Centralized logging configuration through the `pkg/log` package
   - Easier to modify logging behavior globally

3. Future-proof design:
   - Abstraction of logging implementation details
   - Easier to refactor or replace the underlying logging mechanism (`slog`)
   - Reduced dependency on external logging libraries in individual packages

### DIFF
--- a/pkg/utils/file.go
+++ b/pkg/utils/file.go
@@ -26,7 +26,7 @@ func FileWalk(root string, walkFn func(r io.Reader, path string) error) error {
 		}
 
 		if info.Size() == 0 {
-			log.Info("Invalid file size", log.String("path", path), log.Int64("size", info.Size()))
+			log.Info("Invalid file size", log.FilePath(path), log.Int64("size", info.Size()))
 			return nil
 		}
 
--- a/pkg/utils/file.go
+++ b/pkg/utils/file.go
@@ -4,11 +4,12 @@ import (
 	"encoding/json"
 	"io"
 	"io/fs"
-	"log"
 	"os"
 	"path/filepath"
 
 	"golang.org/x/xerrors"
+
+	"github.com/aquasecurity/trivy-db/pkg/log"
 )
 
 func FileWalk(root string, walkFn func(r io.Reader, path string) error) error {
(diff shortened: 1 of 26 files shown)


pkg/utils/file.go
```diff
@@ -25,7 +26,7 @@ func FileWalk(root string, walkFn func(r io.Reader, path string) error) error {
 		}
 
 		if info.Size() == 0 {
-			log.Printf("invalid size: %s\n", path)
+			log.Info("Invalid file size", log.String("path", path), log.Int64("size", info.Size()))
```
DmitriyLewen, line 29: I suggest to use `path` as constant (as in Trivy).
Also we can use `path` for `dir` (e.g. azure package)

DmitriyLewen (approved):
LGTM
left 1 small comment

pkg/utils/file.go
knqyf263: Do you mean adding a function?
https://github.com/aquasecurity/trivy/blob/f9b6d016854086f9c1edb96a1f70c248ae3c9324/pkg/log/handler.go#L281-L284

DmitriyLewen: yes, make a function.
Sorry for the confusion

DmitriyLewen: But perhaps these are extra changes 😄

knqyf263: They are not user-facing messages. IMHO, message consistency is no big deal, but there's no downside of defining functions. Added in https://github.com/aquasecurity/trivy-db/pull/486/commits/518f5629d9bef210dac3c3ac1d7ed08acaf7783b

DmitriyLewen (approved):

PR #488, opened 2025-01-24
refactor: migrate from xerrors to samber/oops for enhanced error handling

knqyf263:
This PR proposes migrating our error handling from xerrors to [github.com/samber/oops](https://github.com/samber/oops).

## Background
- We've been using xerrors primarily because the standard `errors` package doesn't support stack traces
- However, adding context to errors has been cumbersome with xerrors

## Benefits of oops
Please refer to [the oops README](https://github.com/samber/oops) for comprehensive features, but here are the key benefits for our use case:

1. Provides a slog-like interface
   - Easy to add contextual attributes using `oops.With()`, similar to `slog.With()`
   - As we recently migrated to slog (#486), this aligned interface will improve the development experience

2. Enhanced error output
   - Better visualization of error context and domains
   - Makes debugging easier by providing more structured and readable error information

image

This change will make our error handling more maintainable and debugging more effective while maintaining all the benefits we had with xerrors (like stack traces).

## Real-world validation:
We've successfully tested oops in [another project](https://github.com/aquasecurity/vexhub-crawler) where its improved error output [helped us](https://github.com/aquasecurity/vexhub-crawler/actions/runs/12891691388/job/35944215335) identify issues more efficiently.

## Implementation Guidelines

1. Domain Definition (oops.In)
   - Define `oops.In` at the top of each exported (public) function
   - Rationale: Private functions don't need domain definition as they are called within the context of public functions

2. OopsErrorBuilder Usage
   - Use `OopsErrorBuilder` when the same contextual attributes are reused multiple times
   - Direct error creation is preferred for one-off errors
   - Exception: OopsErrorBuilder can be used when there are many attributes, even for one-off errors

3. OopsErrorBuilder Scope Management
   When creating a new `OopsErrorBuilder` from an existing one:
   - Be careful with variable scope
   - Avoid overwriting existing builder variables as attributes will persist after leaving the scope
   - Correct usage: `eb := eb.With()` to create a new variable
   - Incorrect usage: `eb = eb.With()` (overwrites existing variable)

Example:

```go
// Good
func PublicFunction() error {
    // Define domain at the top
    eb := oops.In("domain")
    
    // Reusable builder for common attributes
    eb = eb.With("common_attr", "value").With("other_attr", "value")

    for _, value := range values {
        // Create a new error builder fothe r specific scope
        eb := eb.With("specific", value)
        return eb.Wrapf(err, "error")
    }
    
    // Use directly for one-off errors
    if err != nil {
        return oops.Wrap(err, "failed to do something").With("simple_attr", "value")
    }
}
```

4. Context Attribution Responsibility
   Internal Package Calls:
   - Caller: Generally avoid adding context attributes unless necessary
   - Callee: Responsible for adding appropriate context attributes
   - Rationale: Prevents duplicate attribute attribution at both caller and callee levels

   External Package Calls:
   - Caller: Must add project-specific context attributes
   - Rationale: External packages won't add project-specific context, so it's our responsibility

```go
// Good - Internal package calls
func CallerFunction() error {
    in := oops.In("caller")
    
    // Let the callee handle the context
    if err := internalPkg.DoSomething(); err != nil {
        return oops.Wrapf(err, "internal calls")
    }
    
    // Add context only when needed
    if err := internalPkg.DoSomethingSpecial(); err != nil {
        return oops.Wrap(err, "special case").
            With("special_context", "value")
    }
}

// Good - External package calls
func ExternalCall() error {
    in := oops.In("external")
    
    // Add our context for external calls
    if err := externalPkg.DoSomething(); err != nil {
        return oops.Wrap(err, "external operation failed").
            With("operation", "something").
            With("external_context", "value")
    }
}
```

### DIFF
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -8,8 +8,8 @@ import (
 	"runtime/debug"
 	"strings"
 
+	"github.com/samber/oops"
 	bolt "go.etcd.io/bbolt"
-	"golang.org/x/xerrors"
 
 	"github.com/aquasecurity/trivy-db/pkg/log"
 	"github.com/aquasecurity/trivy-db/pkg/types"
@@ -72,10 +72,12 @@ func Init(dbDir string, opts ...Option) (err error) {
 		opt(dbOptions)
 	}
 
+	eb := oops.With("db_dir", dbDir)
 	if err = os.MkdirAll(dbDir, 0700); err != nil {
-		return xerrors.Errorf("failed to mkdir: %w", err)
+		return eb.Wrapf(err, "failed to mkdir")
 	}
 	dbPath := Path(dbDir)
+	eb = eb.With("db_path", dbPath)
 
 	// bbolt sometimes occurs the fatal error of "unexpected fault address".
 	// In that case, the local DB should be broken and needs to be removed.
@@ -92,7 +94,7 @@ func Init(dbDir string, opts ...Option) (err error) {
 
 	db, err = bolt.Open(dbPath, 0644, dbOptions.boltOptions)
 	if err != nil {
-		return xerrors.Errorf("failed to open db: %w", err)
+		return eb.Wrapf(err, "failed to open db")
 	}
 	return nil
 }
@@ -108,7 +110,7 @@ func Close() error {
 		return nil
 	}
 	if err := db.Close(); err != nil {
-		return xerrors.Errorf("failed to close DB: %w", err)
+		return oops.Wrapf(err, "failed to close DB")
 	}
 	return nil
 }
@@ -120,39 +122,41 @@ func (dbc Config) Connection() *bolt.DB {
 func (dbc Config) BatchUpdate(fn func(tx *bolt.Tx) error) error {
 	err := db.Batch(fn)
 	if err != nil {
-		return xerrors.Errorf("error in batch update: %w", err)
+		return oops.Wrapf(err, "batch update error")
 	}
 	return nil
 }
 
 func (dbc Config) put(tx *bolt.Tx, bktNames []string, key string, value interface{}) error {
 	if len(bktNames) == 0 {
-		return xerrors.Errorf("empty bucket name")
+		return oops.Errorf("empty bucket name")
 	}
 
+	eb := oops.With("bucket_names", bktNames)
 	bkt, err := tx.CreateBucketIfNotExists([]byte(bktNames[0]))
 	if err != nil {
-		return xerrors.Errorf("failed to create '%s' bucket: %w", bktNames[0], err)
+		return eb.Wrapf(err, "failed to create bucket")
 	}
 
 	for _, bktName := range bktNames[1:] {
 		bkt, err = bkt.CreateBucketIfNotExists([]byte(bktName))
 		if err != nil {
-			return xerrors.Errorf("failed to create a bucket: %w", err)
+			return eb.Wrapf(err, "failed to create bucket")
 		}
 	}
 	v, err := json.Marshal(value)
 	if err != nil {
-		return xerrors.Errorf("failed to unmarshal JSON: %w", err)
+		return eb.Wrapf(err, "json marshal error")
 	}
 
 	return bkt.Put([]byte(key), v)
 }
 
 func (dbc Config) get(bktNames []string, key string) (value []byte, err error) {
+	eb := oops.With("bucket_names", bktNames)
 	err = db.View(func(tx *bolt.Tx) error {
 		if len(bktNames) == 0 {
-			return xerrors.Errorf("empty bucket name")
+			return eb.Errorf("empty bucket name")
 		}
 
 		bkt := tx.Bucket([]byte(bktNames[0]))
@@ -174,7 +178,7 @@ func (dbc Config) get(bktNames []string, key string) (value []byte, err error) {
 		return nil
 	})
 	if err != nil {
-		return nil, xerrors.Errorf("failed to get data from db: %w", err)
+		return nil, eb.Wrapf(err, "failed to get data from db")
 	}
 	return value, nil
 }
@@ -185,8 +189,9 @@ type Value struct {
 }
 
 func (dbc Config) forEach(bktNames []string) (map[string]Value, error) {
+	eb := oops.With("bucket_names", bktNames)
 	if len(bktNames) < 2 {
-		return nil, xerrors.Errorf("bucket must be nested: %v", bktNames)
+		return nil, eb.Errorf("bucket must be nested")
 	}
 	rootBucket, nestedBuckets := bktNames[0], bktNames[1:]
 
@@ -243,13 +248,13 @@ func (dbc Config) forEach(bktNames []string) (map[string]Value, error) {
 				return nil
 			})
 			if err != nil {
-				return xerrors.Errorf("db foreach error: %w", err)
+				return eb.Wrapf(err, "db foreach error")
 			}
 		}
 		return nil
 	})
 	if err != nil {
-		return nil, xerrors.Errorf("failed to get all key/value in the specified bucket: %w", err)
+		return nil, eb.Wrapf(err, "failed to get all key/value in the specified bucket")
 	}
 	return values, nil
 }
@@ -257,7 +262,7 @@ func (dbc Config) forEach(bktNames []string) (map[string]Value, error) {
 func (dbc Config) deleteBucket(bucketName string) error {
 	return db.Update(func(tx *bolt.Tx) error {
 		if err := tx.DeleteBucket([]byte(bucketName)); err != nil {
-			return xerrors.Errorf("failed to delete bucket: %w", err)
+			return oops.With("bucket_name", bucketName).Wrapf(err, "failed to delete bucket")
 		}
 		return nil
 	})
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -135,13 +135,13 @@ func (dbc Config) put(tx *bolt.Tx, bktNames []string, key string, value interfac
 	eb := oops.With("bucket_names", bktNames)
 	bkt, err := tx.CreateBucketIfNotExists([]byte(bktNames[0]))
 	if err != nil {
-		return eb.Wrapf(err, "failed to create bucket")
+		return eb.With("bucket_name", bktNames[0]).Wrapf(err, "failed to create bucket")
 	}
 
 	for _, bktName := range bktNames[1:] {
 		bkt, err = bkt.CreateBucketIfNotExists([]byte(bktName))
 		if err != nil {
-			return eb.Wrapf(err, "failed to create bucket")
+			return eb.With("bucket_name", bktName).Wrapf(err, "failed to create bucket")
 		}
 	}
 	v, err := json.Marshal(value)
--- a/pkg/vulndb/db_test.go
+++ b/pkg/vulndb/db_test.go
@@ -1,6 +1,7 @@
 package vulndb_test
 
 import (
+	"errors"
 	"path/filepath"
 	"strings"
 	"testing"
@@ -8,7 +9,6 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"golang.org/x/xerrors"
 	"k8s.io/utils/clock"
 	fake "k8s.io/utils/clock/testing"
 
@@ -27,7 +27,7 @@ func (f fakeVulnSrc) Name() types.SourceID { return "fake" }
 
 func (f fakeVulnSrc) Update(dir string) error {
 	if strings.Contains(dir, "bad") {
-		return xerrors.New("something bad")
+		return errors.New("something bad")
 	}
 	return nil
 }
@@ -71,7 +71,7 @@ func TestTrivyDB_Insert(t *testing.T) {
 			args: args{
 				targets: []string{"unknown"},
 			},
-			wantErr: "unknown is not supported",
+			wantErr: "target not supported",
 		},
 		{
 			name: "sad path: update error",
@@ -82,7 +82,7 @@ func TestTrivyDB_Insert(t *testing.T) {
 			args: args{
 				targets: []string{"fake"},
 			},
-			wantErr: "fake update error",
+			wantErr: "update error",
 		},
 	}
 
@@ -100,8 +100,7 @@ func TestTrivyDB_Insert(t *testing.T) {
 			c := vulndb.New(cacheDir, outputDir, 12*time.Hour, vulndb.WithClock(tt.fields.clock), vulndb.WithVulnSrcs(vulnsrcs))
 			err := c.Insert(tt.args.targets)
 			if tt.wantErr != "" {
-				require.NotNil(t, err)
-				assert.Contains(t, err.Error(), tt.wantErr)
+				assert.ErrorContains(t, err, tt.wantErr)
 				return
 			}
 			require.NoError(t, err)
@@ -172,7 +171,7 @@ func TestTrivyDB_Build(t *testing.T) {
 				"testdata/fixtures/happy/vulnerability-detail.yaml",
 				"testdata/fixtures/sad/advisory-detail.yaml",
 			},
-			wantErr: "failed to unmarshall the advisory detail",
+			wantErr: "json unmarshal error",
 		},
 		{
 			name: "missing advisory detail",
@@ -193,8 +192,7 @@ func TestTrivyDB_Build(t *testing.T) {
 			full := vulndb.New(cacheDir, dbDir, 12*time.Hour)
 			err := full.Build(nil)
 			if tt.wantErr != "" {
-				require.NotNil(t, err)
-				assert.Contains(t, err.Error(), tt.wantErr)
+				assert.ErrorContains(t, err, tt.wantErr)
 				return
 			}
 			require.NoError(t, err)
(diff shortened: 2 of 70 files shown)


knqyf263:
@DmitriyLewen It's just refactoring, but there are many changes. The review is not urgent. Take your time.

pkg/vulnsrc/azure/azure.go
```diff
@@ -199,30 +203,31 @@ func resolveTests(dir string) (map[string]resolvedTest, error) {
...
 	if !ok {
-		return resolvedTest{}, xerrors.Errorf("invalid tests data, can't find ovalstate ref %s, test ref: %s",
-			test.State.StateRef, test.ID)
+		return resolvedTest{}, eb.Errorf("invalid tests data, can't find ovalstate ref")
 	}
 
+	eb = eb.With("data_type", state.Evr.Datatype)
 	if state.Evr.Datatype != "evr_string" {
-		return resolvedTest{}, xerrors.Errorf("state data type (%s): %w", state.Evr.Datatype, ErrNotSupported)
+		return resolvedTest{}, eb.Wrapf(ErrNotSupported, "state data type")
```
DmitriyLewen, line 230: nil:
we need `data_type` only for this error
```suggestion
		return resolvedTest{}, eb.With("data_type", state.Evr.Datatype).Wrapf(ErrNotSupported, "state data type")
```

pkg/vulnsrc/azure/azure.go
```diff
@@ -231,7 +236,7 @@ func followTestRefs(test oval.RpmInfoTest, objects map[string]string, states map
 	}
 
 	if state.Evr.Operation != string(lte) && state.Evr.Operation != string(lt) {
-		return resolvedTest{}, xerrors.Errorf("state operation (%s): %w", state.Evr.Operation, ErrNotSupported)
+		return resolvedTest{}, eb.Wrapf(ErrNotSupported, "state operation")
```
DmitriyLewen, line 239:
```suggestion
		return resolvedTest{}, eb.With("operation", state.Evr.Operation).Wrapf(ErrNotSupported, "state operation")
```

pkg/vulnsrc/azure/oval/oval.go
```diff
@@ -5,46 +5,50 @@ import (
...
 
 	return defs, nil
 }
 
 func ParseTests(dir string) (Tests, error) {
+	eb := oops.In("azure").Tags("oval").With("dir", dir)
 	var tests Tests
 	if err := utils.UnmarshalJSONFile(&tests, filepath.Join(dir, "tests", "tests.json")); err != nil {
-		return tests, xerrors.Errorf("failed to unmarshal tests: %w", err)
+		return tests, eb.Wrapf(err, "json unmarshal error")
```
DmitriyLewen, line 42: OopsErrorBuilder is still auxiliary code
I would prefer not to "draw unnecessary attention" to this variable (not to create `eb` variable) if the error is used once
```suggestion
		return tests, oops.In("azure").Tags("oval").With("dir", dir).Wrapf(err, "json unmarshal error")
```
same for `ParseObjects` and `ParseStates`

pkg/vulnsrc/chainguard/chainguard.go
```diff
@@ -44,21 +44,25 @@ func (vs VulnSrc) Name() types.SourceID {
...
 		return nil
 	})
 	if err != nil {
-		return xerrors.Errorf("error in Chainguard walk: %w", err)
+		return eb.Wrapf(err, "walk error")
 	}
 
 	if err = vs.save(advisories); err != nil {
-		return xerrors.Errorf("error in Chainguard save: %w", err)
+		return eb.Wrapf(err, "save error")
```
DmitriyLewen, line 65: nit:
```suggestion
		return eb.Wrapf(err, "save advisories error")
```

pkg/vulnsrc/debian/debian.go
```diff
@@ -530,9 +541,12 @@ func severityFromUrgency(urgency string) types.Severity {
 
 func (vs VulnSrc) parseDistributions(rootDir string) error {
 	vs.logger.Info("Parsing distributions...")
-	f, err := os.Open(filepath.Join(rootDir, distributionsFile))
+	filePath := filepath.Join(rootDir, distributionsFile)
+	eb := oops.With("root_dir", rootDir).With("file_path", filePath)
```
DmitriyLewen, line 545: maybe just file_path will be enough?

pkg/vulnsrc/debian/debian.go
```diff
@@ -555,21 +569,25 @@ func (vs VulnSrc) parseDistributions(rootDir string) error {
...
 	for code := range vs.distributions {
 		codePath := filepath.Join(dir, code)
 		if ok, _ := utils.Exists(codePath); !ok {
 			continue
 		}
+		eb = eb.With("code_name", code)
 
 		vs.logger.Info("Parsing sources...", log.String("code", code))
 		err := utils.FileWalk(codePath, func(r io.Reader, path string) error {
+			eb := eb.With("file_path", path)
```
DmitriyLewen, line 582: same

pkg/vulnsrc/node/node.go
```diff
@@ -90,42 +90,46 @@ func (vs VulnSrc) Name() types.SourceID {
...
 	if err != nil {
-		return xerrors.Errorf("batch update failed: %w", err)
+		return eb.Wrapf(err, "batch update failed")
 	}
 	return nil
 }
 
 func (vs VulnSrc) walk(tx *bolt.Tx, root string) error {
 	return filepath.Walk(root, func(path string, info os.FileInfo, err error) error {
+		eb := oops.With("path", path)
```
DmitriyLewen, line 122: nit:
```suggestion
		eb := oops.With("file_path", path)
```

pkg/vulnsrc/osv/osv.go
```diff
@@ -90,17 +91,17 @@ func (o OSV) Update(root string) error {
 		}
 		var entry Entry
 		if err := json.NewDecoder(r).Decode(&entry); err != nil {
-			return xerrors.Errorf("JSON decode error (%s): %w", path, err)
+			return oops.With("path", path).Wrapf(err, "JSON decode error")
```
DmitriyLewen, line 94: nit:
```suggestion
			return oops.With("file_path", path).Wrapf(err, "json decode error")
```

pkg/vulnsrc/photon/photon.go
```diff
@@ -45,23 +45,26 @@ func (vs VulnSrc) Name() types.SourceID {
...
+	eb := oops.In("photon").With("root_dir", rootDir)
 
 	var cves []PhotonCVE
 	err := utils.FileWalk(rootDir, func(r io.Reader, path string) error {
+		eb := eb.With("file_path", path)
+
 		var cve PhotonCVE
 		if err := json.NewDecoder(r).Decode(&cve); err != nil {
-			return xerrors.Errorf("failed to decode Photon JSON: %w", err)
+			return eb.Wrapf(err, "json decode error")
```
DmitriyLewen, line 56:
```suggestion
			return eb.With("file_path", path).Wrapf(err, "json decode error")
```

pkg/vulnsrc/redhat-oval/parse.go
```diff
@@ -16,22 +16,24 @@ type rpmInfoTest struct {
...
 	f, err := os.Open(fileName)
 	if err != nil {
-		return xerrors.Errorf("unable to open a file (%s): %w", fileName, err)
+		return eb.Wrapf(err, "unable to open a file")
 	}
 	defer f.Close()
 
 	if err = json.NewDecoder(f).Decode(v); err != nil {
-		return xerrors.Errorf("failed to decode Red Hat OVAL JSON: %w", err)
+		return eb.Wrapf(err, "decode error")
```
DmitriyLewen, line 28:
```suggestion
		return eb.Wrapf(err, "json decode error")
```

pkg/vulnsrc/suse-cvrf/suse-cvrf.go
```diff
@@ -70,30 +70,33 @@ func (vs VulnSrc) Name() types.SourceID {
...
+		return eb.Errorf("unknown distribution")
 	}
 
 	var cvrfs []SuseCvrf
 	err := utils.FileWalk(rootDir, func(r io.Reader, path string) error {
+		eb := eb.With("file_path", path)
 		var cvrf SuseCvrf
 		if err := json.NewDecoder(r).Decode(&cvrf); err != nil {
-			return xerrors.Errorf("failed to decode SUSE CVRF JSON: %w %+v", err, cvrf)
+			return eb.Wrapf(err, "json decode error")
```
DmitriyLewen, line 89:
```suggestion
			return eb.With("file_path", path).Wrapf(err, "json decode error")
```

pkg/db/db.go
```diff
@@ -92,7 +94,7 @@ func Init(dbDir string, opts ...Option) (err error) {
 
 	db, err = bolt.Open(dbPath, 0644, dbOptions.boltOptions)
 	if err != nil {
-		return xerrors.Errorf("failed to open db: %w", err)
+		return eb.Wrapf(err, "failed to open db")
```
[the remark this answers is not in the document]
knqyf263: I actually considered it and concluded to display both since the path calculation logic is in another function.
`Path` is very simple now, but I'm not sure we want to think about the logic of other functions when adding contextual attributes, in general.
What if `Path` is like `return filepath.Join(dbDir, "../../foo/bar/db/trivy.db")`? I think we want to show both values in that case.
I agree they are duplicated, but it depends on the logic of `Path`. Also, the behavior may change in the future.

In addition, I don't think it's bad to show more information for developers as these contexts will be displayed only with `--debug` in Trivy.

pkg/vulndb/db_test.go
```diff
@@ -27,7 +27,7 @@ func (f fakeVulnSrc) Name() types.SourceID { return "fake" }
 
 func (f fakeVulnSrc) Update(dir string) error {
 	if strings.Contains(dir, "bad") {
-		return xerrors.New("something bad")
+		return errors.New("something bad")
```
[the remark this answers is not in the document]
knqyf263: No need to use oops here as it's for testing.

pkg/vulnsrc/amazon/amazon.go
```diff
@@ -101,10 +103,11 @@ func (vs *VulnSrc) walkFunc(r io.Reader, path string) error {
 		vs.logger.Warn("Unsupported Amazon version", "version", version)
 		return nil
 	}
+	eb = eb.With("version", version)
 
 	var alas ALAS
 	if err := json.NewDecoder(r).Decode(&alas); err != nil {
-		return xerrors.Errorf("failed to decode Amazon JSON: %w", err)
+		return eb.Wrapf(err, "json decode error")
```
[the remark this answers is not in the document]
knqyf263: Those contexts are already added above.

pkg/vulnsrc/azure/oval/oval.go
knqyf263: For public functions, I always add domains at the top of the function, otherwise, we can easily forget to add that when we add more lines.

I documented it in the PR description. Please let me know if you have a better idea for guidelines.
https://github.com/aquasecurity/trivy-db/pull/488

pkg/vulnsrc/amazon/amazon.go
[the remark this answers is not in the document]
DmitriyLewen: sorry, i should have been clearer.
i suggest to remove this context above and leave it only in this line (so we use this context once)

pkg/vulnsrc/debian/debian.go
DmitriyLewen: `file_path` contains `root_dir`.
so we are almost duplicating the information
изображение

pkg/db/db.go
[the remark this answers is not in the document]
DmitriyLewen: k, you convinced me 😄 
Let's that both values.

pkg/vulnsrc/debian/debian.go
knqyf263: Regarding, https://github.com/aquasecurity/trivy-db/pull/488#discussion_r1930120127, `dir` and `path` are defined apart.
And `dir` (or `codePath`) is probably helpful for enriching errors from `utils.FileWalk`.
For example, when there is no permission on `codePath`, the context `file_path` will not be added to this error.
https://github.com/knqyf263/trivy-db/blob/b7b971c8ac657b463e70af06241a678d10bb7c7d/pkg/vulnsrc/debian/debian.go#L623

I feel like it's challenging to judge when the context should be added. That's why I took an approach of adding contexts as much as possible.

pkg/vulnsrc/debian/debian.go
DmitriyLewen: But this is a case like in https://github.com/aquasecurity/trivy-db/pull/488#discussion_r1930091841
so let's leave both paths.

knqyf263: Yes, but this is a super simple case. `filepath.Join` is called one line above. I'm not sure we want to show both values, even in this simple case. I added that because I tried to be consistent in the project.

pkg/vulnsrc/azure/oval/oval.go
DmitriyLewen: Okay, it make sense, let's use your way.

pkg/vulnsrc/debian/debian.go
knqyf263: After all, I have come to the following conclusion.

1. Add attributes as much as possible
2. Showing more information is not bad

However, I'm not confident and open to feedback.

DmitriyLewen: perhaps the slog-type interface confused me a bit (the habit of making logs minimal).
After listening to your arguments, I realized that I agree with you.
In "normal" use, additional information (even unnecessary) does not bother us, but in case of errors - it can help (even if the information is partially duplicated (e.g. see small typo))

So I agree with your approach now.

knqyf263: After second thoughts, I just remembered I faced a problem with `filepath.Join` on Windows. `filepath.Join` transparently calls `Clean` and it caused a volume (e.g. `C:\`) problem unexpectedly. I was debugging that by outputting the values before and after `filepath.Join`, and figured out the problem.
https://github.com/golang/go/blob/608acff8479640b00c85371d91280b64f5ec9594/src/path/filepath/path_windows.go#L113

The point here is not whether the function is actually used on Windows, but that it can still be helpful to have both outputs since even a one-line function, like `filepath.Join` can change the value unexpectedly.

knqyf263: I'd keep it for now, but we don't need to stick to this idea. If it turns out many values would be noisy for debugging, we can remove them.

pkg/vulnsrc/amazon/amazon.go
```diff
@@ -117,25 +120,29 @@ func (vs VulnSrc) save() error {
...
-			return xerrors.Errorf("failed to put data source: %w", err)
+			return eb.Wrapf(err, "failed to put data source")
 		}
 		for _, alas := range alasList {
 			for _, cveID := range alas.CveIDs {
 				for _, pkg := range alas.Packages {
 					advisory := types.Advisory{
 						FixedVersion: utils.ConstructVersion(pkg.Epoch, pkg.Version, pkg.Release),
 					}
+					eb := eb.With("vuln_id", cveID).With("package_name", pkg.Name).With("fixed_version", advisory.FixedVersion)
```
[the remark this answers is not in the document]
knqyf263: I realized I added duplicate contexts in both callers and callees. So, I delete contexts in callers.
https://github.com/aquasecurity/trivy-db/pull/488/commits/fd314d0bd0c57cb219e75b349f1ea90883fec1ce

I also documented in the PR description (see No.4 in guidelines).

pkg/db/db.go
```diff
@@ -120,39 +122,41 @@ func (dbc Config) Connection() *bolt.DB {
...
 	if len(bktNames) == 0 {
-		return xerrors.Errorf("empty bucket name")
+		return oops.Errorf("empty bucket name")
 	}
 
+	eb := oops.With("bucket_names", bktNames)
 	bkt, err := tx.CreateBucketIfNotExists([]byte(bktNames[0]))
 	if err != nil {
-		return xerrors.Errorf("failed to create '%s' bucket: %w", bktNames[0], err)
+		return eb.Wrapf(err, "failed to create bucket")
```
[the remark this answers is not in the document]
knqyf263: Done https://github.com/aquasecurity/trivy-db/pull/488/commits/0749c76366dc29d4a6161079dc31de423f36ac69

knqyf263:
@DmitriyLewen All resolved.

pkg/vulnsrc/amazon/amazon.go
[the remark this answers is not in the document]
DmitriyLewen: > I added duplicate contexts in both callers and callees
I found 1 place with duplicates.

Fixed in https://github.com/aquasecurity/trivy-db/pull/488/commits/764a7590862016ca84c0616fe255b08f43bf03ea

DmitriyLewen (approved):
LGTM

knqyf263:
Trivy is next, but it's huge 😱

knqyf263:
It sounds reasonable. I'll work on that when I have time.

DmitriyLewen:
If you have the opportunity, create a issue with sub-issues that I can assign to myself so that our paths don't cross and we don't do same work.

knqyf263:
Sure. It's a good opportunity to try sub-issues in GitHub.

PR #501, opened 2025-02-25
fix(db): return error if panic occurs when opening database

DmitriyLewen:
## Description
if `bbolt` returned panic when initializing db we just create new empty database.

This results in Trivy doesn't show vulnerabilities.
See https://github.com/aquasecurity/trivy/discussions/7758#discussioncomment-12309744

we should return error if we couldn't open trivy-db

example (https://gist.github.com/knqyf263/7ce8e504b6d1f9867e6898f83c89acba is used to reproduce error):

before:
```
Testing race condition with downloader.Download
Starting concurrent downloads to the same destination file...
Download 1: Adding fixed delay of 100 ms
Download 0: Starting
Download 1: Starting
Download 0: Completed
Download 1: Completed

Checking the destination file:
File: /var/folders/8m/p1341n2941jbyc5gm7357x5c0000gn/T/trivy-poc-1180293935/dst/trivy.db, Size: 65536 bytes
WARNING - DB is EMPTY!

Result: Empty file detected! Race condition successfully reproduced.

```

after:
```
Testing race condition with downloader.Download
Starting concurrent downloads to the same destination file...
Download 1: Adding fixed delay of 100 ms
Download 0: Starting
Download 1: Starting
Download 0: Completed
2025/02/25 15:27:54 got error initializing db: failed to open db: runtime error: invalid memory address or nil pointer dereference

```

### DIFF
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -87,7 +87,7 @@ func Init(dbDir string, opts ...Option) (err error) {
 			if err = os.Remove(dbPath); err != nil {
 				return
 			}
-			err = eb.Errorf("failed to open db: %s", r)
+			err = eb.Errorf("db corrupted: %s", r)
 		}
 		debug.SetPanicOnFault(false)
 	}()
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -8,11 +8,10 @@ import (
 	"runtime/debug"
 	"strings"
 
-	"github.com/samber/oops"
-	bolt "go.etcd.io/bbolt"
-
 	"github.com/aquasecurity/trivy-db/pkg/log"
 	"github.com/aquasecurity/trivy-db/pkg/types"
+	"github.com/samber/oops"
+	bolt "go.etcd.io/bbolt"
 )
 
 type CustomPut func(dbc Operation, tx *bolt.Tx, adv interface{}) error
@@ -80,14 +79,11 @@ func Init(dbDir string, opts ...Option) (err error) {
 	eb = eb.With("db_path", dbPath)
 
 	// bbolt sometimes occurs the fatal error of "unexpected fault address".
-	// In that case, the local DB should be broken and needs to be removed.
+	// In that case, the local DB should be broken and we need to return error.
 	debug.SetPanicOnFault(true)
 	defer func() {
 		if r := recover(); r != nil {
-			if err = os.Remove(dbPath); err != nil {
-				return
-			}
-			db, err = bolt.Open(dbPath, 0644, dbOptions.boltOptions)
+			err = eb.Errorf("failed to open db: %s", r)
 		}
 		debug.SetPanicOnFault(false)
 	}()
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -8,10 +8,11 @@ import (
 	"runtime/debug"
 	"strings"
 
-	"github.com/aquasecurity/trivy-db/pkg/log"
-	"github.com/aquasecurity/trivy-db/pkg/types"
 	"github.com/samber/oops"
 	bolt "go.etcd.io/bbolt"
+
+	"github.com/aquasecurity/trivy-db/pkg/log"
+	"github.com/aquasecurity/trivy-db/pkg/types"
 )
 
 type CustomPut func(dbc Operation, tx *bolt.Tx, adv interface{}) error
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -80,10 +80,13 @@ func Init(dbDir string, opts ...Option) (err error) {
 	eb = eb.With("db_path", dbPath)
 
 	// bbolt sometimes occurs the fatal error of "unexpected fault address".
-	// In that case, the local DB should be broken and we need to return error.
+	// In that case, the local DB should be broken and needs to be removed.
 	debug.SetPanicOnFault(true)
 	defer func() {
 		if r := recover(); r != nil {
+			if err = os.Remove(dbPath); err != nil {
+				return
+			}
 			err = eb.Errorf("failed to open db: %s", r)
 		}
 		debug.SetPanicOnFault(false)
--- a/pkg/db/db_test.go
+++ b/pkg/db/db_test.go
@@ -6,25 +6,26 @@ import (
 	"path/filepath"
 	"testing"
 
+	"github.com/aquasecurity/trivy-db/pkg/db"
 	"github.com/stretchr/testify/require"
 	"go.etcd.io/bbolt"
-
-	"github.com/aquasecurity/trivy-db/pkg/db"
 )
 
 func TestInit(t *testing.T) {
 	tests := []struct {
-		name   string
-		dbPath string
-		dbOpts *bbolt.Options
+		name    string
+		dbPath  string
+		wantErr string
+		dbOpts  *bbolt.Options
 	}{
 		{
 			name:   "normal db",
 			dbPath: "testdata/normal.db",
 		},
 		{
-			name:   "broken db",
-			dbPath: "testdata/broken.db",
+			name:    "broken db",
+			dbPath:  "testdata/broken.db",
+			wantErr: "invalid memory address or nil pointer dereference",
 		},
 		{
 			name:   "no db",
@@ -46,6 +47,11 @@ func TestInit(t *testing.T) {
 			}
 
 			err := db.Init(tmpDir, db.WithBoltOptions(tt.dbOpts))
+			if tt.wantErr != "" {
+				require.ErrorContains(t, err, tt.wantErr)
+				return
+			}
+
 			require.NoError(t, err)
 		})
 	}
--- a/pkg/db/db_test.go
+++ b/pkg/db/db_test.go
@@ -6,9 +6,10 @@ import (
 	"path/filepath"
 	"testing"
 
-	"github.com/aquasecurity/trivy-db/pkg/db"
 	"github.com/stretchr/testify/require"
 	"go.etcd.io/bbolt"
+
+	"github.com/aquasecurity/trivy-db/pkg/db"
 )
 
 func TestInit(t *testing.T) {


pkg/db/db.go
```diff
@@ -80,14 +80,11 @@ func Init(dbDir string, opts ...Option) (err error) {
 	eb = eb.With("db_path", dbPath)
 
 	// bbolt sometimes occurs the fatal error of "unexpected fault address".
-	// In that case, the local DB should be broken and needs to be removed.
```
knqyf263, line 83: The database needs to be removed. Otherwise, Trivy always fails due to the broken database.

DmitriyLewen: hm... it make sense
updated in https://github.com/aquasecurity/trivy-db/pull/501/commits/6100c2e3d9837c28b1f7f88d5e7c5358d4c258c1

pkg/db/db.go
```diff
@@ -87,7 +87,7 @@ func Init(dbDir string, opts ...Option) (err error) {
 			if err = os.Remove(dbPath); err != nil {
 				return
 			}
-			db, err = bolt.Open(dbPath, 0644, dbOptions.boltOptions)
+			err = eb.Errorf("failed to open db: %s", r)
```
knqyf263, line 90: If it just fails to open a database, it returns an error here.
https://github.com/aquasecurity/trivy-db/blob/6100c2e3d9837c28b1f7f88d5e7c5358d4c258c1/pkg/db/db.go#L96-L98

As far as I know, it panics only when the database is broken.

```suggestion
			err = eb.Errorf("db corrupted: %s", r)
```

DmitriyLewen: I don't have as much experience as you, so I'll trust your information.

pkg/db/db.go
DmitriyLewen: Yes. I got it right away.
It will also help users - they won't have to manually delete the broken database.

knqyf263:
Can you please try "Ctrl+C" while [copying the database](https://github.com/aquasecurity/trivy/blob/f987e4157494434f6e4e4566fedfedda92167565/pkg/oci/artifact.go#L197), after the database is successfully downloaded into the temporary file, several times and see if Trivy returns an error? I want to make sure Trivy will not silence the panic anymore.

DmitriyLewen:
I am not sure that case with `Ctrl+C` possible.
`go-getter` copies db file first (at least that's how it was on my local PC).
Without `metadata.json` file Trivy still downloads trivy-db.

But i created metadata.json file manually and tested:

only broken db:
```
➜  ls -hl /Users/dmitriy/Library/Caches/trivy/db
total 257984
-rw-r--r--@ 1 dmitriy  staff   126M Feb 25 16:56 trivy.db
➜  ./trivy image alpine:3.20
2025-02-25T17:03:21+06:00       INFO    Loaded  file_path="trivy.yaml"
2025-02-25T17:03:22+06:00       INFO    [vulndb] Need to update DB
2025-02-25T17:03:22+06:00       INFO    [vulndb] Downloading vulnerability DB...
2025-02-25T17:03:22+06:00       INFO    [vulndb] Downloading artifact...        repo="mirror.gcr.io/aquasec/trivy-db:2"
17.44 MiB / 59.39 MiB [------------------------------------------->_________________________________________________________________________________________________________] 29.36% 3.83 MiB p/s ETA 10s^C

```
broken db with metadata.json file:
```
➜  ls -hl /Users/dmitriy/Library/Caches/trivy/db
total 257992
-rw-r--r--@ 1 dmitriy  staff   149B Feb 25 16:57 metadata.json
-rw-r--r--@ 1 dmitriy  staff   126M Feb 25 16:56 trivy.db
➜  ./trivy image alpine:3.20                    
2025-02-25T17:04:10+06:00       INFO    Loaded  file_path="trivy.yaml"
2025-02-25T17:04:11+06:00       FATAL   Fatal error     init error: DB error: error in vulnerability DB initialize: db corrupted: runtime error: invalid memory address or nil pointer dereference
➜  ls -hl /Users/dmitriy/Library/Caches/trivy/db
total 8
-rw-r--r--@ 1 dmitriy  staff   149B Feb 25 16:57 metadata.json
➜  ./trivy image alpine:3.20                    
2025-02-25T17:04:25+06:00       INFO    Loaded  file_path="trivy.yaml"
2025-02-25T17:04:25+06:00       INFO    [vuln] Vulnerability scanning is enabled
2025-02-25T17:04:25+06:00       INFO    [secret] Secret scanning is enabled
2025-02-25T17:04:25+06:00       INFO    [secret] If your scanning is slow, please try '--scanners vuln' to disable secret scanning
2025-02-25T17:04:25+06:00       INFO    [secret] Please see also https://aquasecurity.github.io/trivy/dev/docs/scanner/secret#recommendation for faster secret detection
2025-02-25T17:04:31+06:00       INFO    Detected OS     family="alpine" version="3.20.6"
2025-02-25T17:04:31+06:00       INFO    [alpine] Detecting vulnerabilities...   os_version="3.20" repository="3.20" pkg_num=14
2025-02-25T17:04:31+06:00       INFO    Number of language-specific files       num=0

alpine:3.20 (alpine 3.20.6)

Total: 0 (UNKNOWN: 0, LOW: 0, MEDIUM: 0, HIGH: 0, CRITICAL: 0)

```

We removed `trivy.db` file. But for 2nd scan we still create empty DB.
I think we need to check not only `metadata.json` file in Trivy:
https://github.com/aquasecurity/trivy/blob/5ba9a83a447c4f9e577ae6235c315df71f50b452/pkg/db/db.go#L98-L130
Or we can remove `dbDir` in trivy-db. But case with Trivy looks more logical for me.

knqyf263:
Oh, that's right. The reason for [deleting metadata.json](https://github.com/aquasecurity/trivy/blob/50364b836fde5b71974655e56054ca7fdbbdf6a5/pkg/db/db.go#L157-L160) before downloading the database was to handle Ctrl-C. However, if two processes run simultaneously, the second process ends up corrupting the database, and the first process resets it as a result. This method doesn't effectively avoid the issue. 

Although I think your manual test with manifest.json is sufficient, could you also test by running two processes at the same time, just to be sure?

DmitriyLewen:
> Although I think your manual test with manifest.json is sufficient, could you also test by running two processes at the same time, just to be sure?

I tried this today, but I couldn't get both processes to finish downloading at the same time.
But I'll try again tomorrow and let you know.

DmitriyLewen:
@knqyf263 Trivy returns error (i ran 2 trivy in script, so logs contain messages from both processes):
```
➜ ./script.sh 
2025-02-26T11:05:42+06:00	INFO	[vulndb] Need to update DB
2025-02-26T11:05:42+06:00	INFO	[vulndb] Downloading vulnerability DB...
2025-02-26T11:05:42+06:00	INFO	[vulndb] Downloading artifact...	repo="mirror.gcr.io/aquasec/trivy-db:2"
2025-02-26T11:05:42+06:00	INFO	[vulndb] Need to update DB
2025-02-26T11:05:42+06:00	INFO	[vulndb] Downloading vulnerability DB...
2025-02-26T11:05:42+06:00	INFO	[vulndb] Downloading artifact...	repo="mirror.gcr.io/aquasec/trivy-db:2"
59.46 MiB / 59.46 MiB [-----------------------------] 100.00% 10.81 MiB p/s 5.7s
2025-02-26T11:05:49+06:00	INFO	[vulndb] Artifact successfully downloaded	repo="mirror.gcr.io/aquasec/trivy-db:2"
2025-02-26T11:05:49+06:00	FATAL	Fatal error	init error: DB error: error in vulnerability DB initialize: db corrupted: runtime error: invalid memory address or nil pointer dereference
59.46 MiB / 59.46 MiB [-----------------------------] 100.00% 11.01 MiB p/s 5.6s
2025-02-26T11:05:49+06:00	FATAL	Fatal error	init error: DB error: failed to download vulnerability DB: OCI artifact error: failed to download vulnerability DB: failed to download artifact from mirror.gcr.io/aquasec/trivy-db:2: oci download error: download error: failed to download /var/folders/8m/p1341n2941jbyc5gm7357x5c0000gn/T/trivy2263882786/db.tar.gz: chmod /Users/dmitriy/Library/Caches/trivy/db/trivy.db: no such file or directory

➜ ls -hl /Users/dmitriy/Library/Caches/trivy/db
total 8
-rw-r--r--@ 1 dmitriy  staff   150B 26 фев 11:05 metadata.json
```

But cache dir contains `metadata.json`. So we need to do something with it.

knqyf263:
Indeed, if metadata.json exists and contains a new updated_at, the [NeedsUpdate](https://github.com/aquasecurity/trivy/blob/50364b836fde5b71974655e56054ca7fdbbdf6a5/pkg/db/db.go#L98-L129) function returns false, which ultimately leads to an empty database being used. There are several ideas in my mind.

1. If the database is corrupted, delete metadata.json.
2. Update `NeedsUpdate` to check for the existence of the database file and return true if the file is missing.
3. Create a lock file indicating an update is in progress.

DmitriyLewen:
> If the database is corrupted, delete metadata.json.

I don't like this because we only work with DB file in `db` package. I don't want to merge logic for trivy.db and metadata.json files. Just special for one case.

> Create a lock file indicating an update is in progress.

This might be an overkill. It looks like point 2 would be enough.

> Update NeedsUpdate to check for the existence of the database file and return true if the file is missing.

This is what I thought at first. I would go that way.

knqyf263:
The advantage of option 3 is that Trivy works correctly even when run simultaneously. Currently, it returns an error, so in terms of UX, option 3 is the best. However, running multiple instances of Trivy in standalone mode simultaneously is not a recommended use case, so it might be sufficient to inform the user of this when we detect the corrupted database.

DmitriyLewen:
I think that most users who run more than one Trivy process use client/server mode.
Otherwise we would have encountered a similar problem earlier.
But I do not exclude cases when users do not check the number of vulnerabilities.

Therefore, I suggest starting with point 2 and waiting for errors from users. If we see a large number of errors, we will create a lock file for loading the DB.

knqyf263 (approved):

knqyf263:
@DmitriyLewen You can merge the PR whenever.

Issue #522, opened 2025-04-29
PublishedDate and LastModifiedDate for vulnerabilities are only taken from NVD source


mhbardsley:
It looks like the Trivy DB will only populate the `PublishedDate` and `LastModifiedDate` for a vulnerability if it has been ingested from NVD [1]. As far as I can tell, there doesn't seem to be a good reason for this. Indeed, many of the vulnerability sources Trivy pulls from expose this information, so it should be readily available. Solving this will benefit any tools that use Trivy scan results as a proxy for vulnerability metadata.

**Current state**
(e.g. using `boltcli` [2] on any recent version of `trivy.db`):

```
keyvalues vulnerability CVE-2025-46688
```

returns a JSON object with keys `PublishedDate` and `LastModifiedDate`.

```
keyvalues vulnerability GHSA-xxx9-3xcr-gjj3
```

returns a JSON object without the aforementioned keys.

**Target state**

Both `keyvalues` commands return JSON objects with set `PublishedDate` and `LastModifiedDate` fields.

[1]: https://github.com/aquasecurity/trivy-db/blob/7924342bd057aa3385ca657159dae2482c738547/pkg/vulnsrc/vulnerability/vulnerability.go#L48-L49
[2]: https://github.com/spacewander/boltcli

knqyf263:
It sounds reasonable.

mhbardsley:
Thanks @DmitriyLewen and @knqyf263! I've attached what would be the fix for this just now in https://github.com/aquasecurity/trivy-db/pull/527. Let me know what you think 😄

PR #527, opened 2025-05-06
feat(db): populate vulnerability information with published & modified dates from GHSA (OSV) if NVD dates are not present

mhbardsley:
Addresses https://github.com/aquasecurity/trivy-db/issues/522

This diff achieves two things: firstly, it propagates published and modified date information from GHSAs and other vulnerability sources that are processed using the OSV processor, so they are available for normalization. Then, at the normalization stage, rather than only looking for NVD's published and modified date information (which may be missing), try NVD first and then other vulnerability sources in the order defined by `AllSourceIDs`.

Naturally, because NVD and OSV-based sources are the only ones we extract date information from, this will mean only those sources will succeed, but this generalises to cases where we may wish to extract this information from other sources in the future.

Two new test cases have been added to ensure NVD is preferred over (e.g.) GHSA, unless there is no NVD information available.

### Sizes
```
$ du -b old_db.tar.gz
66327988        old_db.tar.gz
$ du -b new_db.tar.gz 
66376260        new_db.tar.gz

```


mhbardsley:
Please let me know what you think about using `append([]types.SourceID{NVD}, AllSourceIDs...)` for date order of precedence here. I wanted to ensure that NVD takes precedence over anything else, even if severity precedence changes. But it might not be necessary.

pkg/vulnsrc/vulnerability/vulnerability.go
```diff
@@ -179,6 +180,34 @@ func getReferences(details map[types.SourceID]types.VulnerabilityDetail) []strin
 	return refs
 }
 
+func getPublishedDate(details map[types.SourceID]types.VulnerabilityDetail) *time.Time {
+	// Despite AllSourceIDs having an order of precedence, ensure we always use NVD first.
+	for _, source := range append([]types.SourceID{NVD}, AllSourceIDs...) {
```
chrisnovakovic, line 185:
```suggestion
	for _, source := range AllSourceIDs {
```

IMO, this should use whatever precedence is defined by `AllSourceIDs` (which is "NVD first" anyway).

pkg/vulnsrc/vulnerability/vulnerability.go
```diff
@@ -179,6 +180,34 @@ func getReferences(details map[types.SourceID]types.VulnerabilityDetail) []strin
...
+		if d.PublishedDate != nil {
+			return d.PublishedDate
+		}
+	}
+	return nil
+}
+
+func getLastModifiedDate(details map[types.SourceID]types.VulnerabilityDetail) *time.Time {
+	// Despite AllSourceIDs having an order of precedence, ensure we always use NVD first.
+	for _, source := range append([]types.SourceID{NVD}, AllSourceIDs...) {
```
chrisnovakovic, line 199:
```suggestion
	for _, source := range AllSourceIDs {
```

Same as above.

pkg/vulnsrc/vulnerability/vulnerability.go
```diff
@@ -179,6 +180,34 @@ func getReferences(details map[types.SourceID]types.VulnerabilityDetail) []strin
 	return refs
 }
 
+func getPublishedDate(details map[types.SourceID]types.VulnerabilityDetail) *time.Time {
+	// Assume dates use order of precedence from AllSourceIDs
+	for _, source := range AllSourceIDs {
```
DmitriyLewen, line 185: I'm still thinking about checking all sources here.

As I wrote [here](https://github.com/aquasecurity/trivy-db/issues/522#issuecomment-2851690890) - I'm worried about confusion.
There are languages ​​when we use 2 DBs (https://trivy.dev/latest/docs/scanner/vulnerability/#langpkg-data-sources).
So there may be a case when we take dates not from GHSA.
So I see 2 ways:
- as @mhbardsley said - adjust it when parsing the source (but as we can see, this PR already adds dates for Aqua and K8s). But it may not be intuitive.
- adjust sources here.

I prefer the 2nd way, because in this case we can control what dates we add to trivy-db.

knqyf263: I agree with @DmitriyLewen. The dates should be taken from sources as below:

- CVE-XXX-YYY: NVD (empty if NVD doesn't have the dates)
- GHSA-ZZZZ: GHSA

mhbardsley: Thanks - I think I understand the motivation here, but if you have a clear vision for this please feel free to modify my PR.

DmitriyLewen: @mhbardsley I updated PR - https://github.com/aquasecurity/trivy-db/pull/527/commits/d0c4064d40d63159377f974033b0a38c93765b2b

Can you take a look?

Also can you create the actual and new DB and compare the sizes?
also re-check that the new DB contains dates.

mhbardsley:
```
$ du -b old_db.tar.gz
66327988        old_db.tar.gz
$ du -b new_db.tar.gz 
66376260        new_db.tar.gz
```

So a difference of ~50KB.

I've checked and the new `trivy.db` now has the dates for GHSAs :smile:

DmitriyLewen (approved):
@mhbardsley thanks for your contribution!
LGTM.

cc. @knqyf263

pkg/vulnsrc/vulnerability/vulnerability.go
```diff
@@ -179,6 +180,36 @@ func getReferences(details map[types.SourceID]types.VulnerabilityDetail) []strin
 	return refs
 }
 
+func getPublishedDate(vulnID string, details map[types.SourceID]types.VulnerabilityDetail) *time.Time {
+	// We should take PublishedData from
+	//  - NVD - for CVE-IDs
+	//  - GHSA - for GHSA-IDs
+	// cf. https://github.com/aquasecurity/trivy-db/issues/522
+	switch {
+	case strings.HasPrefix(vulnID, "CVE"):
```
knqyf263, line 189: nit: The [crisis](https://therecord.media/cisa-extends-cve-program-contract-with-mitre) surrounding the CVE program has been a topic of discussion, and proposals such as [GCVE](https://gcve.eu/2025/04/16/gcve-allocation-announced/) have been made. Since new programs like CVEZ may also be proposed in the future, I think it's slightly safer to check all the way up to the hyphen.


```suggestion
	case strings.HasPrefix(vulnID, "CVE-"):
```

knqyf263 (approved):
LGTM.

pkg/vulnsrc/vulnerability/vulnerability.go
DmitriyLewen: It make sense. 
Updated in https://github.com/aquasecurity/trivy-db/pull/527/commits/7591542f48e29de1d6de81fdbc47023ab7f0ae9f

knqyf263 (approved):

PR #528, opened 2025-05-06
feat(echo): Add Echo Support

orizerah:
Adding support for Echo Images to Trivy

Vuln List Pull Request: [vuln-list-update](https://github.com/aquasecurity/vuln-list-update/pull/350)
Relevant Discussion: https://github.com/aquasecurity/trivy/discussions/8834


pkg/vulnsrc/echo/echo.go
```diff
@@ -0,0 +1,143 @@
...
+			return eb.With("file_path", filePath).Wrapf(err, "failed to open file")
+		}
+		defer f.Close()
+
+		pkgName := strings.TrimSuffix(entry.Name(), filepath.Ext(entry.Name()))
+		var advisory Advisory
+		if err := json.NewDecoder(f).Decode(&advisory); err != nil {
+			return eb.With("file_path", filePath).Wrapf(err, "json decode error")
+		}
+		advisoryMap[pkgName] = advisory
```
DmitriyLewen, line 79: You'll still check the recommendations in another cycle later.
Perhaps we can save package here?

pkg/vulnsrc/echo/echo.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+
+		if vulnInfo.Severity != "" {
+			severity, err := types.NewSeverity(strings.ToUpper(vulnInfo.Severity))
+			if err == nil {
+				adv.Severity = severity
+			}
+		}
+
+		ids := strings.Fields(cveID)
```
DmitriyLewen, line 120: can there be several CVEs here?

pkg/vulnsrc/echo/echo_test.go
DmitriyLewen, line 1: Can you also add test for `Get` fucntion?

pkg/vulnsrc/echo/echo.go
```diff
@@ -0,0 +1,143 @@
...
+type VulnInfo struct {
+	FixedVersion string `json:"fixed_version"`
+	Severity     string `json:"severity"`
+}
+
+type Advisory map[string]VulnInfo
+
+const (
+	echoDir    = "echo"
+	distroName = "echo"
```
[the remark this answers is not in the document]
orizerah: Done 👍

pkg/vulnsrc/echo/echo.go
orizerah: Nope, only one. Fixed.

pkg/vulnsrc/echo/echo_test.go
orizerah: Added 👍

pkg/vulnsrc/echo/echo.go
orizerah: I'm not sure I understand, you mean to save the package vulnerabilities separately for each package?

DmitriyLewen: NWM. I take a look on my idea again and i think your solution is good.

pkg/vulnsrc/echo/echo.go
```diff
@@ -0,0 +1,139 @@
...
+		if entry.IsDir() {
+			continue
+		}
+
+		filePath := filepath.Join(rootDir, entry.Name())
+		f, err := os.Open(filePath)
+		if err != nil {
+			return eb.With("file_path", filePath).Wrapf(err, "failed to open file")
+		}
+		defer f.Close()
```
DmitriyLewen, line 71: I suggest small refactoring to avoid using `defer` in loop:
```
diff --git a/pkg/vulnsrc/echo/echo.go b/pkg/vulnsrc/echo/echo.go
index c103ac3..ddb109b 100644
--- a/pkg/vulnsrc/echo/echo.go
+++ b/pkg/vulnsrc/echo/echo.go
@@ -63,17 +63,9 @@ func (vs VulnSrc) Update(dir string) error {
 			continue
 		}
 
-		filePath := filepath.Join(rootDir, entry.Name())
-		f, err := os.Open(filePath)
+		pkgName, advisory, err := vs.readAdvisory(rootDir, entry.Name())
 		if err != nil {
-			return eb.With("file_path", filePath).Wrapf(err, "failed to open file")
-		}
-		defer f.Close()
-
-		pkgName := strings.TrimSuffix(entry.Name(), filepath.Ext(entry.Name()))
-		var advisory Advisory
-		if err := json.NewDecoder(f).Decode(&advisory); err != nil {
-			return eb.With("file_path", filePath).Wrapf(err, "json decode error")
+			return eb.With("file_name", entry.Name()).Wrapf(err, "failed to read advisory")
 		}
 		advisoryMap[pkgName] = advisory
 	}
@@ -85,6 +77,23 @@ func (vs VulnSrc) Update(dir string) error {
 	return nil
 }
 
+func (vs VulnSrc) readAdvisory(rootDir, fileName string) (string, Advisory, error) {
+	filePath := filepath.Join(rootDir, fileName)
+	f, err := os.Open(filePath)
+	if err != nil {
+		return "", Advisory{}, oops.Wrapf(err, "failed to open file")
+	}
+	defer f.Close()
+
+	pkgName := strings.TrimSuffix(fileName, filepath.Ext(fileName))
+	var advisory Advisory
+	if err := json.NewDecoder(f).Decode(&advisory); err != nil {
+		return "", Advisory{}, oops.Wrapf(err, "json decode error")
+	}
+
+	return pkgName, advisory, nil
+}
+
 func (vs VulnSrc) save(advisoryMap map[string]Advisory) error {
 		if err := vs.dbc.PutDataSource(tx, distroName, source); err != nil {

```

orizerah: Done :)

orizerah:
@DmitriyLewen I went through all of the comments

pkg/vulnsrc/echo/echo_test.go
```diff
@@ -0,0 +1,161 @@
...
+				Fixtures:   tt.fixtures,
+				WantValues: tt.want,
+				PkgName:    tt.args.pkgName,
+				WantErr:    tt.wantErr,
+			})
+		})
+	}
+}
+
+func Test_readPackageAdvisory(t *testing.T) {
```
DmitriyLewen, line 110: I think this is too much :)

orizerah: Haha okay, I removed it.

DmitriyLewen:
and one more ask:
Can you run [Trivy DB](https://github.com/aquasecurity/trivy-db/blob/main/.github/workflows/cron.yml) workflow in your fork and create container with new trivy-db (e.g. in ghcr.io)?

To show that new trivy-db contains `echo` advisories

orizerah:
@DmitriyLewen I see now that the vuln-list-update pr was reverted, so vuln-list doesn't contain the Echo advisory

DmitriyLewen:
workflow should take `vuln-list` of repo owner (so from your fork)

orizerah:
@DmitriyLewen It's running.
https://github.com/orizerah/trivy-db/actions/runs/15162449074/job/42631579054

Should I re-open the vuln-list-update pr?

DmitriyLewen:
> It's running.

You may need to add credentials for DockerHub and ECR or remove the steps to login and download them.

> Should I re-open the vuln-list-update pr?

Don't need it now, we will write about it later

orizerah:
I ran the workflow
https://github.com/orizerah/trivy-db/actions/runs/15163318963/job/42634524060
You can pull it from
oras pull ghcr.io/orizerah/trivy-db:latest

DmitriyLewen (approved):
LGTM
@orizerah  Thanks for your contribution!

cc. @knqyf263

knqyf263 (approved):

DmitriyLewen:
@orizerah I found one case:
изображение

Does it mean that all version are affected?

orizerah:
@DmitriyLewen Yes, it means it's fixed in all versions

DmitriyLewen:
hmm.. this is a bit weird, but okay.
let's use this. We can change it later (if needed).

knqyf263:
> @DmitriyLewen Umm no, it means it's fixed in all versions

In that case, we don't need to insert the record, right? This advisory will not detect any vulnerabilities.

orizerah:
We can remove these rows from our published advisory instead. That way there is no need for special use cases in trivy-db.

knqyf263:
> We can remove these rows from our published advisory instead. That way there is no need for special use cases in trivy-db.

That would be great.

orizerah:
@knqyf263 Done

knqyf263:
@DmitriyLewen Would you check if this PR works once vuln-list is updated?

DmitriyLewen:
I just ran `vuln-list` update to check this - https://github.com/aquasecurity/vuln-list-update/actions/runs/15317997089/job/43095416156

DmitriyLewen:
I can confirm that trivy-db build from new `vuln-list` contains echo advisories:
изображение

PR #540, opened 2025-06-04
feat(ubuntu): add support 20.04-ESM

DmitriyLewen:
## Description
Ubutnu now supports 20.04 only for Ubuntu pro (ESM).
So we need to add advisories for 20.04-ESM.

New name for 20.04-ESM is `esm-apps/focal`:
`(release_list="trusty trusty/esm xenial esm-infra/xenial esm-apps/xenial bionic esm-infra/bionic esm-apps/bionic focal esm-apps/focal jammy esm-apps/jammy noble oracular plucky"`
took from https://wiki.ubuntu.com/SecurityTeam/BuildEnvironment#line867

Same for files from `vuln-list`:
```
➜ grep -rh "focal" | sort -u | grep esm 
      "esm-apps/focal": {
```

## Test build
I confirm that trivy-db contains new advisories.
e.g.:
изображение

### DIFF
--- a/pkg/vulnsrc/ubuntu/ubuntu.go
+++ b/pkg/vulnsrc/ubuntu/ubuntu.go
@@ -52,10 +52,15 @@ var (
 		"noble":    "24.04",
 		"oracular": "24.10",
 		// ESM versions:
-		"precise/esm":      "12.04-ESM",
-		"trusty/esm":       "14.04-ESM",
+		"precise/esm": "12.04-ESM",
+		"trusty/esm":  "14.04-ESM",
+		// Possible multiple values for one release:
+		//(release_list="trusty trusty/esm xenial esm-infra/xenial esm-apps/xenial bionic esm-infra/bionic esm-apps/bionic focal esm-apps/focal jammy esm-apps/jammy noble oracular plucky")
+		// cf. https://wiki.ubuntu.com/SecurityTeam/BuildEnvironment#line867
 		"esm-infra/xenial": "16.04-ESM",
+		"esm-apps/xenial":  "16.04-ESM",
 		"esm-infra/bionic": "18.04-ESM",
+		"esm-apps/bionic":  "18.04-ESM",
 		"esm-apps/focal":   "20.04-ESM",
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/vulnsrc/ubuntu/ubuntu.go
```diff
@@ -52,10 +52,16 @@ var (
...
+		"precise/esm": "12.04-ESM",
+		"trusty/esm":  "14.04-ESM",
+		// Possible multiple values for one release:
+		// (release_list="trusty trusty/esm xenial esm-infra/xenial esm-apps/xenial bionic esm-infra/bionic esm-apps/bionic focal esm-apps/focal jammy esm-apps/jammy noble oracular plucky")
+		// cf. https://wiki.ubuntu.com/SecurityTeam/BuildEnvironment#line867
 		"esm-infra/xenial": "16.04-ESM",
+		"esm-apps/xenial":  "16.04-ESM",
 		"esm-infra/bionic": "18.04-ESM",
+		"esm-apps/bionic":  "18.04-ESM",
+		"esm-apps/focal":   "20.04-ESM",
```
knqyf263, line 65: If I understand correctly, ESM has two streams, apps and infra.
https://ubuntu.com/security/esm#:~:text='esm%2Dapps'%20and%20',are%20compliant%20with%20FIPS%20regulations.

So, should we also add `esm-infra/focal`?

DmitriyLewen: I didn't find using of `esm-infra/focal`.
But Ubuntu just started supporting `20.04-ESM`, so there is a chance they will add it.
so I don't see any reason not to add it

Added in https://github.com/aquasecurity/trivy-db/pull/540/commits/a2315e2382d83229deab83dce0fcbd784501e574

DmitriyLewen: I was just thinking that maybe we'll have to split these sources (like fips, ksplice for Oracle Linux).
But in any case, we need to wait for user requests first.

knqyf263: >But Ubuntu just started supporting 20.04-ESM, so there is a chance they will add it.

Yes, that's also my guess. The advisory has not yet been issued for the packages within the esm-infra repository, but I believe it will be published soon.

knqyf263: >I was just thinking that maybe we'll have to split these sources (like fips, ksplice for Oracle Linux).

I also found that Ubuntu has [fips](https://ubuntu.com/security/fips), like Oracle. We probably need a special handling for them.

PR #542, opened 2025-06-17
bugfix: added new checks to skip NVD Rejected CVE

gshanbhag525:
There are some CVE which will not have "** REJECT **" keyword, but they are rejected.

Hence added checks to skip such CVEs.
eg.
CVE-2024-4109
CVE-2022-45046


pkg/vulnsrc/vulnerability/vulnerability.go
```diff
@@ -216,7 +218,10 @@ func getRejectedStatus(details map[types.SourceID]types.VulnerabilityDetail) boo
 		if !ok {
 			continue
 		}
-		if strings.Contains(d.Description, rejectVulnerability) {
+		if strings.Contains(d.Description, rejectVulnerability) || strings.Contains(d.Description, rejectedDoNotUse) {
+			return true
+		}
+		if strings.HasPrefix(d.Description, rejectedReason) {
```
knqyf263: I think we can merge this check into the above if-statement.

pkg/vulnsrc/vulnerability/vulnerability.go
```diff
@@ -216,7 +218,10 @@ func getRejectedStatus(details map[types.SourceID]types.VulnerabilityDetail) boo
 		if !ok {
 			continue
 		}
-		if strings.Contains(d.Description, rejectVulnerability) {
+		if strings.Contains(d.Description, rejectVulnerability) || strings.Contains(d.Description, rejectedDoNotUse) {
```
knqyf263: Ideally, we should take the status from `vulnStatus`.

```
jq .vulnStatus api/2024/CVE-2024-4109.json
"Rejected"
```

```suggestion
		// TODO(NVD): Take the vulnerability status from `vulnStatus`
		if strings.Contains(d.Description, rejectVulnerability) || strings.Contains(d.Description, rejectedDoNotUse) {
```

gshanbhag525: But its present only for NVD source.
I believe, For other sources we will still use description. correct me if I am wrong here.

gshanbhag525: hi @knqyf263 
I have updated PR to check rejected VulnStatus for NVD source 
Also further checking for other sources if we have reject keywords in description.
Because
I found some for amazon advisory. [here](https://github.com/aquasecurity/vuln-list/blob/main/amazon/2/ALAS2-2023-2088.json)

my fork trivy build with logs to print rejected vuln from other sources: [here](https://github.com/gshanbhag525/trivy-db/actions/runs/15709004652/job/44262431981)

can you take a quick look.

pkg/vulnsrc/vulnerability/vulnerability.go
```diff
@@ -211,12 +214,15 @@ func getLastModifiedDate(vulnID string, details map[types.SourceID]types.Vulnera
...
+	if nvd, ok := details[NVD]; ok && nvd.VulnStatus == rejectedStatus {
+		return true
+	}
 	for _, source := range AllSourceIDs {
 		d, ok := details[source]
 		if !ok {
 			continue
 		}
-		if strings.Contains(d.Description, rejectVulnerability) {
+		if strings.Contains(d.Description, rejectVulnerability) || strings.Contains(d.Description, rejectedDoNotUse) || strings.HasPrefix(d.Description, rejectedReason) {
```
knqyf263: I don't think we need to limit it to NVD as we may take the status from other sources in the future.

```suggestion
		if d.VulnStatus == rejectedStatus {
			continue
		}
		if strings.Contains(d.Description, rejectVulnerability) || strings.Contains(d.Description, rejectedDoNotUse) || strings.HasPrefix(d.Description, rejectedReason) {
```

pkg/vulnsrc/nvd/nvd.go
```diff
@@ -198,6 +198,7 @@ func (nvd *NVD) Put(tx *bolt.Tx, cve Cve) error {
 		Description:      description,
 		PublishedDate:    &publishedDate,
 		LastModifiedDate: &lastModifiedDate,
+		VulnStatus:       cve.VulnStatus,
```
knqyf263: I'm afraid their status is not consistent, like `REJECTED` or `rejected`. I think we should get it aligned on our end.

```suggestion
		VulnStatus:       strings.ToUpper(cve.VulnStatus),
```

pkg/vulnsrc/vulnerability/vulnerability.go
```diff
@@ -11,7 +11,10 @@ import (
 )
 
 const (
+	rejectedStatus      = "Rejected"
```
knqyf263: Accordingly

```suggestion
	rejectedStatus      = "REJECTED"
```

pkg/types/types.go
```diff
@@ -83,6 +83,7 @@ type VulnerabilityDetail struct {
 	Description      string     `json:",omitempty"`
 	PublishedDate    *time.Time `json:",omitempty"` // Take from NVD or GHSA
 	LastModifiedDate *time.Time `json:",omitempty"` // Take from NVD or GHSA
+	VulnStatus       string     `json:",omitempty"`
```
knqyf263: I would change it to `Status`.

```suggestion
	Status       string     `json:",omitempty"` // Rejected or not
```

knqyf263 (commented):
@DmitriyLewen Any comments?

pkg/vulnsrc/vulnerability/vulnerability.go
```diff
@@ -216,7 +219,10 @@ func getRejectedStatus(details map[types.SourceID]types.VulnerabilityDetail) boo
 		if !ok {
 			continue
 		}
-		if strings.Contains(d.Description, rejectVulnerability) {
+		if d.Status == rejectedStatus {
+			continue
+		}
+		if strings.Contains(d.Description, rejectKeyword) || strings.Contains(d.Description, rejectedDoNotUse) || strings.HasPrefix(d.Description, rejectedReason) {
 			return true
```
DmitriyLewen: I am a bit confusing.
Why we use different logic for status(check next source) and description (return `true`)?

take a look this testcase:
```
--- a/pkg/vulnsrc/vulnerability/vulnerability_test.go
+++ b/pkg/vulnsrc/vulnerability/vulnerability_test.go
@@ -124,6 +124,34 @@ func TestIsRejected(t *testing.T) {
 			},
 			want: true,
 		},
+		{
+			name: "happy path, when vulnerability from redhat and ubuntu is rejected by Nvd (from status only)",
+			details: map[types.SourceID]types.VulnerabilityDetail{
+				vulnerability.RedHat: {
+					ID:          "CVE-2020-1234",
+					CvssScoreV3: 5.6,
+					Title:       "test vulnerability",
+					Description: "a test vulnerability where vendor rates it lower than NVD",
+				},
+				vulnerability.Ubuntu: {
+					ID:          "CVE-2020-1234",
+					CvssScore:   1.2,
+					CvssScoreV3: 3.4,
+					Severity:    types.SeverityLow,
+					SeverityV3:  types.SeverityMedium,
+					Title:       "test vulnerability",
+					Description: "a test vulnerability where vendor rates it lower than NVD",
+				},
+				vulnerability.NVD: {
+					ID:          "CVE-2020-1234",
+					CvssScore:   9.1,
+					Title:       "test vulnerability",
+					Description: "a test vulnerability where vendor rates it lower than NVD",
+					Status:      "Rejected",
+				},
+			},
+			want: true,
+		},
 	}
 
 	for _, tc := range testCases {

```
I think we should get `true`, but see `false`.

pkg/types/types.go
```diff
@@ -83,6 +83,7 @@ type VulnerabilityDetail struct {
 	Description      string     `json:",omitempty"`
 	PublishedDate    *time.Time `json:",omitempty"` // Take from NVD or GHSA
 	LastModifiedDate *time.Time `json:",omitempty"` // Take from NVD or GHSA
+	Status           string     `json:",omitempty"` // Rejected or not
```
DmitriyLewen: We don't keep `Status` into `trivy-db` (IIUC this is correct behavior).

But It happens because `Normalize` function doesn't add `Status`:
https://github.com/aquasecurity/trivy-db/blob/fc85ecb02597bc72bb6c5ce28b6b53dcfa764450/pkg/vulnsrc/vulnerability/vulnerability.go#L44-L53

But i think we need to be clear about this. I see 2 ways:
1. Use ``json:"-"`` tag.
2. Add info in comment.

knqyf263: Ah, yes, good point. We should not store that information in the database.

pkg/vulnsrc/vulnerability/vulnerability.go
gshanbhag525: > I think we should get true, but see false.

The exisitng test case passed because even though Status is not capital case "REJECTED",
 it found rejection keywords in description .
 I need to update the status to REJECTED in test case. because NVD source will put status as capital case.
 For other source we have not yet added Status
 
 Now coming back to your test.
 Since Status is not capital case and no rejection keywords in desc it fails.

DmitriyLewen: let me explain my thought:
possible 4 cases:
| status is rejected             | description contains one of rejected markers | IsRejected now| IsRejects as i thought |
| :---------------- | :------: | :----: | :---- |
| -        |   -   | - | -
| -           |   +   | + | +
| +    |  -   | - | +
| + |  +   | + | +

I said about 3 case (when status is Rejected, my description doesn't have info about that).
I think that we should mark this vulnerability as Rejected.
wdyt?

gshanbhag525: yes, sure

pkg/types/types.go
gshanbhag525: done

pkg/vulnsrc/vulnerability/vulnerability.go
gshanbhag525: done

knqyf263 (approved):

PR #543, opened 2025-06-18
ci: use `AdityaGarg8/remove-unwanted-software` instead of `easimon/maximize-build-space`

DmitriyLewen:
## Description
We got space error - https://github.com/aquasecurity/trivy-db/actions/runs/15714756455/job/44281599307
`vuln-list` and language repos use 32GB space:
```
➜ du -shm ./trivy-db-cache 
31365 ./trivy-db-cache
```
`easimon/maximize-build-space` can't free more then 40GB space - https://github.com/DmitriyLewen/trivy-db/actions/runs/15725438939/job/44314263726#step:2:177
But `AdityaGarg8/remove-unwanted-software` frees 54 GB - https://github.com/DmitriyLewen/trivy-db/actions/runs/15731622643/job/44333867214#step:2:157

So we want to use `dityaGarg8/remove-unwanted-software@v5` to get more free space.

Test run - https://github.com/DmitriyLewen/trivy-db/actions/runs/15731622643/job/44333867214

### DIFF
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -15,7 +15,7 @@ jobs:
     steps:
       # vuln-list dirs + language repositories use more than 31GB of storage
       - name: Maximize build space
-        uses: AdityaGarg8/remove-unwanted-software@v5
+        uses: AdityaGarg8/remove-unwanted-software@90e01b21170618765a73370fcc3abbd1684a7793
         with:
           remove-android: 'true'
           remove-dotnet: 'true'


.github/workflows/cron.yml
```diff
@@ -13,14 +13,18 @@ jobs:
     name: Build DB
     runs-on: ubuntu-24.04
     steps:
+      # vuln-list dirs + language repositories use more than 31GB of storage
       - name: Maximize build space
-        uses: easimon/maximize-build-space@v10
+        uses: AdityaGarg8/remove-unwanted-software@v5
```
knqyf263: We should use the commit hash for security reasons.

DmitriyLewen: Right! Thanks!
Fixed in https://github.com/aquasecurity/trivy-db/pull/543/commits/43eb3a650e21c96d2ab3170901c86bfb20185394

knqyf263 (approved):

.github/workflows/cron.yml
knqyf263: NOTE: I would like to avoid relying on third-party actions (especially those that are not widely used) as much as possible, so I plan to remove them in the future.

simar7: @knqyf263 since we use this action quite a bit, should be fork it and maintain it under aquasec? That way we could eliminate security issues such as the ones you mention.

Of course if we find another way to not use such actions (which are kind of a hack anyway), it'll be the best.

DmitriyLewen: > should be fork it and maintain it under aquasec?

I also thought about that.
But i were worried that we need to understand how it works and monitor changes in runners.

knqyf263: Since they are simple shell scripts, we can write a similar one from scratch.

PR #546, opened 2025-06-25
feat: Add Root.io vulnerability source support

chait-slim:
Summary
  - Adds a new vulnerability source for Root.io security patches
  - Supports Debian, Ubuntu, and Alpine-based container images
  - Implements constraint-format version ranges for accurate package matching

  Implementation Details
  Key Features:
  - Multi-OS Support: Single VulnSrc implementation handles Debian, Ubuntu, Alpine variants
  - Constraint Versioning: Uses version ranges (e.g., ">=1.2.3, <2.0.0") instead of fixed versions
  - Bucket Naming: Platform-specific buckets like "root.io debian 11", "root.io ubuntu 20.04"
  - Feed Processing: Reads from rootio/ directory in vuln-list repository
  - Error Handling: Uses samber/oops for structured error handling with context
  - Testing: Includes happy/sad path tests with fixtures for all supported OS types


pkg/vulnsrc/rootio/rootio.go
```diff
@@ -0,0 +1,211 @@
...
+const (
+	rootioDir      = "rootio"
+	platformFormat = "root.io %s %s" // "root.io {baseOS} {version}"
+)
+
+var (
+	source = types.DataSource{
+		ID:   vulnerability.RootIO,
+		Name: "Root.io Security Patches",
+		URL:  "https://root.io/",
```
DmitriyLewen: I think it would be more correct to use URL for `feed`
```suggestion
		URL:  "https://api.root.io/external/patch_feed",
```

pkg/vulnsrc/rootio/rootio.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+)
+
+type Option func(src *VulnSrc)
```
DmitriyLewen: It looks like you don't overwrite VulnSrc fields.
So you can remove this.
Or you need to add `With*` functions

pkg/vulnsrc/rootio/rootio.go
```diff
@@ -0,0 +1,211 @@
...
+
+type VulnSrc struct {
+	baseOS OSType
+	put    db.CustomPut
```
DmitriyLewen: I don't think we need it.
If we will need this - we will update it in another PR.

pkg/vulnsrc/rootio/testdata/happy/vuln-list/rootio/alpine-320.json
knqyf263, line 1: Is this the same format as the one used in the vuln-list? https://api.root.io/external/patch_feed has a distribution key.

pkg/vulnsrc/rootio/rootio.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+	return nil
+}
+
+func (vs VulnSrc) Get(osVer, pkgName string) ([]types.Advisory, error) {
+	eb := oops.In("rootio").With("base_os", vs.baseOS).With("os_version", osVer).With("package_name", pkgName)
+
+	// Generate bucket name: "root.io {baseOS} {osVer}"
+	bucket := fmt.Sprintf(platformFormat, string(vs.baseOS), osVer)
+	advisories, err := vs.dbc.GetAdvisories(bucket, pkgName)
```
knqyf263: @chait-slim I think we also need to get advisories from the distributor's bucket and merge them with Root.io advisories as below.
https://github.com/aquasecurity/trivy/blob/1885e35fa6d9d2fce66556af258bbd6048a226d1/pkg/detector/ospkg/rootio/vulnsrc.go#L42-L73

Otherwise, Trivy detects only vulnerabilities for which Root.io distributes a patch. Am I missing something?

pkg/vulnsrc/rootio/rootio.go
DmitriyLewen, line 1: It looks like we can do that simpler.
Can you recheck?
```diff
diff --git a/pkg/vulnsrc/rootio/rootio.go b/pkg/vulnsrc/rootio/rootio.go
index 74f08a3..da2ff37 100644
--- a/pkg/vulnsrc/rootio/rootio.go
+++ b/pkg/vulnsrc/rootio/rootio.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"io"
 	"path/filepath"
-	"strings"
 
 	"github.com/samber/oops"
 	bolt "go.etcd.io/bbolt"
@@ -42,7 +41,6 @@ type VulnSrc struct {
 func NewVulnSrc(baseOS OSType, opts ...Option) VulnSrc {
 	src := VulnSrc{
 		baseOS: baseOS,
-		put:    defaultPut,
 		dbc:    db.Config{},
 		logger: log.WithPrefix(fmt.Sprintf("rootio-%s", baseOS)),
 	}
@@ -62,50 +60,21 @@ func (vs VulnSrc) Update(dir string) error {
 	rootDir := filepath.Join(dir, "vuln-list", rootioDir)
 	eb := oops.In("rootio").With("root_dir", rootDir).With("base_os", vs.baseOS)
 
-	var feeds []RootIOFeed
 	err := utils.FileWalk(rootDir, func(r io.Reader, path string) error {
-		var rawFeed RawRootIOFeed
-		if err := json.NewDecoder(r).Decode(&rawFeed); err != nil {
+		var cveFeed CVEFeed
+		if err := json.NewDecoder(r).Decode(&cveFeed); err != nil {
 			return eb.With("file_path", path).Wrapf(err, "json decode error")
 		}
 
-		// Extract data for our specific base OS and convert to internal format
-		var rawDistroData []RawDistroData
-		switch vs.baseOS {
-		case Alpine:
-			rawDistroData = rawFeed.Alpine
-		case Debian:
-			rawDistroData = rawFeed.Debian
-		case Ubuntu:
-			rawDistroData = rawFeed.Ubuntu
-		}
-
-		// Convert each distro version to our internal RootIOFeed format
-		for _, distro := range rawDistroData {
-			feed := RootIOFeed{
-				BaseOS:  string(vs.baseOS),
-				Version: distro.DistroVersion,
-				Patches: make(map[string][]Patch),
-			}
-
-			// Convert packages to patches
-			for _, pkg := range distro.Packages {
-				for cveID, cveInfo := range pkg.Pkg.CVEs {
-					patch := Patch{
-						VulnerabilityID:    cveID,
-						VulnerableVersions: cveInfo.VulnerableRanges,
-					}
-					// Use first fixed version if available
-					if len(cveInfo.FixedVersions) > 0 {
-						patch.FixedVersion = cveInfo.FixedVersions[0]
+		for os, distroData := range cveFeed {
+			for _, data := range distroData {
+				platform := fmt.Sprintf(platformFormat, os, data.DistroVersion)
+				for _, pkg := range data.Packages {
+					if err := vs.save(platform, pkg.Pkg); err != nil {
+						return err
 					}
-					feed.Patches[pkg.Pkg.Name] = append(feed.Patches[pkg.Pkg.Name], patch)
 				}
 			}
-
-			if len(feed.Patches) > 0 {
-				feeds = append(feeds, feed)
-			}
 		}
 
 		return nil
@@ -114,18 +83,20 @@ func (vs VulnSrc) Update(dir string) error {
 		return eb.Wrapf(err, "walk error")
 	}
 
-	if err = vs.save(feeds); err != nil {
-		return eb.Wrapf(err, "save error")
-	}
-
 	return nil
 }
 
-func (vs VulnSrc) save(feeds []RootIOFeed) error {
+func (vs VulnSrc) save(platform string, pkg PackageInfo) error {
 	vs.logger.Info("Saving Root.io DB", "base_os", vs.baseOS)
+
+	eb := oops.With("platform", platform)
 	err := vs.dbc.BatchUpdate(func(tx *bolt.Tx) error {
-		for _, feed := range feeds {
-			if err := vs.commit(tx, feed); err != nil {
+		if err := vs.dbc.PutDataSource(tx, platform, source); err != nil {
+			return eb.Wrapf(err, "failed to put data source")
+		}
+
+		for cve, cveInfo := range pkg.CVEs {
+			if err := vs.commit(tx, platform, pkg.Name, cve, cveInfo); err != nil {
 				return err
 			}
 		}
@@ -137,28 +108,28 @@ func (vs VulnSrc) save(feeds []RootIOFeed) error {
 	return nil
 }
 
-func (vs VulnSrc) commit(tx *bolt.Tx, feed RootIOFeed) error {
-	eb := oops.With("base_os", feed.BaseOS).With("version", feed.Version)
+func (vs VulnSrc) commit(tx *bolt.Tx, platform, pkgName, cveID string, cveInfo CVEInfo) error {
+	eb := oops.With("platform", platform).With("package", pkgName).With("cveID", cveID)
 
-	// Create platform bucket name: "root.io {baseOS} {version}"
-	platformName := fmt.Sprintf(platformFormat, strings.ToLower(feed.BaseOS), feed.Version)
+	// Create advisory with constraint format in VulnerableVersions
+	adv := types.Advisory{
+		VulnerableVersions: cveInfo.FixedVersions, // Store constraint format here
+	}
 
-	if err := vs.dbc.PutDataSource(tx, platformName, source); err != nil {
-		return eb.Wrapf(err, "failed to put data source")
+	// Set fixed version if available
+	if len(cveInfo.FixedVersions) > 0 {
+		adv.FixedVersion = cveInfo.FixedVersions[0]
 	}
 
-	for pkgName, patches := range feed.Patches {
-		for _, patch := range patches {
-			patchData := PatchData{
-				PlatformName: platformName,
-				PackageName:  pkgName,
-				Patch:        patch,
-			}
-			if err := vs.put(vs.dbc, tx, patchData); err != nil {
-				return eb.With("package", pkgName).With("cve", patch.VulnerabilityID).Wrapf(err, "put error")
-			}
-		}
+	if err := vs.dbc.PutAdvisoryDetail(tx, cveID, pkgName, []string{platform}, adv); err != nil {
+		return eb.Wrapf(err, "failed to save advisory")
+	}
+
+	// For optimization
+	if err := vs.dbc.PutVulnerabilityID(tx, cveID); err != nil {
+		return eb.Wrapf(err, "failed to save the vulnerability ID")
 	}
+
 	return nil
 }
 
@@ -173,39 +144,3 @@ func (vs VulnSrc) Get(osVer, pkgName string) ([]types.Advisory, error) {
 	}
 	return advisories, nil
 }
-
-type PatchData struct {
-	PlatformName string
-	PackageName  string
-	Patch        Patch
-}
-
-func defaultPut(dbc db.Operation, tx *bolt.Tx, advisory any) error {
-	patchData, ok := advisory.(PatchData)
-	if !ok {
-		return oops.Errorf("unknown type")
-	}
-
-	eb := oops.With("platform", patchData.PlatformName).With("package", patchData.PackageName).With("cve", patchData.Patch.VulnerabilityID)
-
-	// Create advisory with constraint format in VulnerableVersions
-	adv := types.Advisory{
-		VulnerableVersions: patchData.Patch.VulnerableVersions, // Store constraint format here
-	}
-
-	// Set fixed version if available
-	if patchData.Patch.FixedVersion != "" {
-		adv.FixedVersion = patchData.Patch.FixedVersion
-	}
-
-	if err := dbc.PutAdvisoryDetail(tx, patchData.Patch.VulnerabilityID, patchData.PackageName, []string{patchData.PlatformName}, adv); err != nil {
-		return eb.Wrapf(err, "failed to save advisory")
-	}
-
-	// For optimization
-	if err := dbc.PutVulnerabilityID(tx, patchData.Patch.VulnerabilityID); err != nil {
-		return eb.Wrapf(err, "failed to save the vulnerability ID")
-	}
-
-	return nil
-}
diff --git a/pkg/vulnsrc/rootio/types.go b/pkg/vulnsrc/rootio/types.go
index f460c01..54da21f 100644
--- a/pkg/vulnsrc/rootio/types.go
+++ b/pkg/vulnsrc/rootio/types.go
@@ -1,50 +1,31 @@
 package rootio
 
-// RootIOFeed represents the Root.io patch feed structure (internal format)
-type RootIOFeed struct {
-	BaseOS  string             `json:"base_os"` // debian, ubuntu, alpine
-	Version string             `json:"version"` // OS version (e.g., "11", "20.04", "3.20")
-	Patches map[string][]Patch `json:"patches"` // package name -> patches
-}
-
-// RawRootIOFeed represents the actual Root.io API feed format
-type RawRootIOFeed struct {
-	Alpine []RawDistroData `json:"alpine,omitempty"`
-	Debian []RawDistroData `json:"debian,omitempty"`
-	Ubuntu []RawDistroData `json:"ubuntu,omitempty"`
-}
+// CVEFeed represents the root structure of the CVE feed API response
+type CVEFeed map[string][]DistroData
 
-// RawDistroData represents distribution data from the API
-type RawDistroData struct {
-	DistroVersion string           `json:"distroversion"`
-	Packages      []RawPackageData `json:"packages"`
+// DistroData represents vulnerability data for a specific distribution version
+type DistroData struct {
+	DistroVersion string        `json:"distroversion"`
+	Packages      []PackageData `json:"packages"`
 }
 
-// RawPackageData represents package data from the API
-type RawPackageData struct {
-	Pkg RawPackageInfo `json:"pkg"`
+// PackageData represents a package with its vulnerability information
+type PackageData struct {
+	Pkg PackageInfo `json:"pkg"`
 }
 
-// RawPackageInfo represents package info from the API
-type RawPackageInfo struct {
-	Name string                `json:"name"`
-	CVEs map[string]RawCVEInfo `json:"cves"`
+// PackageInfo contains the package name and CVE information
+type PackageInfo struct {
+	Name string             `json:"name"`
+	CVEs map[string]CVEInfo `json:"cves"`
 }
 
-// RawCVEInfo represents CVE info from the API
-type RawCVEInfo struct {
+// CVEInfo contains vulnerability details for a specific CVE
+type CVEInfo struct {
 	VulnerableRanges []string `json:"vulnerable_ranges"`
 	FixedVersions    []string `json:"fixed_versions"`
 }
 
-// Patch represents a single patch entry from Root.io
-type Patch struct {
-	VulnerabilityID string `json:"vulnerability_id"` // CVE-ID
-	// VulnerableVersions contains constraint format (e.g., ">=1.2.3, <2.0.0")
-	VulnerableVersions []string `json:"vulnerable_versions"`
-	FixedVersion       string   `json:"fixed_version,omitempty"`
-}
-
 // OSType represents supported base operating systems
 type OSType string
 

```

pkg/vulnsrc/rootio/rootio.go
```diff
@@ -0,0 +1,211 @@
...
+func (vs VulnSrc) Name() types.SourceID {
+	return source.ID
+}
+
+func (vs VulnSrc) Update(dir string) error {
+	rootDir := filepath.Join(dir, "vuln-list", rootioDir)
+	eb := oops.In("rootio").With("root_dir", rootDir).With("base_os", vs.baseOS)
+
+	var feeds []RootIOFeed
+	err := utils.FileWalk(rootDir, func(r io.Reader, path string) error {
```
DmitriyLewen: IIUC it should be single file:
https://github.com/aquasecurity/vuln-list-update/blob/d3bc87388df9cbce20081af06ba382093adf173d/rootio/rootio.go#L72-L76
Do we need run `utils.FileWalk`?
And `testdata` contains separated files

pkg/vulnsrc/rootio/rootio.go
knqyf263: Fixed in https://github.com/aquasecurity/trivy-db/pull/546/commits/c3715854bcd8b58ec98d82e7d6a0ab272a50375d

pkg/vulnsrc/rootio/rootio.go
knqyf263: Removed https://github.com/aquasecurity/trivy-db/pull/546/commits/c0a9688e386b48b4771599f45eb12f7a56c42150

pkg/vulnsrc/rootio/rootio.go
knqyf263: Removed https://github.com/aquasecurity/trivy-db/pull/546/commits/5851f50bd5a8fab02325c2aa617dea58c14e4c30

pkg/vulnsrc/rootio/rootio.go
knqyf263: Refactored in https://github.com/aquasecurity/trivy-db/pull/546/commits/5851f50bd5a8fab02325c2aa617dea58c14e4c30

pkg/vulnsrc/rootio/rootio.go
knqyf263: I'm also wondering if the test data format is correct.
https://github.com/aquasecurity/trivy-db/pull/546#discussion_r2168326814

knqyf263: @chait-slim I've updated the test data, but correct me if I'm wrong.
https://github.com/aquasecurity/trivy-db/pull/546/commits/304e759db9e1fe7791d3a5bd0571382dcb47fe39#diff-9e8538cadd0a1396b248749743bb093061e0ab915d0270a364c37e7d2a6c9322

pkg/vulnsrc/rootio/testdata/happy/vuln-list/rootio/alpine-320.json
chait-slim: Youre right. I've made some changes and didnt retest it. The correct format is what the API presents, not what the test has

pkg/vulnsrc/rootio/rootio.go
chait-slim: If I understand you correctly the result here should be:
1. Find the distros advisory <- This will always exist since we are using distros already created in the db. But it means Root needs to run last on update
2. Update the advisory with the Root data 



Does that make sense?

DmitriyLewen: There are following cases:
```
- Root.io feed includes
   - Case 1: Only Root.io has a patch
   - Case 2: Both Root.io and Debian have a patch
- Debian feed includes
   - Case 3: Debian doesn't have a patch, but affected
      - 3-1: But Root.io has a patch (same as case 1)
      - 3-2 Root.io also doesn't have a patch
   - Case 4: Debian has a patch
      - 4-1: Root.io also has a patch (same as case 2) 
      - 4-2: Root.io doesn't have a patch
```
We think that `Root.io` feed will not include 3-2 and 4-2 cases.
That is why we assume:
take distros and Root.io advisories and merge them (Root.io has high priority).

in other words, for cases when Root.io doesn't have advisory for vulnID - Trivy will check this vulnID from the vendor bucket and show it (if it exists).
Do you agree with this behavior?
or do you plan to enrich the feed with advisories for which you do not have fixes yet?

benjifin: Understood the question and think it's the correct behaviour as you suggest- we don't enrich the feed with additional CVEs we don't have fixes for.
However we do at times enrich the feed in a use case not listed here explicitly: places where the distro has not have a fix and does not report the CVE at all, and root does have a fix and will report both the CVE + FIX (Alpine for example does not include it's unfixed CVE in it's feed, some other distros have a similar pattern)

DmitriyLewen: IIUC you say about `case 1`. 
Debian doesn't have patch, Root.io has patch.
In this case after merge advisories Trivy will show vulnerability from Root.io

pkg/vulnsrc/rootio/rootio.go
chait-slim: You are correct @knqyf263 . Sorry for this misalignment

pkg/vulnsrc/rootio/rootio.go
chait-slim: @DmitriyLewen  💯 what youve described

DmitriyLewen: I updated `Get` function - https://github.com/aquasecurity/trivy-db/pull/546/commits/618ed06dd0e5990a4a122df0dd9b882ac04115d9
and added new tests.
@benjifin @chait-slim tell me if i missed something.

chait-slim: @DmitriyLewen Just to validate, the comment here:
`// Merge the advisories from the original distributors with Root.io's advisories`
Its not really merging but more: override distro-only fix if root fix exists. Correct?

pkg/vulnsrc/rootio/testdata/happy/vuln-list/rootio/test-feed.json
```diff
@@ -0,0 +1,67 @@
...
+  "debian": [
+    {
+      "distroversion": "11",
+      "packages": [
+        {
+          "pkg": {
+            "name": "openssl",
+            "cves": {
+              "CVE-2023-0464": {
+                "vulnerable_ranges": [">=1.1.1, <1.1.1t"],
```
chait-slim: @DmitriyLewen The tests here are missing a Root fix. This is not an issue per-se, but I'm concerned they dont test a real use case. For example here, add a fixed version: 1.1.1t-1+deb11u1.root.io.1

pkg/vulnsrc/rootio/rootio.go
chait-slim: @DmitriyLewen Looks great, thanks. I have one comment on the tests: the test json does not have any root fixes/range with root versions. My concern there is that it might not be testing the real use case. A real example of a snippet from a json:
`        {
          "pkg": {
            "name": "openssl",
            "cves": {
              "CVE-2024-13176": {
                "vulnerable_ranges": [
                  "<3.0.15-1~deb12u1.root.io.1", ">3.0.15-1~deb12u1.root.io.1 <3.0.16-1~deb12u1"
                ],
                "fixed_versions": [
                  "3.0.15-1~deb12u1.root.io.1",
                  "3.0.16-1~deb12u1"
                ]
              }
            }
          }
        },`

pkg/vulnsrc/rootio/testdata/happy/vuln-list/rootio/test-feed.json
DmitriyLewen: I think the file was just copied from vuln-list-update.
I updated test files in vuln-list-update - https://github.com/aquasecurity/vuln-list-update/pull/357/commits/c04ca887ebe285e9d39fabd08f607738f743031a
and used this file in this PR - https://github.com/aquasecurity/trivy-db/pull/546/commits/55d3c0dd28da7d36baaca44940e6e6d45a9f8247

pkg/vulnsrc/rootio/rootio.go
DmitriyLewen: > @DmitriyLewen Just to validate, the comment here:
// Merge the advisories from the original distributors with Root.io's advisories
Its not really merging but more: override distro-only fix if root fix exists. Correct?

We merge maps, so same keys will be overwrited.
But i added comment in https://github.com/aquasecurity/trivy-db/pull/546/commits/f6595cc208c8ea12ff0c10b71ae5ebf10d277267

DmitriyLewen: > @DmitriyLewen Looks great, thanks. I have one comment on the tests: the test json does not have any root fixes/range with root versions. My concern there is that it might not be testing the real use case. A real example of a snippet from a json:

I added this test case - https://github.com/aquasecurity/trivy-db/pull/546/commits/55d3c0dd28da7d36baaca44940e6e6d45a9f8247

also i see that actual feed uses another format.
I said about same case here - https://github.com/aquasecurity/vuln-list-update/pull/357#issuecomment-3007898913
Are you updated cases with `||`, right?

benjifin: Thanks for catching this - The feed needs to be updated with your request to move from `||` to seperate lines in an array.
We will do that shortly - you are correct that it should be:

```
"vulnerable_ranges": [
                  "<2.9.14+dfsg-1.3~deb12u1.root.io.3",
                  ">2.9.14+dfsg-1.3~deb12u1.root.io.3 <2.9.14+dfsg-1.3~deb12u1.root.io.5"
                ],
```

We will update the feed accordingly.

pkg/vulnsrc/rootio/rootio.go
```diff
@@ -0,0 +1,225 @@
...
+
+func (vs VulnSrc) Name() types.SourceID {
+	return source.ID
+}
+
+func (vs VulnSrc) Update(dir string) error {
+	feedFilePath := filepath.Join(dir, "vuln-list", rootioDir, feedFileName)
+	eb := oops.In("rootio").With("file_path", feedFilePath)
+
+	feedFile, err := os.Open(feedFilePath)
```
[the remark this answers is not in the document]
DmitriyLewen: fixed in https://github.com/aquasecurity/trivy-db/pull/546/commits/00172e17cb6cf5cf1bb01cbecd0eb9e3219fa73f

DmitriyLewen:
Test build in CI/CD:
https://github.com/DmitriyLewen/trivy-db/actions/runs/15923198333/job/44914474180
new DB - https://github.com/DmitriyLewen/trivy-db/pkgs/container/trivy-db

knqyf263:
We modified the existing image for testing. If you have an official Root.io image, we can give it a try.

knqyf263:
Yes, it's pullable. @DmitriyLewen Can you take a look?

DmitriyLewen:
I created fix for `dpkg` - github.com/aquasecurity/trivy-db/pull/546
It should works now:
before:
```bash
➜  trivy -q image rootioinc/common-elasticsearch:9.0.1-bookworm-slim_rootio --table-mode detailed | grep zip -A 6 
...
│ zip                │ CVE-2018-13410      │          │              │ 3.0-13                     │               │ Info-ZIP Zip 3.0, when the -T and -TT command-line options   │
│                    │                     │          │              │                            │               │ are used,...                                                 │
│                    │                     │          │              │                            │               │ https://avd.aquasec.com/nvd/cve-2018-13410                   │
├────────────────────┼─────────────────────┼──────────┼──────────────┼────────────────────────────┼───────────────┼──────────────────────────────────────────────────────────────┤
│ zlib1g             │ CVE-2023-45853      │ CRITICAL │ will_not_fix │ 1:1.2.13.dfsg-1.root.io.1  │               │ zlib: integer overflow and resultant heap-based buffer       │
│                    │                     │          │              │                            │               │ overflow in zipOpenNewFileInZip4_6                           │
│                    │                     │          │              │                            │               │ https://avd.aquasec.com/nvd/cve-2023-45853                   │
└────────────────────┴─────────────────────┴──────────┴──────────────┴────────────────────────────┴───────────────┴──────────────────────────────────────────────────────────────┘

```
after:
```bash
➜ ./trivy -q image rootioinc/common-elasticsearch:9.0.1-bookworm-slim_rootio --table-mode detailed | grep zip -A 3
│ zip              │ CVE-2018-13410      │ CRITICAL │          │ 3.0-13                     │               │ Info-ZIP Zip 3.0, when the -T and -TT command-line options   │
│                  │                     │          │          │                            │               │ are used,...                                                 │
│                  │                     │          │          │                            │               │ https://avd.aquasec.com/nvd/cve-2018-13410                   │
└──────────────────┴─────────────────────┴──────────┴──────────┴────────────────────────────┴───────────────┴──────────────────────────────────────────────────────────────┘

```

DmitriyLewen:
This issue should not affect Alpine images, but it is best to double-check all supported distributions.

> In terms of process, @knqyf263 is there any procedure for a hot fix for this, or this will now wait till the next release?

We are already merging the fix. After merging the PR, I will write you a commit with the fix (to build the binary) and a canary image with this fix (depending on which approach is more convenient for you).
if you confirm that there are no other problems - we will cut `v0.64.1`

benjifin:
Checked on Alpine and is working as expected :)

knqyf263:
You can track this milestone.
https://github.com/aquasecurity/trivy/milestone/56?closed=1

pkg/vulnsrc/rootio/rootio.go
DmitriyLewen: Hello @chait-slim  
Sorry for the ping.

We found that the root.io feed contains `||`.  
We previously agreed that separate array elements would be used instead of a combined expression (see the message above), so Trivy-db does not handle this case.  
However, the presence of `||` causes false positives.

Is this expected on your side?

chait-slim: Im the one that should be sorry @DmitriyLewen!
Its a complete mishap on our side. We are pushing the fix now. The feed should be fixed next cycle

DmitriyLewen: Thanks a lot for the quick response and for the fix!
We’re really looking forward to the updates.

PR #554, opened 2025-07-10
fix(rootio): add severities from base OS advisories

DmitriyLewen:
## Description
There are cases when advisory bucket contains severity (e.g. for Debian).
We need to use this severity for `root.io` advisories (if Base OS and root.io contains same vulnID).
e.g. if there are advisories for both `Root.io` and `Debian` - we need to add Debian severity (if exists) into `root.io` advisory and return in `Get` function

### DIFF
--- a/pkg/vulnsrc/rootio/rootio_test.go
+++ b/pkg/vulnsrc/rootio/rootio_test.go
@@ -264,8 +264,8 @@ func TestVulnSrc_Get(t *testing.T) {
 			want: []types.Advisory{
 				{
 					VulnerabilityID: "CVE-2024-10041",
-					Status:          2,
-					Severity:        1,
+					Status:          types.StatusAffected,
+					Severity:        types.SeverityLow,
 				},
 				{
 					VulnerabilityID:    "CVE-2024-22365",


DmitriyLewen:
I'm not sure if this is the right logic.
You want to change the data source in trivy-db, but this is wrong.
Trivy-db contains information only from the `root.io` feed (for root.io recommendations).
In this context, we don't care how root.io gets, merges and fixes this information.

But after your suggestion, I saw that this can be changed in the `Get` function + use your idea with `source` + `baseOS` as ID.
We merge information from root.io + baseOS => therefore we can "merge" data sources here.
I added my thoughts in https://github.com/aquasecurity/trivy-db/pull/554/commits/668fd653a743e3004db9bfcc2bef2e3ae5a312f1 (you can see different dataSources in tests)

knqyf263:
How about the second idea? `debian+rootio` is not wrong as it just means Root.io advisories for Debian-based images.

But I initially thought we could merge advisories when merging them. If you prefer that, I'm fine with that. I just want to simplify https://github.com/aquasecurity/trivy/pull/9181, as it's quite complicated compared to what we want to do.

knqyf263:
>and if we don't do this - we will still have to enrich the advisories in Get function and then it makes sense to add all the logic there.

Their stages are different. Storing the data source with the base information is only for Root.io advisories, which is not relevant to Debian advisories. It also means it doesn't require dynamic changes. We can statically insert data beforehand.

On the other hand, getting severities from Debian advisories is more like a dynamic change. It's challenging to do that while inserting data.

For me, it's simple: anything that can be handled statically is done at the time of data insertion, while tasks that require dynamic processing are performed at the time of data retrieval.

DmitriyLewen:
You're probably right.
These changes are too minor and users are unlikely to notice the difference, so there's no immediate need for them.

I used your logic - https://github.com/aquasecurity/trivy-db/pull/554/commits/6489e43b50dfcc19c15d8dffa335cf0f9b4a3717
Also I simplify your logic (we have `baseOS`, so there is no points to create `Platrofm` struct)

knqyf263:
>Do you mean that we don't need to check Debian in VendorSeverity, because we save severity in AdvisoryDetails?

I was just trying to understand the advantage of your approach.

>In my case - no

Ah, then I now feel like your solution is more confusing. It's mixing up the severity source (`SeveritySource`) and the data source (`DataSource.ID`). We should consider them separately. In the following cases, both utilize advisories from the same data source (Root.io) and don't use Debian advisories for vulnerability detection.

>- for Root.io only:
>    - id: root.io
>- for both Debian + Root.io:
>    - id: root.io+debian

They use different sources for the severity, which means the above is inaccurate, and it should be

- for Root.io only:
  - id: root.io
  - severity source: nvd (or anything chosen in auto detection)
- for Root.io (and Debian also has the advisory for the same vulnerability)
  - id: root.io
  - severity source: debian


Otherwise, your suggestion should be as below in case we use the severity from a different source.

- for Alpine only:
  - id: alpine
- for both Alpine + Red Hat
  - id: alpine+redhat


When going back to the beginning, I feel that the impression of `+` being inaccurate may have arisen because our interpretations of `+` were different. I think the issue was particularly that "+" gave the impression of being something additional. 

- Me: `debian+root.io` (Debian-based Root.io)
    - I tried to explain it in https://github.com/aquasecurity/trivy-db/pull/554#issuecomment-3057305082
- You: `debian+root.io` (Use both Debian and Root.io advisories)
    - I guess that's why you mentioned `baseInfo + extraInfo -> fixedVersions + Severity -> root.io + debian.`.

In Debian versioning, when modifications are made to the upstream version, versioning like "+debian" is used, so I thought `+` could be one possible option—but it seems that ended up causing confusion. Therefore, I'd suggest `root.io/debian` or `root.io//debian`, which just means Debian-based Root.io or Root.io advisories for Debian. It doesn't mean using both Root.io and Debian advisories, since the data source ID must not represent the severity source.

- for Root.io (and Debian doesn't have the advisory for the same vulnerability):
  - id: `root.io//debian` (Debian-based Root.io)
  - severity source: nvd (or anything chosen in auto detection)
- for Root.io (and Debian also has the advisory for the same vulnerability)
  - id: `root.io//debian` (Debian-based Root.io)
  - severity source: debian

DmitriyLewen:
> Otherwise, your suggestion should be as below in case we use the severity from a different source.

Yeah, you are right - my solution also has certain nuances in understanding

> I think the issue was particularly that "+" gave the impression of being something additional.

yes, that's exactly how I looked at it (using `-based` didn't change my perception of it)

> Therefore, I'd suggest root.io/debian or root.io//debian, which just means Debian-based Root.io or Root.io advisories for Debian. It doesn't mean using both Root.io and Debian advisories, since the data source ID must not represent the severity source.

`/` (or `//`) may also add confusion. `/` can be interpreted as `OR`.
e.g. for `root.io/debian` may mean either `root.io` or `debian` source.
Perhaps you should just use the `for` separator? (e.g. `root.io-for-debian`)?

knqyf263:
>The name and URL remain, so it's still helpful for users to identify the data source.

As mentioned [earlier](https://github.com/aquasecurity/trivy-db/pull/554#issuecomment-3057054785), the name and URL are for humans (and end-users), and the ID is for machine processing (and maintainers). So, anything should be fine unless we maintainers are confused, like `root.io__debian`, `root.io::debian`, `root.io--debian`, and whatever.

I thought using symbols might be better since this value is handled programmatically (like Trivy introduced the `file://` prefix for license files), but if you prefer a more human-friendly string like `root.io-for-debian`, I'm totally fine with that too. It's up to you to decide.

DmitriyLewen:
Update PR - https://github.com/aquasecurity/trivy-db/pull/554/commits/8d3698bf073dcbf41b25d61281ad535273fe47c9
 
> if you prefer a more human-friendly string like root.io-for-debian

We save this info into `json` format. 
So I thought it was better to use human-friendly string.

I also updated `#9181` - https://github.com/aquasecurity/trivy/pull/9181/commits/161e6f30a7d0a86e43d62329891f7803bad772f2

DmitriyLewen:
> What if we omit this information in JSON

We can't omit DataSource (or DataSource.ID), because `VendorSeverity` uses `DataSource.ID`.

It is because of `VendorSeverity` that I want to set a human-readable string.
I worked with issues related to`VendorSeverity` and I think (I am not 100% sure about this) that some users work with this map manually.


Anyway we can' know all user usecases, also we spend on this task sufficient amount of time, so i suggest to use currect logic (especially since it suited both of us).

pkg/vulnsrc/rootio/rootio.go
```diff
@@ -203,14 +212,30 @@ func (vs VulnSrcGetter) Get(osVer, pkgName string) ([]types.Advisory, error) {
...
+	allAdvsSlice := lo.Values(allAdvs)
+	sort.Slice(allAdvsSlice, func(i, j int) bool {
+		return allAdvsSlice[i].VulnerabilityID < allAdvsSlice[j].VulnerabilityID
+	})
+
+	return allAdvsSlice, nil
+}
+
+func (vs VulnSrcGetter) BaseOS() types.SourceID {
+	return vs.baseOS
```
knqyf263: Do we still need this method?

DmitriyLewen: Thanks! Removed in https://github.com/aquasecurity/trivy-db/pull/554/commits/6343c871946b3c5b7e8725d37c14a0855438996f

knqyf263:
Regardless of the bug, while `DataSource.ID` is used when selecting the severity, IIRC, it is meaningless to output in the report JSON. Am I missing anything?

I thought it might be used with `trivy convert`, but I believe the `--vuln-severity-source` flag is not something that needs to be supported in `trivy convert`.

knqyf263:
I was simply stating my general view that IDs are meant to be processed by machines and therefore don’t necessarily need to be human-friendly strings. I didn’t mean to imply that they must not be human-friendly! So I think it’s perfectly fine to use a human-friendly string if you prefer — I’ll leave that up to you👍

If including the ID in the JSON report is likely to confuse users, it might be better to hide it; but if no one seems to care, I think it’s fine to leave it in.

PR #555, opened 2025-07-16
feat(db): add DataSource.BaseID field

DmitriyLewen:
## Description
We added baseOS suffixes for `Root.io` advisories in #554.

But after [this discussion](https://github.com/aquasecurity/trivy/pull/9181#discussion_r2209370563) we understand that better and easier way: add `BaseID` field for `DataSource`.

- At the moment this field will be filled for `Root.io` advisories only:
  ```
  "DataSource": {
      "BaseID": "debian",
      "ID": "root.io",
      "Name": "Root.io Security Advisories (debian)",
      "URL": "https://api.root.io/external/patch_feed"
  }
  ```
  BaseOS suffix for Name ( e.g. (Debian) ) to make it clearer what BaseID means.

- The rest `DataSource` will remain unchanged.

## Related PRs
- [x] #554

### DIFF
--- a/pkg/db/advisory.go
+++ b/pkg/db/advisory.go
@@ -40,7 +40,12 @@ func (dbc Config) GetAdvisories(source, pkgName string) ([]types.Advisory, error
 
 		advisory.VulnerabilityID = vulnID
 		if !lo.IsEmpty(v.Source) {
-			advisory.DataSource = &v.Source
+			advisory.DataSource = &types.DataSource{
+				ID:     v.Source.ID,
+				Name:   v.Source.Name,
+				URL:    v.Source.URL,
+				BaseID: v.Source.BaseID,
+			}
 		}
 
 		results = append(results, advisory)
--- a/pkg/db/advisory.go
+++ b/pkg/db/advisory.go
@@ -3,6 +3,7 @@ package db
 import (
 	"encoding/json"
 
+	"github.com/samber/lo"
 	"github.com/samber/oops"
 	bolt "go.etcd.io/bbolt"
 
@@ -38,12 +39,8 @@ func (dbc Config) GetAdvisories(source, pkgName string) ([]types.Advisory, error
 		}
 
 		advisory.VulnerabilityID = vulnID
-		if v.Source != (types.DataSource{}) {
-			advisory.DataSource = &types.DataSource{
-				ID:   v.Source.ID,
-				Name: v.Source.Name,
-				URL:  v.Source.URL,
-			}
+		if !lo.IsEmpty(v.Source) {
+			advisory.DataSource = &v.Source
 		}
 
 		results = append(results, advisory)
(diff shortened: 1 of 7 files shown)


pkg/types/types.go
```diff
@@ -99,6 +99,10 @@ type DataSource struct {
 	ID   SourceID `json:",omitempty"`
 	Name string   `json:",omitempty"`
 	URL  string   `json:",omitempty"`
+
+	// BaseID shows Base source of advisories.
+	// e.g. `Root.io` based on Debian/Ubuntu/Alpine advisories.
+	BaseID SourceID `json:",omitempty"`
```
knqyf263: `omitzero` is equivalent to `omitempty` in case of string, but there's no reason to use `omitempty` after [Go 1.24](https://github.com/aquasecurity/trivy-db/pull/556).

```suggestion
	BaseID SourceID `json:",omitzero"`
```

pkg/db/advisory.go
```diff
@@ -38,12 +39,8 @@ func (dbc Config) GetAdvisories(source, pkgName string) ([]types.Advisory, error
...
 
 		advisory.VulnerabilityID = vulnID
-		if v.Source != (types.DataSource{}) {
-			advisory.DataSource = &types.DataSource{
-				ID:   v.Source.ID,
-				Name: v.Source.Name,
-				URL:  v.Source.URL,
-			}
+		if !lo.IsEmpty(v.Source) {
+			advisory.DataSource = &v.Source
```
knqyf263: Won't it cause any problems now? It was probably fixed in Go 1.22, but I'm still concerned about this kind of code. I want to make sure it now works as expected.
https://go.dev/blog/loopvar-preview

DmitriyLewen: I managed to reproduce this case in trivy-db:
- can't upgrade to `1.21` (trivy-db requires `1.23`)
- couldn't get this case for trivy-db (simpler tests work as described in https://go.dev/blog/loopvar-preview), but it doesn't mean I didn't just miss something.

so I decided to use the old logic, because I'm 100% sure of it - https://github.com/aquasecurity/trivy-db/pull/555/commits/476cfba57f33de13e0faaa1094aa75063a839110

knqyf263: I've struggled with loop variable issues many times, so even if I think it's probably already been fixed and should be fine, I still have doubts. If we can be certain that there's no issue, I think the new code is better since it's simpler. I've approved it, so I'll leave the final decision up to you.

DmitriyLewen: Since you added this secure code - most likely you encountered loop variable issues.
Since I can't check this now and be 100% sure that there is no problem - I will leave the secure code (especially since this code is not much more complicated than the first implementation)
in other situations (where it can be 100% checked) - we will use the simplified code

PR #557, opened 2025-07-20
feat(rocky): Add Rootio Rocky Support

urimils:
Add support for Rocky in Root.io flow.
Since Rocky uses Arch as well as os and pkg versions, added an adapter so both options can be used where "Get" is required

### DIFF
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -62,22 +62,6 @@ type GetParams struct {
 	Arch    string
 }
 
-type TrioGetter interface {
-	Get(string, string, string) ([]types.Advisory, error)
-}
-
-type TwoGetAdapter struct {
-	G Getter
-}
-
-func (a TwoGetAdapter) Get(osVer, pkgName, _ string) ([]types.Advisory, error) {
-	advs, err := a.G.Get(osVer, pkgName)
-	if err != nil {
-		return nil, err
-	}
-	return advs, nil
-}
-
 type Config struct{}
 
 type Option func(*Options)
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -62,6 +62,22 @@ type GetParams struct {
 	Arch    string
 }
 
+type TrioGetter interface {
+	Get(string, string, string) ([]types.Advisory, error)
+}
+
+type TwoGetAdapter struct {
+	G Getter
+}
+
+func (a TwoGetAdapter) Get(osVer, pkgName, _ string) ([]types.Advisory, error) {
+	advs, err := a.G.Get(osVer, pkgName)
+	if err != nil {
+		return nil, err
+	}
+	return advs, nil
+}
+
 type Config struct{}
 
 type Option func(*Options)
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -56,6 +56,22 @@ type Getter interface {
 	Get(string, string) ([]types.Advisory, error)
 }
 
+type TrioGetter interface {
+	Get(string, string, string) ([]types.Advisory, error)
+}
+
+type TwoGetAdapter struct {
+	G Getter
+}
+
+func (a TwoGetAdapter) Get(osVer, pkgName, _ string) ([]types.Advisory, error) {
+	advs, err := a.G.Get(osVer, pkgName)
+	if err != nil {
+		return nil, err
+	}
+	return advs, nil
+}
+
 type Config struct{}
 
 type Option func(*Options)
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -53,7 +53,13 @@ type Operation interface {
 }
 
 type Getter interface {
-	Get(string, string) ([]types.Advisory, error)
+	Get(GetParams) ([]types.Advisory, error)
+}
+
+type GetParams struct {
+	Release string
+	PkgName string
+	Arch    string
 }
 
 type Config struct{}
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -56,22 +56,6 @@ type Getter interface {
 	Get(string, string) ([]types.Advisory, error)
 }
 
-type TrioGetter interface {
-	Get(string, string, string) ([]types.Advisory, error)
-}
-
-type TwoGetAdapter struct {
-	G Getter
-}
-
-func (a TwoGetAdapter) Get(osVer, pkgName, _ string) ([]types.Advisory, error) {
-	advs, err := a.G.Get(osVer, pkgName)
-	if err != nil {
-		return nil, err
-	}
-	return advs, nil
-}
-
 type Config struct{}
 
 type Option func(*Options)
--- a/pkg/vulnsrc/amazon/amazon_test.go
+++ b/pkg/vulnsrc/amazon/amazon_test.go
@@ -150,7 +150,7 @@ func TestVulnSrc_Get(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			vs := amazon.NewVulnSrc()
-			vulnsrctest.TestGet(t, vs, vulnsrctest.TestGetArgs{
+			vulnsrctest.TestGet(t, db.TwoGetAdapter{G: vs}, vulnsrctest.TestGetArgs{
 				Fixtures:   tt.fixtures,
 				WantValues: tt.want,
 				GetParams: db.GetParams{
--- a/pkg/vulnsrc/amazon/amazon_test.go
+++ b/pkg/vulnsrc/amazon/amazon_test.go
@@ -153,9 +153,11 @@ func TestVulnSrc_Get(t *testing.T) {
 			vulnsrctest.TestGet(t, db.TwoGetAdapter{G: vs}, vulnsrctest.TestGetArgs{
 				Fixtures:   tt.fixtures,
 				WantValues: tt.want,
-				Release:    tt.version,
-				PkgName:    tt.pkgName,
-				WantErr:    tt.wantErr,
+				GetParams: db.GetParams{
+					Release: tt.version,
+					PkgName: tt.pkgName,
+				},
+				WantErr: tt.wantErr,
 			})
 		})
 	}
--- a/pkg/vulnsrc/amazon/amazon_test.go
+++ b/pkg/vulnsrc/amazon/amazon_test.go
@@ -5,7 +5,6 @@ import (
 	"path/filepath"
 	"testing"
 
-	"github.com/aquasecurity/trivy-db/pkg/db"
 	"github.com/aquasecurity/trivy-db/pkg/types"
 	"github.com/aquasecurity/trivy-db/pkg/utils"
 	"github.com/aquasecurity/trivy-db/pkg/vulnsrc/amazon"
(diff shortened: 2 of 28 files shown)


pkg/db/db.go
```diff
@@ -56,6 +56,22 @@ type Getter interface {
 	Get(string, string) ([]types.Advisory, error)
 }
 
+type TrioGetter interface {
```
DmitriyLewen: I think that we can create one common interface for all (except RedHat) sources.
i created #558 

After these changes we will easy use any source for root.io

@knqyf263 wdyt?

pkg/vulnsrc/rootio/rootio_test.go
```diff
@@ -171,6 +172,7 @@ func TestVulnSrc_Get(t *testing.T) {
 			args: args{
 				osVer:   "11",
 				pkgName: "openssl",
+				arch:    "amd64",
```
DmitriyLewen: this change can confusing (debian advisories don't use arches).
So let's add arch only for rocky related advisories.

DmitriyLewen (commented):
hello @urimils 
I left a comments.

Also i don't see Rocky advisories in [feed](https://api.root.io/external/cve_feed).

chait-slim:
> Contributor

Rocky feeds will be added on our side very shortly (matter of a week or two) with exactly the same format as Debian and Ubuntu. We our opening this PR now to allow Trivy to catch these on the July release

pkg/db/db.go
chait-slim: Thats exactly the rational behind this part of the PR - the differences between the two argument and three argument getters (otherwise it would of been a very simple addition). Do you want me to rebase above the #558 branch?

DmitriyLewen: > Do you want me to rebase above the https://github.com/aquasecurity/trivy-db/pull/558 branch?

No need this. You will rebase main branch after merge #558

chait-slim: Ok, so I will wait for the merge there before updating this PR (I just wanted to be proactive)

DmitriyLewen: For now you can edit other comments :)

chait-slim: Done (:

pkg/vulnsrc/rootio/testdata/happy/vuln-list/rootio/cve_feed.json
```diff
@@ -64,5 +64,27 @@
...
     }
+  ],
+  "rocky": [
+    {
+      "distroversion": "8",
+      "packages": [
+        {
+          "pkg": {
+            "name": "nginx",
+            "cves": {
```
DmitriyLewen, line 75: Will there be architectures in the `root.io` feed?
I think we can't use Rocky's arches because:
1. There is a case where Rocky's advisory doesn't contain a CVE, but root.io has a fix for that CVE.
2. There seems to be a case where `root.io` and `Rocky` advisories contain different architectures.

chait-slim: From our point of view archs don't really matter:
1. A vulnerability is relevant for both archs and we create fixes for both
2. A vulnerability is only relevant for one arch and we create a fix for it

DmitriyLewen: let's see the following example:
There are `CVE-2000-0001` for pkgA (affected versions are `> 0.0.1 < 0.0.5`) for `amd64` only.
You create fix for this CVE and add it into feed (`vulnerable_ranges` is `> 0.0.1 < 0.0.5-root.io.1`)

User scans `pkgA@0.0.2` for `aarch64`.
We find that feed contains `CVE-2000-0001` for this package and show this package as vulnerable.
This is false positive.

chait-slim: I see two options here:
1. I add arch per cve for rocky
2. There can be logic that checks - if no advisory from redhat, dont add the root vulnerability

DmitriyLewen: > I add arch per cve for rocky

 You mean in root.io feed, right?
I think this is best solution.

in this case advisories will be "transparent" for users.
User will clearly see vulnerable version, patched version and architecture.
User will not need to refer to other advisory source, etc.

chait-slim: Ok, after discussing internally, I will update the feed for *all* cves with the archs (to make the feed consistent). From Trivy's pov there shouldn't be any change - the arch will be ignored for Debian, Ubuntu, Alpine and used for Rocky, <any additional src that requires>

DmitriyLewen: we only need arches for Rocky.
The rest of distrs are at your discretion.

BTW changes will be required in vuln-list-update
we did not specify the arch in schema:
https://github.com/aquasecurity/vuln-list-update/blob/be8391dd6bbdc696d65fd50adeb629aabde0a206/rootio/types.go#L23-L27

chait-slim: Thanks, will update

pkg/db/db.go
DmitriyLewen: @chait-slim we merged #558

chait-slim:
@DmitriyLewen Rebased

pkg/vulnsrc/amazon/amazon_test.go
```diff
@@ -5,7 +5,6 @@ import (
 	"path/filepath"
 	"testing"
 
-	"github.com/aquasecurity/trivy-db/pkg/db"
```
DmitriyLewen: We use this import:
https://github.com/aquasecurity/trivy-db/blob/a897efc986f288eeed502d396dae58692de2db33/pkg/vulnsrc/amazon/amazon_test.go#L155

pkg/vulnsrc/azure/azure_test.go
```diff
@@ -4,7 +4,6 @@ import (
 	"path/filepath"
 	"testing"
 
-	"github.com/aquasecurity/trivy-db/pkg/db"
```
DmitriyLewen: same

pkg/vulnsrc/debian/debian_test.go
```diff
@@ -4,7 +4,6 @@ import (
 	"path/filepath"
 	"testing"
 
-	"github.com/aquasecurity/trivy-db/pkg/db"
```
DmitriyLewen: same

pkg/vulnsrc/echo/echo_test.go
```diff
@@ -4,7 +4,6 @@ import (
 	"path/filepath"
 	"testing"
 
-	"github.com/aquasecurity/trivy-db/pkg/db"
```
DmitriyLewen: same

pkg/vulnsrc/photon/photon_test.go
```diff
@@ -1,10 +1,10 @@
 package photon
 
 import (
+	"github.com/aquasecurity/trivy-db/pkg/db"
```
DmitriyLewen: it will be linter error

pkg/vulnsrc/rootio/rootio_test.go
```diff
@@ -360,6 +360,72 @@ func TestVulnSrc_Get(t *testing.T) {
...
 			wantErr: "failed to get advisories",
 		},
+		{
+			name:     "only Root.io rocky advisories",
+			baseOS:   vulnerability.Rocky,
+			fixtures: []string{"testdata/fixtures/happy.yaml"},
+			args: args{
+				osVer:   "8",
+				pkgName: "openssl",
+				arch:    "amd64",
```
DmitriyLewen, line 371: you need to add new field in args:
https://github.com/aquasecurity/trivy-db/blob/9bb3ae6e5a396ec554963b52de3be1e519671e7e/pkg/vulnsrc/rootio/rootio_test.go#L153-L156

pkg/vulnsrc/rootio/rootio_test.go
```diff
@@ -360,6 +360,72 @@ func TestVulnSrc_Get(t *testing.T) {
 			},
 			wantErr: "failed to get advisories",
 		},
+		{
+			name:     "only Root.io rocky advisories",
+			baseOS:   vulnerability.Rocky,
+			fixtures: []string{"testdata/fixtures/happy.yaml"},
```
DmitriyLewen, line 367: add `"testdata/fixtures/data-source.yaml",` here (to see datasource)

pkg/vulnsrc/rootio/testdata/fixtures/happy.yaml
```diff
@@ -77,4 +77,51 @@
...
+    - bucket: openssl
+      pairs:
+        - key: CVE-2023-0464
+          value:
+            FixedVersion: ""
+            Status: 2
+        - key: CVE-2024-13176
+          value:
+            FixedVersion: ""
+            Status: 2
```
DmitriyLewen, line 127: IIRC rocky always contains arch.
Tell me if you know CVEs without arch

pkg/vulnsrc/rootio/testdata/fixtures/happy.yaml
```diff
@@ -77,4 +77,51 @@
...
+            VulnerableVersions:
+              - "<1.1.1k-7.el8_6.root.io.1"
+            PatchedVersions:
+              - "1.1.1k-7.el8_6.root.io.1"
+        - key: CVE-2024-13176
+          value:
+            VulnerableVersions:
+              - "<3.0.7-1.el8.root.io.2"
+            PatchedVersions:
+              - "3.0.7-1.el8.root.io.2"
```
DmitriyLewen, line 115: As we discussed before - root.io advisories will contain arches.
Add arches here.

pkg/vulnsrc/suse-cvrf/suse-cvrf_test.go
```diff
@@ -7,7 +7,6 @@ import (
 
 	"github.com/stretchr/testify/assert"
 
-	"github.com/aquasecurity/trivy-db/pkg/db"
```
DmitriyLewen: as for amazon_test.go

pkg/vulnsrc/rootio/rootio_test.go
DmitriyLewen, line 1: let's add testcase for `Update` function

pkg/vulnsrc/rootio/rootio.go
DmitriyLewen, line 1: We need to add filter be arch in `Get` function.
example from Rocky:
https://github.com/aquasecurity/trivy-db/blob/56ec1e48223854c1688fe24ba8906be58ff253e2/pkg/vulnsrc/rocky/rocky.go#L296-L298

pkg/vulnsrc/rootio/rootio_test.go
urimils: done

DmitriyLewen:
Hello @chait-slim,
Is there any news about Rocky?
It looks like the feed still doesn’t contain Rocky advisories.

PR #558, opened 2025-07-21
feat(db): use GetParams (release, pkgName, arch) for `Getter` interface

DmitriyLewen:
## Description
We have 3 implementation for `Get` function:
- `Get(osVer, pkgName)` - `Getter` interface
- `Get(osVer, pkgName, arch)` - rocky linux + oracle linux
- `Get(pkgName, repositories, nvrs)` - RedHat

This PR refactors `Getter` interface to include an architecture parameter.
These changes create one common interface for all Get function (except RedHat). 
Sources with changed `Get` function - Alma, Alpine, Amazon, Azure, Chainguard, Debian, Echo, Minimos, Oracle, Photon, Rocky, SUSE, Ubuntu, and Wolfi.

These changes will also make it easier to use sources inside another source (eg root.io + debian)

### DIFF
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -53,7 +53,9 @@ type Operation interface {
 }
 
 type Getter interface {
-	Get(string, string) ([]types.Advisory, error)
+	// Get uses the following parameters:
+	// `osVer`, `pkgName`, `arch`
+	Get(string, string, string) ([]types.Advisory, error)
 }
 
 type Config struct{}
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -53,9 +53,13 @@ type Operation interface {
 }
 
 type Getter interface {
-	// Get uses the following parameters:
-	// `osVer`, `pkgName`, `arch`
-	Get(string, string, string) ([]types.Advisory, error)
+	Get(GetParams) ([]types.Advisory, error)
+}
+
+type GetParams struct {
+	Release string
+	PkgName string
+	Arch    string
 }
 
 type Config struct{}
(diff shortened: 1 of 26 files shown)


pkg/db/db.go
```diff
@@ -53,7 +53,9 @@ type Operation interface {
 }
 
 type Getter interface {
-	Get(string, string) ([]types.Advisory, error)
+	// Get uses the following parameters:
+	// `osVer`, `pkgName`, `arch`
+	Get(string, string, string) ([]types.Advisory, error)
```
knqyf263: We may add more params in the future. What if using a struct?

```suggestion
	Get(GetParams) ([]types.Advisory, error)
```

knqyf263 (commented):
This affects Aqua. We need to check with them.

pkg/db/db.go
DmitriyLewen: This makes sense - in this case we won't have to update the Trivy code (after adding new parameters).
Updated in https://github.com/aquasecurity/trivy-db/pull/558/commits/f95b696a092b777963efc6a78d0f7e29bacf2324

DmitriyLewen: This makes sense - in this case we won't have to update the Trivy code (after adding new parameters).
Updated in https://github.com/aquasecurity/trivy-db/pull/558/commits/c78fde1c884db57f7c42b9f5e025c34a0cd22c2b

DmitriyLewen:
> This affects Aqua. We need to check with them.

Yeah. `alma`, `oracle-oval` and `rocky` contain `Get` function in `DB` interface.
I used `Getter` interface for them.

DmitriyLewen:
Created https://github.com/aquasecurity/trivy/pull/9239 with required changes for Trivy

DmitriyLewen:
Aqua team offline approved this PR.

knqyf263 (approved):
Looks great!

PR #566, opened 2025-08-19
feat(seal): add seal source

DmitriyLewen:
## Description
This PR adds support for Seal Security OSV vulnerability source and refactors the OSV package to use functional options. The implementation includes comprehensive ecosystem support for Linux distributions (Alpine, CentOS, CBL-Mariner,
  Debian, Oracle Linux, Red Hat, Ubuntu) and introduces a new Ecosystem struct to handle versioned ecosystems.

## Important changes
  - New Seal Security source: Added pkg/vulnsrc/seal/ with OSV-based vulnerability data source supporting multiple Linux distributions
  - OSV functional options refactor: Replaced New(dir, sourceID, dataSources, transformer, bucketNameFunc) with New(..., opts ...Option) pattern using WithTransformer() and WithBucketNameFunc() options
  - Ecosystem struct enhancement: Refactored osv.Advisory.Ecosystem from types.Ecosystem to osv.Ecosystem{Name, Version} struct to support versioned ecosystems like "redhat:8"
  - Extended ecosystem support: Added constants for Alpine, CentOS, CBL-Mariner, Debian, Oracle Linux, Red Hat, and Ubuntu ecosystems
  - Bucket naming strategy: Added BucketNameFunc to allow custom bucket naming for different vulnerability sources (e.g., Seal uses format: "seal redhat 8")

## Related PRs
- [x] aquasecurity/vuln-list-update/pull/364


pkg/vulnsrc/bucket/bucket.go
```diff
@@ -263,6 +265,39 @@ func NewSUSELinuxEnterpriseMicro(version string) Bucket {
...
+
+func (p sealBucket) Ecosystem() ecosystem.Type {
+	// The OSV package uses the ecosystem to avoid duplicates.
+	// So we use the base ecosystem here.
+	return p.baseEcosystem
+}
+
+func (p sealBucket) DataSource() types.DataSource {
+	return p.datasource
+}
```
knqyf263: I intended something like below.

```suggestion
type sealBucket struct {
	base  DataSourceBucket
}

func (p sealBucket) Name() string {
	return fmt.Sprintf("seal %s, p.base.Name())
}

func (p sealBucket) Ecosystem() ecosystem.Type {
	return p.base.Ecosystem()
}

func (p sealBucket) DataSource() types.DataSource {
	return p.base.DataSource()
}
```

DmitriyLewen: We’ll have different cases for the seal advisory bucket names:
- seal alpine
- seal debian
- seal Red Hat 8

Is this okay?

Also we don't have datasource for `osbucket`.
I suggest to use:
```go
type sealBucket struct {
	base       Bucket
	dataSource types.DataSource
}
```
You can see changes in https://github.com/aquasecurity/trivy-db/pull/566/commits/ec216fb6da7d004a7813388e68fd323d1d4d3ce5

DmitriyLewen: I wanted to use `Bucket` for the `NewSeal` function, but it doesn’t work for the `Get` function.

We could move the logic to update the `DataSource` (add the `BaseID`) into the `Bucket` package, but I’m still not sure about this.
@knqyf263, wdyt?

knqyf263: >I wanted to use Bucket for the NewSeal function, but it doesn’t work for the Get function.

Can you elaborate on this? Do you mean something like the following?

```
func NewSeal(base Bucket, dataSource types.DataSource) (Bucket, error) {
```

knqyf263: >seal alpine
>seal debian
>seal Red Hat 8

I haven't been able to follow the entire discussion, so I may be overlooking something, but why is only Red Hat versioned while the other distributions are not?

DmitriyLewen: > Do you mean something like the following?
func NewSeal(base Bucket, dataSource types.DataSource) (Bucket, error) {

Yeah. I wanted to use `Bucket` instead of `ecosystem`, but if we pass `Bucket` parameters to the function, we’ll need to convert the ecosystem to a `Bucket` in two places before calling `NewSeal`:
- the `Get` function
- the OSV package

I don’t think using the `osv` package in the `Get` function is good practice. So I think the better place for this is the `NewSeal` function (as i created in https://github.com/aquasecurity/trivy-db/pull/566/commits/ec216fb6da7d004a7813388e68fd323d1d4d3ce5).

DmitriyLewen: > I haven’t been able to follow the entire discussion, so I may be overlooking something, but why is only Red Hat versioned while the other distributions are not?

The first approach used multiple ecosystems per distro and version:
- alpine: 3.21
- ubuntu: 22.04
- redhat: 8
- etc.

But after the OSV ecosystem change (https://github.com/aquasecurity/vuln-list-update/pull/364#issuecomment-3248063376), `Seal` kept versioning only for Red Hat.
I think they use a single package for all Alpine/Debian/Ubuntu releases, but different packages for RedHat releases.
@SealSagish, could you elaborate on this?

DmitriyLewen: > We could move the logic to update the DataSource (add the BaseID) into the Bucket package, but I’m still not sure about this.

My thought:
```diff
diff --git a/pkg/vulnsrc/bucket/bucket.go b/pkg/vulnsrc/bucket/bucket.go
index 3e5a389..9dd5f6f 100644
--- a/pkg/vulnsrc/bucket/bucket.go
+++ b/pkg/vulnsrc/bucket/bucket.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/aquasecurity/trivy-db/pkg/ecosystem"
 	"github.com/aquasecurity/trivy-db/pkg/types"
+	"github.com/aquasecurity/trivy-db/pkg/vulnsrc/vulnerability"
 )
 
 const separator = "::"
@@ -287,18 +288,27 @@ func (p sealBucket) DataSource() types.DataSource {
 
 // NewSeal creates a bucket for Seal ecosystem
 func NewSeal(baseEco ecosystem.Type, baseEcoVer string, dataSource types.DataSource) (Bucket, error) {
+	eb := oops.In("seal").With("base_ecosystem", baseEco).With("base_ecosystem_ver", baseEcoVer)
+
+	if lo.IsEmpty(dataSource) {
+		return nil, eb.Errorf("data source cannot be empty")
+	}
+
 	bkt := sealBucket{
 		dataSource: dataSource,
 	}
 	switch baseEco {
 	case ecosystem.Alpine:
 		bkt.base = NewAlpine("")
+		bkt.dataSource.BaseID = vulnerability.Alpine
 	case ecosystem.Debian:
 		bkt.base = NewDebian("")
+		bkt.dataSource.BaseID = vulnerability.Debian
 	case ecosystem.RedHat:
 		bkt.base = NewRedHat(baseEcoVer)
+		bkt.dataSource.BaseID = vulnerability.RedHat
 	default:
-		return nil, oops.With("base_ecosystem", baseEco).Errorf("unsupported base ecosystem for Seal bucket")
+		return nil, eb.Errorf("unsupported base ecosystem for Seal bucket")
 	}
 
 	return bkt, nil
diff --git a/pkg/vulnsrc/osv/osv.go b/pkg/vulnsrc/osv/osv.go
index 2df4b5d..c0e7048 100644
--- a/pkg/vulnsrc/osv/osv.go
+++ b/pkg/vulnsrc/osv/osv.go
@@ -423,11 +423,6 @@ func (o OSV) convertEcosystem(raw string) (bucket.Bucket, error) {
 }
 
 func (o OSV) sealBucket(baseEcosystem string) (bucket.Bucket, error) {
-	ds, ok := o.dataSources[ecosystem.Seal]
-	if !ok {
-		return nil, oops.With("ecosystem", "seal").With("base ecosystem", baseEcosystem).Errorf("data source cannot be empty")
-	}
-
 	var eco ecosystem.Type
 	// Separate base ecosystem and version (if exists)
 	// e.g. "Alpine", "Red Hat:8", "Debian"
@@ -435,18 +430,15 @@ func (o OSV) sealBucket(baseEcosystem string) (bucket.Bucket, error) {
 	switch ecoStr {
 	case "alpine":
 		eco = ecosystem.Alpine
-		ds.BaseID = vulnerability.Alpine
 	case "debian":
 		eco = ecosystem.Debian
-		ds.BaseID = vulnerability.Debian
 	case "red hat":
 		eco = ecosystem.RedHat
-		ds.BaseID = vulnerability.RedHat
 	default:
 		return nil, oops.With("ecosystem", "seal").With("base ecosystem", baseEcosystem).Errorf("unsupported base ecosystem")
 	}
 
-	return bucket.NewSeal(eco, ver, ds)
+	return bucket.NewSeal(eco, ver, o.dataSources[ecosystem.Seal])
 }
 
 func versionContains(ranges []VersionRange, version string) (bool, error) {
diff --git a/pkg/vulnsrc/seal/seal.go b/pkg/vulnsrc/seal/seal.go
index c957ee4..0b0ede7 100644
--- a/pkg/vulnsrc/seal/seal.go
+++ b/pkg/vulnsrc/seal/seal.go
@@ -67,7 +67,7 @@ func NewVulnSrcGetter(baseEcosystem ecosystem.Type) VulnSrcGetter {
 func (vs VulnSrcGetter) Get(params db.GetParams) ([]types.Advisory, error) {
 	eb := oops.In("seal").With("base_ecosystem", vs.baseEcosystem).With("os_version", params.Release).With("package_name", params.PkgName)
 
-	bkt, err := bucket.NewSeal(vs.baseEcosystem, params.Release, types.DataSource{})
+	bkt, err := bucket.NewSeal(vs.baseEcosystem, params.Release, source)
 	if err != nil {
 		return nil, eb.Wrapf(err, "failed to create a bucket name")
 	}

```

knqyf263: >I don’t think using the osv package in the Get function is good practice. So I think the better place for this is the NewSeal function (as i created in https://github.com/aquasecurity/trivy-db/commit/ec216fb6da7d004a7813388e68fd323d1d4d3ce5).

I don’t mind either way. However, since the processing related to Seal Security is quite specific and not defined in the OSV specifications, I moved it to the seal package. Please let me know if you find any issues.
https://github.com/aquasecurity/trivy-db/pull/566/commits/2647b6fdd47db12209e6ade1dfb2103b7f5776fb

DmitriyLewen: LGTM 👍

DmitriyLewen: > But after the OSV ecosystem change (https://github.com/aquasecurity/vuln-list-update/pull/364#issuecomment-3248063376), Seal kept versioning only for Red Hat.
I think they use a single package for all Alpine/Debian/Ubuntu releases, but different packages for RedHat releases.
@SealSagish, could you elaborate on this?


@SealSagish one more question — I saw advisories only for CBL-Mariner, but Microsoft has renamed this OS to Azure. Do you plan on adding advisories for Azure? (I need to add this OS to [the list of supported versions](https://github.com/DmitriyLewen/trivy/blob/1c9ce372bca1932efd853e2ff3d5acdc30916fa4/pkg/detector/ospkg/seal/provider.go#L12)).

DmitriyLewen: The Seal team responded to me offline:
```
As a rule, in the Debian/Alpine ecosystems, the same package version can be installed for multiple distro major versions.
This is not the case in Red Hat. In principle, el7 is only supposed to be found on Red Hat:7, el8 only on Red Hat:8, and so forth.
So in the Red Hat ecosystem packages are strongly coupled with specific major versions.
That's why it makes logical sense to create an advisory for Red Hat:7, but it makes no sense to create a unique advisory for Debian:7.

Regarding Azure Linux - in general we plan to add support for additional Linux distributions in the future.
Azure Linux isn't one of the distros we currently have.
```

cc. @knqyf263

knqyf263: @SealSagish Debian uses different versions streams for each major version. Even for the same package, they are separated into 2.4 series in bullseye, 2.5 series in bookworm, and 2.6 series in trixie, for example. Since Debian distributes patches for each version, the patch versions differ by major version (e.g. [CVE-2019-13057](https://security-tracker.debian.org/tracker/CVE-2019-13057)). Regardless of which version the user is using, does Seal Security provide only a single fixed version for all major versions?

SealSagish: @knqyf263 Seal Security provides its own series of fixes for a specific vulnerable version.
As an example, in the example CVE you mentioned above - `2.4.47+dfsg-3` is an affected version.
An example of a Seal Security version that fixes this specific version would be `2.4.47+dfsg-3+sp1` for the first version in the series. Further fixes for the same version would yield the `2.4.47+dfsg-3+sp2, 2.4.47+dfsg-3+sp3`, etc... 
Multiple CVEs could be remediated with each "sealed" version of the branch.

This means that the fixed version is per public version and not per CVE.

knqyf263: In that case, what will happen to users who are using, for example, 2.5.12 in Bookworm and are affected by the vulnerability? Do they need to downgrade to 2.4.47+dfsg-3+sp1?

SealSagish: No. They will need to use 2.5.12+sp1 if it exists. We usually tailor versions to the user's existing version rather than they adapt to use ours

knqyf263: It means Seal provides two versions, 2.4.47+dfsg-3+sp1 and 2.5.12+sp1 for Debian. They are usually used in different major versions. How do you distribute advisories in this case? Will two version ranges be written with both of them having `"ecosystem": "seal:Debian"`?

SealSagish: Exactly what you said

knqyf263: @DmitriyLewen Do you think it would be a problem to write multiple version ranges in OS packages without distinguishing major versions? For now, I don’t feel there’s any particular issue, but I’m checking carefully because if we need to split bucket names in the future, it would become a breaking change.

@SealSagish If it’s marked as an experimental feature, users will understand that breaking changes may be introduced, so it’s possible to merge it right now and fix it later if needed.

DmitriyLewen: Since Seal uses OSV, I would consider this situation in terms of ranges rather than fixed versions.
If the ranges for different major versions don’t overlap, this shouldn’t cause any issues.
And if I understood correctly, Seal uses the narrowest possible range (just adding a suffix to the version).

The only nuance is that this might raise questions from users when they see the fixed versions in the Trivy report.

knqyf263: Ah, yes. With the current implementation, does Trivy indeed show multiple fixed versions?

DmitriyLewen: yes (added several ranges for testing):
```bash
├───────────────────────┼─────────────────────┼──────────┤              ├──────────────────────┼─────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────┤
│ seal-openssl          │ CVE-2024-12797      │ HIGH     │              │ 3.0.11-1~deb12u2+sp1 │ 3.0.11-1~deb12u2+sp999, 3.1.11-1~deb12u2+sp999, │ openssl: RFC7250 handshakes with unauthenticated servers     │
│                       │                     │          │              │                      │ 3.2.11-1~deb12u2+sp999                          │ don't abort as expected                                      │
│                       │                     │          │              │                      │                                                 │ https://avd.aquasec.com/nvd/cve-2024-12797                   │
│                       ├─────────────────────┼──────────┤              │                      ├─────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────┤


```

knqyf263: @SealSagish Is it acceptable?

SealSagish: @DmitriyLewen
Can you elaborate on which ranges you used in this example? 
As you said, we use only the narrowest range to not catch multiple versions in the vulnerable range.
Example - `3.0.11+sp1` will have the range `[3.0.11-3.0.11+sp999]`. While there might be also a range `[3.2.11-3.2.11+sp999]`, it never overlaps other ranges in the Seal advisory.
 
Does Trivy show all the fixed versions regardless of matching ranges or does it only show multiple fixed versions where there are overlaps?

DmitriyLewen: I added few ranges:
```json
 "events": [
            {
              "introduced": "3.0.11-1~deb12u2"
            },
            {
              "fixed": "3.0.11-1~deb12u2+sp999"
            },
            {
              "introduced": "3.1.11-1~deb12u2"
            },
            {
              "fixed": "3.1.11-1~deb12u2+sp999"
            },
            {
              "introduced": "3.2.11-1~deb12u2"
            },
            {
              "fixed": "3.2.11-1~deb12u2+sp999"
            }
          ]
        }
```

> Does Trivy show all the fixed versions regardless of matching ranges or does it only show multiple fixed versions where there are overlaps?

Trivy simply shows all versions from the `patchedVersions` array.
We populate this array with every version listed in the `fixed` field of an `event`:
https://github.com/DmitriyLewen/trivy-db/blob/2647b6fdd47db12209e6ade1dfb2103b7f5776fb/pkg/vulnsrc/osv/osv.go#L341

For my example:
	1.	We add 3 versions to the `patchedVersions` array: "3.0.11-1~deb12u2+sp999", "3.1.11-1~deb12u2+sp999", "3.2.11-1~deb12u2+sp999".
	2.	Trivy then displays these versions in the report.

knqyf263: @DmitriyLewen Is it possible to split advisories when retrieving them with Get in trivy-db, or when there are multiple ranges on the Trivy side? The patched versions would also need to be adjusted accordingly, but if there is no overlap between the ranges, it might work correctly.
https://github.com/aquasecurity/trivy-db/pull/566/files#diff-a8a1c79d6296f805924a307fcd92bd0eeb51216ce88ca6c51a12d1e7f524417dR99-R102

```
{
    "VulnerabilityID": "CVE-2024-12797",
    "VulnerableVersions": [
       ">1.1, < 1.2",
       ">1.3, < 1.4",
    ],
    "PatchedVersions": ["1.2", "1.4"]
}
```

↓

```
[
  {
      "VulnerabilityID": "CVE-2024-12797",
      "VulnerableVersions": [
         ">1.1, < 1.2",
      ],
      "PatchedVersions": ["1.2"]
  },
  {
      "VulnerabilityID": "CVE-2024-12797",
      "VulnerableVersions": [
         ">1.3, < 1.4",
      ],
      "PatchedVersions": ["1.4"]
  },
```

If we use this format when saving to the DB, the size will increase, so it seems better to split them at retrieval time. In the future, it would be nice if we could change it to something like `Ranges: [VulnerableVersion, PatchVersion]`, but for now that’s difficult, so a workaround is needed.

DmitriyLewen: No problem. i will update this PR today.

Wouldn’t it be better to show the user all fixed versions?
Or, since some versions are made for specific customers, could that only confuse them?

knqyf263: >Wouldn’t it be better to show the user all fixed versions?

I think that is true when it comes to programming language libraries. For example, if you are using Django 3.1.0 and 3.1.1 is a bugfix release, it is still useful to know that it has been fixed in 4.0.2. 

However, in the case of Seal Security, bugfix versions for different major versions are bundled together and distributed, so I don’t think the same applies. For instance, someone using 2.4.47 on Debian 10 cannot upgrade to 2.5.12 unless they upgrade to Debian 11. In other words, displaying 2.4.47+dfsg-3+sp1 is fine, but even if 2.5.12+sp1 is shown, it cannot be used. Therefore, I think showing multiple bugfix versions would only cause confusion. 

@SealSagish  For these reasons, I believe it is more appropriate to separate major versions (such as `seal:debian 10`). We'll release it as `seal:debian` with the workaround for now, but I hope you will also consider splitting major versions on the advisory side in the future.

DmitriyLewen: @SealSagish Could there be multiple ranges within a single OS?
e.g. `>=1.21-1+deb11u1, <1.21-1+deb11u1+sp999` + `>=1.21-6+deb11u1, <1.21-6+deb11u1+sp999`

DmitriyLewen: updated logic - https://github.com/aquasecurity/trivy-db/pull/566/commits/e93927ed6882e8e906e3d6ff979d76c044ad5a8a

Trivy shows only one fixed version :
```bash
├───────────────────────┼─────────────────────┼──────────┤              ├──────────────────────┼────────────────────────┼──────────────────────────────────────────────────────────────┤
│ seal-openssl          │ CVE-2024-12797      │ HIGH     │              │ 3.0.11-1~deb12u2+sp1 │ 3.0.11-1~deb12u2+sp999 │ openssl: RFC7250 handshakes with unauthenticated servers     │
│                       │                     │          │              │                      │                        │ don't abort as expected                                      │
│                       │                     │          │              │                      │                        │ https://avd.aquasec.com/nvd/cve-2024-12797                   │
│                       ├─────────────────────┼──────────┤              │                      │                        ├──────────────────────────────────────────────────────────────┤

```

SealSagish: @DmitriyLewen There absolutely can be more than 1 range within a single OS (and same distribution). 
Let's take a look at https://osv.dev/vulnerability/DEBIAN-CVE-2025-11083 as an example. If you go to the Debian:12 tab you'd see that there are many affected versions. Each could potentially be remediated individually on Seal's platform, meaning that there will be a range for as many versions as we fixed from this list. And that's for Debian:12 alone. Changing the data on our end, not only will probably be a great pain on our end, but I don't really know if it benefits the cause on your end as well sadly.

DmitriyLewen: Got it, thanks.

I think in most cases it’s enough for users to see the nearest fixed version.
If they want to use a newer one, they can upgrade and scan again, or check the Seal feed.
But first and foremost, this will resolve the confusion around OS versions.

knqyf263: >but I don't really know if it benefits the cause on your end as well sadly.

We've been explaining that it would benefit users/us. If the major version differs, different version series will be installed. By properly separating these, we can potentially prevent incorrect patch versions from being displayed, and it also makes advisories easier to read because humans do not need to look at unrelated version ranges. For example, Debian 12 users do not need to look at the affected ranges for Debian 11. The same applies when we assist in resolving issues reported by users. In fact, other security vendors distribute advisories in this way.

However, as @DmitriyLewen mentioned, since we have implemented a workaround here, the issue is currently avoided and is not an urgent matter.

PR #570, opened 2025-09-03
feat(rootio): add language package support for security patches

chait-slim:
Extends Root.io vulnerability source to support language ecosystem packages
  (pip, npm, rubygems, maven, go, nuget, cargo) in addition to OS packages.

  - Add NewEcosystemVulnSrcGetter() for language-specific lookups
  - Implement updateLanguagePackages() to process language feeds
  - Support Root.io version suffix pattern (e.g., 4.0.2.root.io)
  - Add comprehensive tests for language package functionality
  - Maintain full backward compatibility with OS packages


chait-slim:
cc: @knqyf263 @DmitriyLewen

pkg/vulnsrc/rootio/rootio.go
```diff
@@ -121,6 +167,56 @@ func (vs VulnSrc) Update(dir string) error {
...
+			}
+			return eb.Wrapf(err, "failed to open feed file %s", feedFilePath)
+		}
+
+		var rawLangFeed RawLanguageFeed
+		if err = json.NewDecoder(feedFile).Decode(&rawLangFeed); err != nil {
+			feedFile.Close()
+			return eb.With("file_path", feedFilePath).Wrapf(err, "json decode error")
+		}
+		feedFile.Close()
```
DmitriyLewen: You can create a separate function to open and parse a language file.  
Then you can use `defer` to close the language file.

pkg/vulnsrc/rootio/rootio.go
```diff
@@ -121,6 +167,56 @@ func (vs VulnSrc) Update(dir string) error {
 	return nil
 }
 
+func (vs VulnSrc) updateLanguagePackages(dir string) error {
+	// Process each supported ecosystem
+	for _, ecosystem := range vs.supportedEcosystems {
+		feedFilePath := filepath.Join(dir, "vuln-list", rootioDir, string(ecosystem), "feed.json")
```
DmitriyLewen: use const for file name

pkg/vulnsrc/rootio/rootio.go
```diff
@@ -146,6 +242,30 @@ func (vs VulnSrc) save(baseOS string, feeds map[string][]Feed) error {
...
+			}
+		}
+		return nil
+	})
+	if err != nil {
+		return oops.Wrapf(err, "batch update error")
+	}
+	return nil
+}
+
```
DmitriyLewen: it looks like you can use `save` function.

chait-slim:
@DmitriyLewen addressed all comments

pkg/vulnsrc/rootio/rootio.go
```diff
@@ -24,9 +24,12 @@ import (
 )
 
 const (
-	rootioDir      = "rootio"
-	feedFileName   = "cve_feed.json"
-	platformFormat = "root.io %s %s" // "root.io {baseOS} {version}"
+	rootioDir       = "rootio"
+	osFeedFileName  = "os_feed.json"  // OS vulnerability data
+	appFeedFileName = "app_feed.json" // Language ecosystem vulnerability data
+	feedFileName    = "cve_feed.json" // Legacy OS feed filename for backward compatibility
```
DmitriyLewen: We updated these files in vuln-list-update PR

Can you update this PR?
1. New file paths
2. IIUC now both files use same struct.

pkg/vulnsrc/rootio/rootio.go
```diff
@@ -113,23 +180,80 @@ func (vs VulnSrc) Update(dir string) error {
...
+		ecosystem := types.Ecosystem(ecosystemStr)
+
+		// Check if this ecosystem is supported
+		if !slices.Contains(vs.supportedEcosystems, ecosystem) {
+			vs.logger.Warn("Unsupported ecosystem in app feed", "ecosystem", ecosystem)
+			continue
+		}
+
+		// Convert to internal format
+		platformName := fmt.Sprintf(ecosystemFormat, ecosystem)
```
DmitriyLewen: You can use Bucket and Bucket Name now - https://github.com/aquasecurity/trivy-db/pull/572

pkg/vulnsrc/rootio/rootio.go
```diff
@@ -185,7 +312,26 @@ func NewVulnSrcGetter(baseOS types.SourceID) VulnSrcGetter {
 	}
 }
 
+// NewEcosystemVulnSrcGetter creates a getter for language ecosystem packages
+func NewEcosystemVulnSrcGetter(ecosystem types.Ecosystem) VulnSrcGetter {
```
DmitriyLewen: I think we can use NewVulnSrcGetter

pkg/vulnsrc/rootio/rootio.go
```diff
@@ -233,6 +379,27 @@ func (vs VulnSrcGetter) Get(params db.GetParams) ([]types.Advisory, error) {
 	return allAdvsSlice, nil
 }
 
+func (vs VulnSrcGetter) getEcosystemAdvisories(params db.GetParams) ([]types.Advisory, error) {
```
DmitriyLewen: As I wrote in https://github.com/aquasecurity/trivy/pull/9427#discussion_r2393809076 - I think we need to merge advisories (from default sources + root.io) here

pkg/vulnsrc/rootio/rootio.go
```diff
@@ -94,14 +138,37 @@ func (vs VulnSrc) Update(dir string) error {
...
+					// Old format
+					pkgName = pkg.Pkg.Name
+					cvesMap = pkg.Pkg.CVEs
+				} else {
+					// New format
+					pkgName = pkg.Name
+					cvesMap = pkg.CVEs
+				}
+
+				for cveID, cveInfo := range cvesMap {
```
DmitriyLewen: IIUC we no longer need to check for the old/new format.

pkg/vulnsrc/rootio/rootio.go
```diff
@@ -113,28 +169,94 @@ func (vs VulnSrc) Update(dir string) error {
...
+	appFeedPath := filepath.Join(dir, "vuln-list", rootioDir, appSubDir, feedFileName)
+	eb := oops.In("rootio").With("file_path", appFeedPath)
+
+	var rawAppFeed RawAppFeed
+	if err := vs.openAndParseJSON(appFeedPath, &rawAppFeed); err != nil {
+		// App feed might not exist yet, skip if file not found
+		if os.IsNotExist(err) {
+			vs.logger.Debug("App feed not found")
+			return nil
+		}
```
DmitriyLewen: We will merge this PR only after the PR for vuln-list-update is merged.
So this code is not needed here.

However, if we forget to remove it, we might miss an error in case vuln-list does not contain application files.

pkg/vulnsrc/rootio/rootio.go
```diff
@@ -41,6 +44,16 @@ var (
 		vulnerability.Debian,
```
DmitriyLewen: we can use ecosystems here.

pkg/vulnsrc/rootio/rootio.go
```diff
@@ -186,6 +203,10 @@ func NewVulnSrcGetter(baseOS types.SourceID) VulnSrcGetter {
 }
 
 func (vs VulnSrcGetter) Get(params db.GetParams) ([]types.Advisory, error) {
```
DmitriyLewen, line 199: We need to think about this.

How we will get advisories for language packages.

chait-slim: WDYT about the implementation chosen? generating a bucket based on the ecosystem and having only a single attribute being used

DmitriyLewen: We can’t use this approach.

If we use the default bucket naming (<ecosystem>::<data source>), Trivy will display Root.io fixed versions for all users.
That would be incorrect — root.io advisories should only be shown for root.io images.

Therefore, the only option is to use the provider logic for language packages and merge advisories in the Get function.
``` mermaid
flowchart TD
  A[Inspect packages in result] --> B{Does version include “root.io” suffix?}
  B -- Yes --> C[Create provider]
  C --> D[Use Get function. We should merge standard advisories + root.io advisories in Get function]
  B -- No --> F[Use standard detector]
```

chait-slim: I think I've solved it now

pkg/vulnsrc/rootio/rootio_test.go
```diff
@@ -132,7 +132,161 @@ func TestVulnSrc_Update(t *testing.T) {
...
+				{
+					Key: []string{
+						"advisory-detail",
+						"CVE-2024-1234",
+						"pip::Root.io Security Patches",
+						"django",
+					},
+					Value: types.Advisory{
+						VulnerableVersions: []string{"<4.0.2"},
+						PatchedVersions:    []string{"4.0.2+root.io.1"},
```
DmitriyLewen, line 241: Hello @chait-slim,
I’m working on updating this and the related Trivy PR.

I found that versions with the +root.io suffix are local versions in Python —
see [PEP 440](https://packaging.python.org/en/latest/specifications/version-specifiers/#local-version-identifiers).
However, such versions shouldn’t be used in repositories.

Did you plan to change this?

chait-slim: Hi @DmitriyLewen, according to [PEP440](https://packaging.python.org/en/latest/specifications/version-specifiers/#local-version-identifiers):
> Local version identifiers SHOULD NOT be used when publishing upstream projects to a public index server, but MAY be used to identify private builds created directly from the project source. Local version identifiers SHOULD be used by downstream projects when releasing a version that is API compatible with the version of the upstream project identified by the public version identifier, but contains additional changes (such as bug fixes)

We are the downstream project in this case and we are pushing to a private index server, rather than a public one. So we do abide to this

DmitriyLewen: Hello @chait-slim 

One more question:
I looked more closely at PEP 440. It states:
> “Except where specifically noted below, local version identifiers MUST NOT be permitted in version specifiers, and local version labels MUST be ignored entirely when checking if candidate versions match a given version specifier.” 

So we can’t use local versions in comparison specifiers (e.g., < 4.0.3+root.io.1).
(Our https://github.com/aquasecurity/go-pep440-version returns an error in such cases.)

"", "==", and "!=" are allowed for local versions.
For vulnerableVersions, we can do the following instead:
	•	<= 4.0.3 instead of < 4.0.3+root.io.1
	•	<= 4.0.3 || 4.0.3+root.io.1 instead of < 4.0.3+root.io.2

Would this approach work for you,
or are you planning to use local versions together with "<", ">", etc.? (i ask, because i saw these cases in tests in PRs)

chait-slim: Actually, 100% yes. For application patches we deal with exact versions. I will update the test cases

DmitriyLewen: > I will update the test cases

No need to do this, please.
I’m working on updating this and the Trivy PR, and I already have some local changes :)
I’ll update the tests myself.

chait-slim: @DmitriyLewen Just to validate, looking at all the PRs, I dont see outstanding comments I need to address on my side

DmitriyLewen: Hello @chait-slim,
I have a question about `npm`:
IIUC, `npm` doesn’t compare [build metadata](https://semver.org/#spec-item-10)￼.
So for `npm`, `4.17.20+root.io.5` == `4.17.20+root.io.1`.

How do you want to compare these versions?

chait-slim: @DmitriyLewen Sorry for the late response. Following this comment we made some changes on our side. For now, I've reduced the checks for python and java

DmitriyLewen: > For now, I've reduced the checks for python and java

Just to make sure I understood you correctly — you mean that your feed will include advisories only for Python and Java at the initial stage of support, right?

> python and java

We've already discussed Python and found a solution.
I'll check the Maven comparator and get back to you.

chait-slim: Yes sir.
Thanks

DmitriyLewen: Hello @chait-slim 
one small question:

Are the versions in the tests correct?
Do you intend to use the `+root.io.x` suffix for Maven?

This version is valid and compares correctly.
```bash
➜ java -jar ~/.m2/repository/org/apache/maven/maven-artifact/3.9.11/maven-artifact-3.9.11.jar "6.0.0" "6.0.0+root.io.1"
Display parameters as parsed by Maven (in canonical form and as a list of tokens) and comparison result:
1. 6.0.0 -> 6; tokens: [6]
   6.0.0 < 6.0.0+root.io.1
2. 6.0.0+root.io.1 -> 6-+root.io.1; tokens: [6, [+root, io, 1]]
```

However, in Maven the `+` character is treated as a common character inside the qualifier, not as a separator (unlike -, _, or .).(https://maven.apache.org/pom.html#version-order-specification)
That’s why I want to double-check whether you intentionally want to use `+` here, instead of a more typical separator such as `-`.

DmitriyLewen: Hello @chait-slim 

> So we can’t use local versions in comparison specifiers (e.g., < 4.0.3+root.io.1).
(Our [aquasecurity/go-pep440-version](https://github.com/aquasecurity/go-pep440-version?rgh-link-date=2025-10-29T13%3A24%3A39Z) returns an error in such cases.)
"", "==", and "!=" are allowed for local versions.
For vulnerableVersions, we can do the following instead:
• <= 4.0.3 instead of < 4.0.3+root.io.1
• <= 4.0.3 || 4.0.3+root.io.1 instead of < 4.0.3+root.io.2
Would this approach work for you,
or are you planning to use local versions together with "<", ">", etc.? (i ask, because i saw these cases in tests in PRs)

> Actually, 100% yes. For application patches we deal with exact versions. I will update the test cases

I have news about this one.
After repeated feedback about using local versions, we decided to offer an alternative.
We prepared a PR for our library to add the possibility to use local versions in Python specifiers.
https://github.com/aquasecurity/go-pep440-version/pull/10

Enabling this option will require changes in Trivy's code, so let me know if you decide to use it before updating your feed.

PR #571, opened 2025-09-09
feat(ghsa): use nuget advisories in lower case

DmitriyLewen:
## Description

  This PR implements case-insensitive storage for NuGet packages in the vulnerability database while maintaining backward compatibility. The changes include:

  - NuGet Package Normalization: Modified NormalizePkgName in pkg/vulnsrc/vulnerability/vulnerability.go:261 to convert NuGet package names to lowercase, making them case-insensitive as per NuGet specification
  - Duplicate Advisory Storage: Enhanced TransformAdvisories in pkg/vulnsrc/ghsa/ghsa.go:143 to store both lowercase and original-case versions of NuGet advisories for backward compatibility
  - Comprehensive Test Coverage: Added test case in pkg/vulnsrc/ghsa/ghsa_test.go:446 with example NuGet advisory data to verify the dual storage functionality

  The implementation stores duplicate entries during a transition period - one with the normalized (lowercase) package name and another with the original case to ensure existing downstream consumers continue to work.

## Reasons for Change

  1. NuGet Case-Insensitivity Compliance: NuGet package names are case-insensitive by design, but the current implementation was storing them with their original case, leading to potential inconsistencies and missed vulnerability matches
  2. Backward Compatibility Requirements: Existing downstream tools and integrations may rely on the original case format, so we need to maintain both versions during a migration period
  3. Standardization: This change aligns the vulnerability database with NuGet's official case-insensitive behavior, improving accuracy and consistency of vulnerability detection
  4. Future Migration Path: The TODO comment indicates this is a transitional approach, with plans to eventually remove the original-case entries once downstream consumers migrate to the lowercase format

## DB changes
buckets:
- before:  
  изображение
- after: 
  изображение

sizes:
- before:
	```bash
	➜  du -sh ./assets-original/trivy.db 
	715M    ./assets-original/trivy.db
	```
- after:
	```bash
	➜  du -sh ./assets/trivy.db
	716M    ./assets/trivy.db
	```

### DIFF
--- a/pkg/vulnsrc/ghsa/ghsa.go
+++ b/pkg/vulnsrc/ghsa/ghsa.go
@@ -126,28 +126,25 @@ func (t *transformer) TransformAdvisories(advisories []osv.Advisory, entry osv.E
 		// Fill severity from GHSA
 		advisories[i].Severity = severity
 
-		// Replace a git URL with a CocoaPods package name in a Swift vulnerability
-		// and store it as a CocoaPods vulnerability.
-		if adv.Ecosystem == vulnerability.Swift {
+		switch adv.Ecosystem {
+		case vulnerability.Swift:
+			// Replace a git URL with a CocoaPods package name in a Swift vulnerability
+			// and store it as a CocoaPods vulnerability.
 			adv.Severity = severity
 			adv.Ecosystem = vulnerability.Cocoapods
 			for _, pkgName := range t.cocoaPodsSpecs[adv.PkgName] {
 				adv.PkgName = pkgName
 				advisories = append(advisories, adv)
 			}
-		}
-
-		// Skip a standard Go package as we use the Go Vulnerability Database (govulndb) for standard packages.
-		if adv.Ecosystem == vulnerability.Go {
+		case vulnerability.Go:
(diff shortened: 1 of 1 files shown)


pkg/vulnsrc/ghsa/ghsa.go
```diff
@@ -138,6 +143,17 @@ func (t *transformer) TransformAdvisories(advisories []osv.Advisory, entry osv.E
 				advisories[i].Ecosystem = "" // An empty ecosystem is skipped later
 			}
 		}
+
+		// NuGet is case-insensitive, so we store advisories in lowercase.
+		// However, for backward compatibility, we also keep advisories with the original package name.
+		// TODO: drop storing the original-case entry and keep only the lowercase key once downstream users have migrated.
+		if adv.Ecosystem == vulnerability.NuGet {
```
knqyf263: nit: We may want to use switch here.

```
switch adv.Ecosystem { 
  case vulnerability.Swift:
  case vulnerability.Go:
  case vulnerability.NuGet:
```
}

DmitriyLewen: Make sense. Updated in https://github.com/aquasecurity/trivy-db/pull/571/commits/5bc3cdf61ab2be911b978715ddfcfce117b017db

PR #572, opened 2025-09-11
refactor: centralize bucket naming logic in dedicated package

knqyf263:
## Summary

This refactoring centralizes bucket naming logic from across the codebase into a dedicated `pkg/vulnsrc/bucket` package, replacing scattered string formatting with type-safe bucket functions. The main drivers for this change are:

1. **Distributed logic consolidation**: Bucket naming logic was scattered across 25+ vulnerability source files, leading to code duplication and inconsistent patterns
2. **Vendor flexibility requirement**: Support for scenarios where third-party vendors (e.g., Seal Security) provide patches for existing OS distributions, requiring flexible bucket name generation
3. **Logic reusability**: Enable reuse of existing bucket naming patterns - for example, transforming "alpine 3.11" to "seal alpine 3.11" by building on the existing Alpine bucket logic

### New Infrastructure
- **`pkg/ecosystem/ecosystem.go`**: Centralized ecosystem type definitions, replacing distributed constants
- **`pkg/vulnsrc/bucket/bucket.go`**: Core bucket implementation with type-safe interface

### Bucket Interface Pattern
```go
type Bucket interface {
    Name() string
    Ecosystem() ecosystem.Type
    DataSource() types.DataSource
}
```

### Architecture
- **Standard OS buckets**: Use `newOS()` helper with consistent "ecosystem version" naming (e.g., "alpine 3.20")
- **Special naming buckets**: Embed `osBucket` and override `Name()` for custom formats (e.g., "Red Hat")
- **Language ecosystem buckets**: Use `newLang()` helper with "ecosystem::datasource" naming

## Migration Pattern

**Before:**
```go
const platformFormat = "alma %s"
platformName := fmt.Sprintf(platformFormat, majorVer)
```

**After:**
```go
platformName := bucket.NewAlma(majorVer).Name()
```

## Backward Compatibility

- ✅ **100% maintained**: All existing bucket names preserved exactly
- ✅ **Test compatibility**: Implementation matches all test expectations  
- ✅ **Special naming preserved**: "Red Hat", "amazon linux", "Azure Linux", etc.

## Code Quality Improvements

- **Type safety**: Replaced error-prone string formatting with type-safe functions
- **Import optimization**: Removed unused `fmt` imports, added structured bucket imports
- **Const simplification**: Single-line constants where bucket functions replace formatting

## Extensibility Examples

The new architecture enables flexible vendor scenarios:

```go
// Future vendor extension (conceptual)
type sealSecurityBucket struct {
    osBucket
}

func (b sealSecurityBucket) Name() string {
    return "seal " + b.osBucket.Name()  // "seal alpine 3.11"
}

sealAlpine311 := bucket.NewSeal(ecosystem.Alpine, "3.11").Name() // "seal alpine 3.11"
// Reuses Alpine bucket logic with vendor prefix
```

### DIFF
--- a/pkg/vulnsrc/osv/osv.go
+++ b/pkg/vulnsrc/osv/osv.go
@@ -223,9 +223,6 @@ func (o OSV) parseAffected(entry Entry, vulnIDs, aliases, references []string) (
 			// Skip unsupported ecosystems
 			continue
 		}
-		if lo.IsNil(bkt) { // `bkt == nil` does not work here because of the interface type
-			continue
-		}
 		pkgName := vulnerability.NormalizePkgName(bkt.Ecosystem(), affected.Package.Name)
 		eb := eb.With("ecosystem", bkt.Ecosystem()).With("package_name", pkgName)
 


pkg/vulnsrc/osv/osv.go
```diff
@@ -378,39 +380,37 @@ func parseSeverity(severities []Severity) (string, float64, error) {
...
-		return vulnerability.Maven
-	case "nuget":
-		return vulnerability.NuGet
-	case "hex":
-		return vulnerability.Erlang
-	case "pub":
-		return vulnerability.Pub
-	case "swifturl", "purl-type:swift":
-		// GHSA still uses "purl-type:swift" for Swift advisories.
-		// cf. https://github.com/github/advisory-database/blob/db1cdfb553e48f18aa27d7e929d200563451391a/advisories/github-reviewed/2023/07/GHSA-jq43-q8mx-r7mq/GHSA-jq43-q8mx-r7mq.json#L20
```
knqyf263, line 406: GHSA migrated to SwiftURL
https://github.com/github/advisory-database/blob/9852d42105965196945fdf80e091e10ecfb2f303/advisories/github-reviewed/2023/07/GHSA-jq43-q8mx-r7mq/GHSA-jq43-q8mx-r7mq.json#L20

pkg/vulnsrc/osv/osv.go
```diff
@@ -378,39 +380,37 @@ func parseSeverity(severities []Severity) (string, float64, error) {
...
-		return vulnerability.Swift
-	case "bitnami":
-		return vulnerability.Bitnami
-	case "kubernetes":
-		return vulnerability.Kubernetes
+func (o OSV) convertEcosystem(raw Ecosystem) bucket.Bucket {
+	eco, _, _ := strings.Cut(string(raw), ":")
```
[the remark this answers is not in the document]
knqyf263: The second part will be needed soon.

pkg/ecosystem/ecosystem.go
```diff
@@ -0,0 +1,70 @@
...
+	MinimOS     Type = "minimos"
+)
+
+// String returns the string representation of the ecosystem type
+func (t Type) String() string {
+	return string(t)
+}
+
+// All contains all supported ecosystems (for backward compatibility)
+var All = []Type{
```
DmitriyLewen: These are only language ecosystems.  
Perhaps we want to change the slice name.

pkg/vulnsrc/bucket/bucket.go
```diff
@@ -1,13 +1,313 @@
...
 
 const separator = "::"
 
-func Name(ecosystem types.Ecosystem, dataSource string) string {
-	return fmt.Sprintf("%s%s%s", ecosystem, separator, dataSource)
+// Bucket interface for generating bucket names
+type Bucket interface {
+	Name() string
+	Ecosystem() ecosystem.Type
+	DataSource() types.DataSource
```
DmitriyLewen, line 22: Let’s add a comment that OS advisories don’t use the data source in bucket names.
```suggestion
	DataSource() types.DataSource // Will be empty for osBucket's
```

pkg/vulnsrc/osv/types.go
```diff
@@ -8,14 +8,20 @@ import (
...
+	ecosystemMaven      = "maven"
+	ecosystemNuGet      = "nuget"
+	ecosystemHex        = "hex"
+	ecosystemPub        = "pub"
+	ecosystemSwiftURL   = "swifturl"
+	ecosystemBitnami    = "bitnami"
+	ecosystemKubernetes = "kubernetes"
 )
 
 type Ecosystem string
```
DmitriyLewen: Do we still need this type here now?

pkg/vulnsrc/osv/osv.go
```diff
@@ -378,39 +380,37 @@ func parseSeverity(severities []Severity) (string, float64, error) {
...
-		return vulnerability.Pub
-	case "swifturl", "purl-type:swift":
-		// GHSA still uses "purl-type:swift" for Swift advisories.
-		// cf. https://github.com/github/advisory-database/blob/db1cdfb553e48f18aa27d7e929d200563451391a/advisories/github-reviewed/2023/07/GHSA-jq43-q8mx-r7mq/GHSA-jq43-q8mx-r7mq.json#L20
-		return vulnerability.Swift
-	case "bitnami":
-		return vulnerability.Bitnami
-	case "kubernetes":
-		return vulnerability.Kubernetes
+func (o OSV) convertEcosystem(raw Ecosystem) bucket.Bucket {
```
DmitriyLewen: nit:
```suggestion
// convertEcosystem converts the OSV ecosystem to Bucket.
// If the ecosystem is not supported or dataSources don't contain the ecosystem, it returns nil.
func (o OSV) convertEcosystem(raw Ecosystem) bucket.Bucket {
```

Firstly, I thought that we don’t handle the case when `dataSources` doesn’t have the ecosystem, but after searching, I found this in the `bucket` package:
https://github.com/knqyf263/trivy-db/blob/ae9b1fba6f4766ada0287ae914eeb224061d7072/pkg/vulnsrc/bucket/bucket.go#L66-L68
Let’s add a comment for people like me 😄

pkg/vulnsrc/debian/types.go
```diff
@@ -16,7 +16,7 @@ type VulnerabilityDetail struct {
 	Description string
 }
 
-type bucket struct {
+type Bucket struct {
```
DmitriyLewen: Do we need to make this struct exported?

knqyf263: It's just conflicting with the `bucket` package. We can rename it if you have a good idea.

pkg/vulnsrc/osv/osv.go
knqyf263: Actually, this was a part I struggled with quite a bit, but as you seemed confused, it looks like the design isn’t very good after all. However, unless we do it this way, we’ll end up checking the dataSource in every ecosystem as below. It may be redundant, but from a readability perspective, do you think that approach is better?

```
	case ecosystemGo:
		if ds, ok := o.dataSources[ecosystem.Go]; ok {
			return bucket.NewGo(ds)
		}
	case ecosystemNpm:
		if ds, ok := o.dataSources[ecosystem.Npm]; ok {
			return bucket.NewNpm(ds)
		}
	case ecosystemNpm:
		if ds, ok := o.dataSources[ecosystem.Pip]; ok {
			return bucket.NewNip(ds)
		}
... (repeat for all ecosystems) ...
```

pkg/vulnsrc/osv/types.go
knqyf263: Good point. Will remove it.

pkg/vulnsrc/debian/types.go
DmitriyLewen: Now that we have the Bucket interface, this structure can be a bit confusing (for example, we never insert severity for the bucket name).

it could just be a `entry`.

pkg/vulnsrc/osv/osv.go
DmitriyLewen: > if ds, ok := o.dataSources[ecosystem.Go]; ok {
			return bucket.NewGo(ds)
		}
		
I definitely don’t like this approach here.
There are too many unnecessary lines of code.

If you’re worried that the comment I mentioned might be missed, we can add a comment before the switch stating that the new* functions check whether a data source exists for the ecosystem.

pkg/ecosystem/ecosystem.go
knqyf263: I've moved it to `aqua`.
https://github.com/aquasecurity/trivy-db/pull/572/commits/5ad8b63ce952d9b7bdb2f627ccfe912fb85e0fb2

pkg/vulnsrc/bucket/bucket.go
knqyf263: I split into two interfaces for clarity.
https://github.com/aquasecurity/trivy-db/pull/572/commits/6f961b236ca40955dd6c42bcb112b223519d3bab

pkg/vulnsrc/osv/types.go
knqyf263: Removed
https://github.com/aquasecurity/trivy-db/pull/572/commits/da7d6378dc417b05d0fda963cf48a0634eace351

pkg/vulnsrc/osv/osv.go
knqyf263: In-band errors are not [recommended in Go](https://go.dev/wiki/CodeReviewComments#in-band-errors). I refactored it.
https://github.com/aquasecurity/trivy-db/pull/572/commits/91c78a1cc0db6cf5ba9b24ad8bdf2066c7794748

pkg/vulnsrc/debian/types.go
knqyf263: After all, I chose `advisoryKey`.
https://github.com/aquasecurity/trivy-db/pull/572/commits/6ac84678ed701379c047405a3ca2c3ec590828d2

pkg/vulnsrc/osv/osv.go
```diff
@@ -212,12 +218,16 @@ func (o OSV) parseAffected(entry Entry, vulnIDs, aliases, references []string) (
...
 	uniqAdvisories := map[string]Advisory{}
 	for _, affected := range entry.Affected {
-		ecosystem := convertEcosystem(affected.Package.Ecosystem)
-		if ecosystem == vulnerability.Unknown {
+		bkt, err := o.convertEcosystem(affected.Package.Ecosystem)
+		if err != nil {
+			// Skip unsupported ecosystems
+			continue
+		}
+		if lo.IsNil(bkt) { // `bkt == nil` does not work here because of the interface type
```
DmitriyLewen: nit:
```suggestion
			if err != nil || lo.IsNil(bkt) { // `bkt == nil` does not work here because of the interface type
			// Skip unsupported ecosystems
			continue
		}
```

DmitriyLewen (approved):
LGTM

pkg/vulnsrc/osv/osv.go
knqyf263: Thanks. I forgot to remove it.
https://github.com/aquasecurity/trivy-db/pull/572/commits/953d63d577db8fd096242647e55e8fe5b3e87ab2

PR #594, opened 2025-11-11
feat(suse): Implement handling for SUSE Linux Enterprise Server 16.0

dirkmueller:
Starting with 16, SUSE stopped using the "SP" suffixing and changed to a major.minor format.

### DIFF
--- a/pkg/vulnsrc/suse-cvrf/suse-cvrf.go
+++ b/pkg/vulnsrc/suse-cvrf/suse-cvrf.go
@@ -279,7 +279,8 @@ func (vs VulnSrc) getOSVersion(platformName string) string {
 			return ""
 		}
 
-		ss := strings.Fields(strings.ReplaceAll(platformName, "-", " "))
+		// Handle both 15-SP7 and 16.0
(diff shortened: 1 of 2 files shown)


pkg/vulnsrc/suse-cvrf/suse-cvrf.go
```diff
@@ -279,7 +279,12 @@ func (vs VulnSrc) getOSVersion(platformName string) string {
 			return ""
 		}
 
+		// SUSE Linux Enterprise Server 15-SP7
 		ss := strings.Fields(strings.ReplaceAll(platformName, "-", " "))
+		if strings.Contains(platformName, ".") {
+			// SUSE Linux Enterprise Server 16.0
+			ss = strings.Fields(strings.ReplaceAll(platformName, ".", " "))
+		}
```
knqyf263: If it replaces a dot with a space anyway, do we need to check if it contains a dot?

```suggestion
		// SUSE Linux Enterprise Server 16.0
		ss = strings.Fields(strings.ReplaceAll(platformName, ".", " "))
```

dirkmueller: we do need to continue handling "-" -> " " replacement also. but I get your point..

PR #595, opened 2025-11-13
feat(julia): define the Julia ecosystem

mbauman:
This is the next step in the work towards supporting the new Julia security advisory database (https://github.com/aquasecurity/trivy/discussions/9071).  ~~This depends upon my work in vuln-list-update: https://github.com/aquasecurity/vuln-list-update/pull/386. I'm marking this as a draft as it currently contains a temporary commit to enable CI on my fork (5e1d0a84744613d8bdf25a3c66b75f5dc72bb22f) and it requires https://github.com/aquasecurity/vuln-list-update/pull/386 to actually gather the data.~~

This has now been updated to use the [Julia SecurityAdvisories.jl database](https://github.com/JuliaLang/SecurityAdvisories.jl) directly, by way of the [`generated/osv`](https://github.com/JuliaLang/SecurityAdvisories.jl/tree/generated/osv) branch.  These are valid OSV-formatted advisories using the v1.7.4 schema, running OSV's linting as part of the repository CI.

## CI on my fork

This is against commit https://github.com/mbauman/trivy-db/commit/3721339d452badb8191f277412e0fd53378da94a on my fork's branch `jlsec`

* [x] Go tests: https://github.com/mbauman/trivy-db/actions/runs/19908946353
* [x] DB build: https://github.com/mbauman/trivy-db/actions/runs/19908968268
* [x] DB build against the `jlsec-ci` branch, which changes the workflow to skip dockerhub and ECR, and use my GHCR credentials with https://github.com/mbauman/trivy-db/commit/ddbe16aa550b3815b0b44bbd5db58f93fd01e1e8: https://github.com/mbauman/trivy-db/actions/runs/19909929151
    * [x] Artifact uploaded: https://github.com/mbauman/trivy-db/pkgs/container/trivy-db/598689605?tag=2

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,7 @@ db-fetch-langs:
 	## required to convert GHSA Swift repo links to Cocoapods package names
 	wget -qO - https://github.com/CocoaPods/Specs/archive/master.tar.gz | tar xz -C $(CACHE_DIR)/cocoapods-specs --strip-components=1
 	wget -qO - https://github.com/kubernetes-sigs/cve-feed-osv/archive/main.tar.gz | tar xz -C $(CACHE_DIR)/k8s-cve-feed --strip-components=1
-	wget -qO - https://github.com/JuliaLang/SecurityAdvisories.jl/archive/refs/heads/generated/osv.zip | tar xz -C $(CACHE_DIR)/julia --strip-components=1
+	wget -qO - https://github.com/JuliaLang/SecurityAdvisories.jl/archive/refs/heads/generated/osv.tar.gz | tar xz -C $(CACHE_DIR)/julia --strip-components=1
 
 .PHONY: db-build
 db-build: trivy-db


pkg/vulnsrc/vulnerability/vulnerability.go
```diff
@@ -264,7 +264,7 @@ func NormalizePkgName(ecoType ecosystem.Type, pkgName string) string {
 		// Nuget is case-insensitive
 		// cf. https://github.com/aquasecurity/trivy/issues/9451
 		pkgName = strings.ToLower(pkgName)
-	case ecosystem.Go, ecosystem.Cocoapods:
+	case ecosystem.Go, ecosystem.Cocoapods, ecosystem.Julia:
 		// Go and Cocoapods are case-sensitive
```
aviks: Maybe change the comment to match the code?

mbauman:
```suggestion
		// Go, Cocoapods, and Julia are all case-sensitive
```

DmitriyLewen (commented):
Hello @mbauman 
Thanks for your work!

I left a small comment.

Could you test building using the pipelines?
1.	Create forks of vuln-list-update and vuln-list.
2.	Build the Julia advisories using the pipeline in your vuln-list-update fork.
3.	Create a fork of trivy-db.
4.	Build trivy-db in CI/CD.
5.	Add links to the pipeline runs and the built trivy-db to the PR description.

mbauman:
I started working my way through the pipelines, but this is not trivial.  I have updated the vuln-list-update PR, linking to the GitHub action I ran on my fork/branch that deployed to my fork of vuln-list.

The build-db action on my fork/branch here (https://github.com/mbauman/trivy-db/actions/runs/19373062921/job/55433538025) successfully created the database, but I've not configured dockerhub/GHCR logins to allow the asset to actually go anywhere.  Is this actually required to move forward here?

pkg/vulnsrc/julia/julia.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/aquasecurity/trivy-db/pkg/vulnsrc/vulnerability"
+)
+
+const sourceID = vulnerability.Julia
+
+var osvDir = filepath.Join("vuln-list", "osv", "julia")
+
+type DatabaseSpecific struct {
+	URL string `json:"url"`
+}
```
DmitriyLewen: Do you plan to add these URLs into advisories?

If no - you can remove this struct.
If yes - you need to add TransformAdvisories function

DmitriyLewen:
>  I've not configured dockerhub/GHCR logins to allow the asset to actually go anywhere. Is this required to move forward here?

Usually, I delete the Docker and ECR login and [registries](https://github.com/aquasecurity/trivy-db/blob/729fb118f08043b12b84c2b9aa8f257204b1a63f/.github/workflows/cron.yml#L91-L92) and keep only the GHCR image during a “test run.”
Having such an image speeds up verification because:
1.	There’s no need to rebuild the database (download vuln-list, build the DB, etc.).
2.	In Trivy, I can simply point to this image using --db-repository, which makes testing much faster.

pkg/vulnsrc/julia/julia.go
mbauman: Good catch; this was a holdover of my copying from another ecosystem. I don't believe we need to do anything with our database_specific fields here.

.github/workflows/cron.yml
DmitriyLewen, line 1: I should have said that you could create a separate branch from this one for that change.
Sorry about that.

You can revert the commit for this file.

mbauman: Done!

DmitriyLewen (approved):

pkg/vulnsrc/julia/julia.go
```diff
@@ -0,0 +1,36 @@
...
+
+	"github.com/aquasecurity/trivy-db/pkg/ecosystem"
+	"github.com/aquasecurity/trivy-db/pkg/types"
+	"github.com/aquasecurity/trivy-db/pkg/vulnsrc/osv"
+	"github.com/aquasecurity/trivy-db/pkg/vulnsrc/vulnerability"
+)
+
+const sourceID = vulnerability.Julia
+
+var osvDir = filepath.Join("julia")
```
knqyf263:
```suggestion
const juliaDir = "julia"
```

DmitriyLewen: Yes, I was fixing this and double-checking it.
It’s already corrected in https://github.com/aquasecurity/trivy-db/pull/595/commits/35aaebea424d7bd624072d6be9f6de163b92b7e4

DmitriyLewen:
@knqyf263 @mbauman 

Let’s discuss the `aliases` and `upstream` fields.
Some advisories use the `aliases` field -https://github.com/JuliaLang/SecurityAdvisories.jl/blob/8d003e2d65ac71834e332847ffd108b8be1142ba/2025/JLSEC-2025-1.json#L6-L9
while others use the upstream field - https://github.com/JuliaLang/SecurityAdvisories.jl/blob/8d003e2d65ac71834e332847ffd108b8be1142ba/2025/JLSEC-2025-10.json#L6-L8

Even though an advisory is “adapted” for a specific package (for example, the vulnerable versions are adjusted for a Julia package), it’s still the same vulnerability, so we should show its original ID.

There is a similar situation with OS vendors: even if the package is adapted, repackaged, etc. for a specific package manager, the vulnerability is still the same as in the upstream advisory, and we show the upstream CVE, even though the vendor may have their own ID (e.g. RHSA-XXX).

If I understand correctly, we should also add and display the fields from the upstream section.
But please correct me if I’m wrong.

knqyf263:
>If I understand correctly, we should also add and display the fields from the upstream section.

Yes, I think so.

>Should we store GHSA IDs in VendorIDs, or only JLSEC IDs?

Let's start with only JLSEC-IDs. We can add GHSA-IDs later if needed.

DmitriyLewen:
I hesitated for a while about whether we should parse the upstream field only for Julia, or for the entire OSV package (since other OSV vendors don’t use this field)

But I eventually decided to enable it for all of them.
@knqyf263  Let me know if you think it would be better to apply this change only to Julia.

pkg/vulnsrc/julia/testdata/happy/julia/JLSEC-2025-10.json
```diff
@@ -0,0 +1,52 @@
...
+        {
+            "ranges": [
+                {
+                    "type": "SEMVER",
+                    "events": [
+                        {
+                            "introduced": "0"
+                        },
+                        {
+                            "fixed": "1.0.7+0"
```
DmitriyLewen, line 21: @mbauman  BTW, I wanted to ask:
If I understood the documentation correctly, Julia follows SemVer — https://docs.julialang.org/en/v1/base/base/#Versioning.
But SemVer ignores build metadata - https://semver.org/#spec-item-10
So the +0 suffix isn’t needed.

Or am I missing something?

mbauman: That is correct that Julia follows SemVer, but we do have some packages that have changed their upstream versions within a build delta.  I decided to just record the exact version numbers (including build metadata) in the advisories themselves, even though we don't respect it in version comparisons.

mbauman:
Yeah, the upstream field is new as of v1.7.0 of the schema. I am fairly certain that we should be using alias for advisories like [`JLSEC-2025-5`](https://github.com/JuliaLang/SecurityAdvisories.jl/blob/main/advisories/published/2025/JLSEC-2025-5.md) and upstream for advisories like [`JLSEC-2025-6`](https://github.com/JuliaLang/SecurityAdvisories.jl/blob/main/advisories/published/2025/JLSEC-2025-6.md), but I agree the spec's language isn't entirely clear.  Looking at that upstream OpenBLAS advisory on osv.dev, for example, you can see that nearly all redistributions of OpenBLAS have issued `upstream` advisories, not `alias`ed ones: https://osv.dev/vulnerability/CVE-2021-4048.  Other existing OSV ecosystems have started to similarly move to the upstream field. For example, here's R doing the same: https://github.com/RConsortium/r-advisory-database/issues/9.

Makefile
```diff
@@ -74,6 +74,7 @@ db-fetch-langs:
 	## required to convert GHSA Swift repo links to Cocoapods package names
 	wget -qO - https://github.com/CocoaPods/Specs/archive/master.tar.gz | tar xz -C $(CACHE_DIR)/cocoapods-specs --strip-components=1
 	wget -qO - https://github.com/kubernetes-sigs/cve-feed-osv/archive/main.tar.gz | tar xz -C $(CACHE_DIR)/k8s-cve-feed --strip-components=1
+	wget -qO - https://github.com/JuliaLang/SecurityAdvisories.jl/archive/refs/heads/generated/osv.zip | tar xz -C $(CACHE_DIR)/julia --strip-components=1
```
mbauman: Actually, it looks like this needs to use the tarball instead of a zip:

```suggestion
	wget -qO - https://github.com/JuliaLang/SecurityAdvisories.jl/archive/refs/heads/generated/osv.tar.gz | tar xz -C $(CACHE_DIR)/julia --strip-components=1
```

DmitriyLewen: I have no idea why I used `zip` here.
Thanks for spotting it and fixing it!

PR #258, opened 2022-11-09
feat: "override" database

afdesk:
* close https://github.com/aquasecurity/trivy/issues/2261

PR #599, opened 2025-12-10
feat: add advisory override mechanism

knqyf263:
## Summary
- Add a mechanism to patch or delete vulnerability advisories from upstream sources without modifying the original cached files
- Patches are applied at file read time via `utils.FileWalk` using [jd](https://github.com/josephburnett/jd) diff/patch operations

### Use cases
- **ID remapping**: Add CVE-ID alias to GHSA-only advisories, making CVE the primary key
- **Delete**: Remove false positive or unnecessary advisories  
- **Add references**: Add reference URLs to existing advisories

### Key features
- Path suffix matching with `/` prefix for proper boundary
- Lazy loading of diff files (only read when path matches)
- Delete support via empty patch result (0 bytes)
- Overrides are loaded from `overrides/` directory by default

### Why jd format?
We use jd diff format instead of JSON Patch (RFC 6902) or JSON Merge Patch (RFC 7396):

| Format | Pros | Cons |
|--------|------|------|
| **JSON Patch** | Standard RFC 6902 | Not intuitive, verbose syntax |
| **JSON Merge Patch** | Simple for replacements | Cannot add/remove individual array elements, difficult to reject entire advisory |
| **jd format** | Human-readable, flexible, can generate diffs for any change including file deletion | Non-standard |

### Why override at file read time (FileWalk)?

Several alternative approaches were considered:

1. **Override at DB insertion time**: Difficult because the data structure varies by source and is not known until runtime
2. **Override in each `Update()` method**: Would require changes across many VulnSrc implementations
3. **Per-advisory config approach** (like #258): Cannot handle source-specific transformations flexibly

The current approach (patching raw JSON at file read time) was chosen because:
- Developers can test patches locally using `jd` CLI before committing
- Changes are isolated to `utils.FileWalk` without touching individual VulnSrc implementations
- jd format supports any JSON transformation including array element additions and file deletion

### Limitations

- **Large JSON files**: Difficult to apply diffs to huge single-file feeds (e.g., [Root.io cve_feed.json](https://github.com/aquasecurity/vuln-list/blob/f6bc49d1937b3204b34154daf6b8248b29f8a14d/rootio/cve_feed.json))
- **Non-JSON formats**: Cannot apply patches to non-JSON/YAML files (though vuln-list stores data in JSON format)

For these edge cases, alternative approaches (e.g., source-specific handling or upstream fixes) would be needed.

### Performance

Performance overhead is negligible:

| Build | Overhead |
|-------|----------|
| GHSA only | ~1.3s |
| Full build | ~3.1s (~1.2%) |

### Potential concerns

If upstream advisories are updated, the diff may fail to apply correctly. However:
- Overrides are intended only for truly critical vulnerabilities (maybe once a year or less)
- Overrides are temporary - they should be removed once upstream is fixed
- Set semantics (`{}` in jd) prevent duplicate entries if upstream adds the same value

### Future improvements

- Adding `--overrides` flag to specify a custom overrides directory would require larger changes and will be addressed in a separate PR if needed.

### Example override
This PR also adds an override for GHSA-9qr9-h5gf-34mp to add CVE-2025-55182 alias.

## Test plan
- [x] Unit tests for `pkg/override` package
- [x] Verified with real GHSA build that CVE-2025-55182 becomes the key instead of GHSA-9qr9-h5gf-34mp
- [x] Verified full build performance overhead (~3.1s, ~1.2%)

### DIFF
--- a/pkg/override/override.go
+++ b/pkg/override/override.go
@@ -54,7 +54,7 @@ func Load(overridesDir string) (*Patches, error) {
 	}
 
 	for _, p := range cfg.Patches {
-		eb := eb.With("target", p.Target, "diff", p.Diff)
+		eb := eb.With("target", p.Target).With("diff", p.Diff)
 		if p.Diff == "" {
 			return nil, eb.Errorf("patch entry missing 'diff' field")
 		} else if !filepath.IsLocal(p.Diff) {


knqyf263:
will fix tests

pkg/override/override.go
```diff
@@ -0,0 +1,139 @@
...
+		return nil, eb.Wrapf(err, "failed to parse config.yaml")
+	}
+
+	patches := &Patches{
+		entries:      make([]PatchEntry, 0, len(cfg.Patches)),
+		overridesDir: overridesDir,
+	}
+
+	for _, p := range cfg.Patches {
+		eb := eb.With("target", p.Target, "diff", p.Diff)
```
DmitriyLewen: nit:
```suggestion
		eb := eb.With("target", p.Target).With("diff", p.Diff)
```

pkg/override/override.go
```diff
@@ -0,0 +1,139 @@
...
+
+// hasSuffix checks if path ends with suffix, ensuring proper path boundary.
+// The suffix must start with '/' to ensure path boundary matching.
+// e.g., "/foo/bar.json" matches "/bar.json" but not "ar.json"
+func hasSuffix(path, suffix string) bool {
+	if len(path) < len(suffix) {
+		return false
+	}
+	return strings.HasSuffix(path, suffix)
+}
```
DmitriyLewen: Do we need this function?

We already validated the `/` prefix earlier https://github.com/knqyf263/trivy-db/blob/f38d46266897396ec08bb6e257b483d3dbd6ce1d/pkg/override/override.go#L65-L67
Also, this function doesn’t include that check anyway.

For the comparison, we can simply use strings.HasSuffix(path, suffix).
strings.HasSuffix doesn’t error even if the suffix is longer than the path:
https://go.dev/play/p/xGCJaYsgRi6

DmitriyLewen (commented):
LGTM 
left 2 small comments

DmitriyLewen:
As an idea — we can implement this later if we start using this mechanism more often:
we could add a workflow that runs weekly, checks whether the entries we override are still present, and sends a notification so we can review whether the override is still needed.

pkg/override/override.go
knqyf263: As [documented](https://pkg.go.dev/github.com/samber/oops#With), `With` can take a list of KVs. Do you prefer repeating `With` for readability?
https://go.dev/play/p/30kFZS8V6o1

DmitriyLewen: > Do you prefer repeating With for readability?

Yes, I think it’s easier to read the code when pairs are separated.

pkg/override/override.go
knqyf263: Fixed in 0f2fc10.

pkg/override/override.go
knqyf263: Fixed in b47ea17.

knqyf263:
@DmitriyLewen I've addressed your review comments.

PR #601, opened 2025-12-10
feat(rootio): Support usage of Severity from root.io feed

urimils:
Following https://github.com/aquasecurity/vuln-list-update/pull/393
We've updated the feed to include Severity for each CVE. This PR handles using that info as part of the detection flow.

Also one small update to Makefile of what looks like a bug.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,7 +81,7 @@ db-build: trivy-db
 	./trivy-db build --cache-dir ./$(CACHE_DIR) --output-dir ./$(OUT_DIR) --update-interval 24h
 
 .PHONY: db-compact
-db-compact: $(GOBIN)/bbolt out/trivy.db
+db-compact: $(GOBIN)/bbolt ./$(OUT_DIR)/trivy.db
 	mkdir -p ./$(ASSET_DIR)
 	$(GOBIN)/bbolt compact -o ./$(ASSET_DIR)/trivy.db ./$(OUT_DIR)/trivy.db
 	cp ./$(OUT_DIR)/metadata.json ./$(ASSET_DIR)/metadata.json


pkg/vulnsrc/rootio/rootio.go
```diff
@@ -211,9 +222,6 @@ func (vs VulnSrcGetter) Get(params db.GetParams) ([]types.Advisory, error) {
 	}
 
 	rootAdvs := lo.SliceToMap(advs, func(adv types.Advisory) (string, types.Advisory) {
-		// Debian may contain severity in the advisory.
-		// We need to save this severity for the Root.io advisory.
-		adv.Severity = allAdvs[adv.VulnerabilityID].Severity
```
DmitriyLewen, line 216: You checked the `cveInfo.Severity` field:
https://github.com/urimils/trivy-db/blob/47a5bb5398d4cd1d7e551d7f1e125c27f5f5537c/pkg/vulnsrc/rootio/rootio.go#L104

Is it possible that an advisory doesn’t contain severity info?
That seems possible.

So if the Root.io feed doesn’t include severity but Debian does, should we use Debian’s severity?
Or do you prefer to use our severity selection logic?
(https://trivy.dev/docs/latest/guide/scanner/vulnerability/#severity-selection)

urimils: > Is it possible that an advisory doesn’t contain severity info?
That seems possible.


This [line](https://github.com/urimils/trivy-db/blob/47a5bb5398d4cd1d7e551d7f1e125c27f5f5537c/pkg/vulnsrc/rootio/rootio.go#L104) is for defensive programming. I do not expect it to ever be empty, nor to return an error.

> So if the Root.io feed doesn’t include severity but Debian does, should we use Debian’s severity?

We're actually doing that when creating the feed now. We get severities from Debian, if that doesn't exist we fall back to the `vulnerabilities` bucket in trivy-db. Currently it's redundant information, but once we add alias packages to the feed we need it for them. And it's best to use a shared flow, at the price of some redundancy.

DmitriyLewen: Just to confirm that I understood you correctly:
Root.io will include Debian severity (or severity from other sources) in their own feed, so we don’t need to add Debian severity on our side anymore (as it’s currently done in your code), right?

urimils: Correct.

urimils: Also Alpine/Ubuntu.

DmitriyLewen (approved):

PR #626, opened 2026-02-03
refactor: use redhat dataSource for redhat-oval

DmitriyLewen:
## Description                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                  
  This PR consolidates the Red Hat OVAL data source to use the existing RedHat data source identifier instead of maintaining a separate RedHatOVAL identifier.   

See more details in https://github.com/aquasecurity/trivy/pull/9192#issuecomment-3834727738
                                    
## Changes                                                                                                                                                                                                                                                                                   
  - Data Source Consolidation: Changed redhat-oval vulnsrc to use vulnerability.RedHat instead of vulnerability.RedHatOVAL                                                                                                                        
  - Updated Data Source Metadata:                                                                                                                                                                                                                 
    - ID: redhat-oval → redhat                                                                                                                                                                                                                    
    - Name: Red Hat OVAL v2 → Red Hat                                                                                                                                                                                                             
    - URL: https://www.redhat.com/security/data/oval/v2/ → https://access.redhat.com/security/cve/                                                                                                                                                
  - Removed RedHatOVAL constant: Deleted the separate RedHatOVAL source ID from vulnerability/const.go and AllSourceIDs slice                                                                                                                     
  - Updated Tests: Modified all test expectations in redhat-oval_test.go and test fixtures to reflect the new data source values                                                                                                                  
                                                                                                                                                                                                                                                  
## Related  PRs
- aquasecurity/trivy/pull/9192
- aquasecurity/trivy/pull/10300


knqyf263:
I'd suggest a slightly different approach: **keep the `RedHatOVAL` constant but only change the `DataSource`**.

`Name()` (which returns `RedHatOVAL`) is used as a tracker identifier — for example, in aqua-db's `skipVulnSrcs` to selectively skip the OSS OVAL tracker and replace it with a custom one. If `Name()` returns `"redhat"`, it becomes impossible to distinguish the OVAL tracker from the Security Data API tracker.

On the other hand, `DataSource.ID` is what users see in `SeveritySource` and `--vuln-severity-source`. Unifying this to `"redhat"` is the right call since users don't care about the internal OVAL vs Security Data API distinction.

Here's the suggested diff:

**`pkg/vulnsrc/redhat-oval/redhat-oval.go`** — change only `DataSource`, not `Name()`:
```diff
 	source = types.DataSource{
-		ID:   vulnerability.RedHatOVAL,
-		Name: "Red Hat OVAL v2",
-		URL:  "https://www.redhat.com/security/data/oval/v2/",
+		// Use the same ID as the Security Data API ("redhat") so that users see
+		// a single, unified "redhat" severity source instead of the internal
+		// "redhat-oval" distinction, which is meaningless to them.
+		ID:   vulnerability.RedHat,
+		Name: "Red Hat",
+		URL:  "https://access.redhat.com/security/cve/",
 	}
```

**`pkg/vulnsrc/vulnerability/const.go`** — keep `RedHatOVAL`, only remove it from `AllSourceIDs`:
```diff
 const (
 	NVD                   types.SourceID = "nvd"
 	RedHat                types.SourceID = "redhat"
-	RedHatOVAL            types.SourceID = "redhat-oval"
+	RedHatOVAL            types.SourceID = "redhat-oval" // Used only for tracker Name() identification
 	RedHatCSAFVEX         types.SourceID = "redhat-csaf-vex"
```

```diff
 var AllSourceIDs = []types.SourceID{
 	NVD,
 	RedHat,
-	RedHatOVAL,
 	Debian,
```

`AllSourceIDs` is used to iterate over `VulnerabilityDetail` entries in the `vulnerability` bucket, but OVAL doesn't store data there (it uses the advisory bucket), so removing `RedHatOVAL` from this list has no effect.

DmitriyLewen:
Yes, that makes sense.
Updated in d7c7ecb3a177c337b230a5e2f3dd5317c354ab2b

BTW, in your diff there are no changes for the `Name()` function (previously `source.ID` was used), let me know if this was intentional.

Also I built the database and checked - everything works correctly.
- `Red Hat` buckets exist.
- CVEs from `Vulnerability` bucket contain information (severity, CVSS, References) from Red Hat Security API.

PR #629, opened 2026-02-10
fix(makefile): separate download and extraction steps for stability

### LINKED ISSUE #627: TrivyDB job failing.
FYI guys, the last few runs of the TrivyDB job have failed.
Regards.

DmitriyLewen:
## Summary 
Replace piped `wget | tar` with two-step download-then-extract approach                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                               
## Details
When using pipe, if wget encounters a recoverable error (timeout, connection drop), it closes the pipe on its end. This causes tar to receive EOF and exit completely, even though wget could have recovered. 
By separating the steps, wget can use its built-in retry mechanism (20 attempts by default) without affecting tar. 

## Test runs
- https://github.com/DmitriyLewen/trivy-db/actions/runs/21856329850/job/63073946275
- https://github.com/DmitriyLewen/trivy-db/actions/runs/21856334341/job/63073959061

## Related issues
- Closes #627

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,10 +9,11 @@ ASSET_DIR ?= assets
 # Usage: $(call download_and_extract,URL,TARGET_DIR)
 # This approach prevents pipe failures when wget encounters recoverable errors
 define download_and_extract
-	@echo "Downloading $(1)..."
-	wget -q $(1) -O $(CACHE_DIR)/tmp-archive.tar.gz
-	tar xzf $(CACHE_DIR)/tmp-archive.tar.gz -C $(2) --strip-components=1
-	rm -f $(CACHE_DIR)/tmp-archive.tar.gz
+	@echo "Downloading $(1)..." && \
+	TMP_FILE=$$(mktemp) && \
+	wget -q $(1) -O "$$TMP_FILE" && \
+	tar xzf "$$TMP_FILE" -C $(2) --strip-components=1 && \
+	rm -f "$$TMP_FILE"
 endef
 
 GOPATH=$(shell go env GOPATH)


Makefile
```diff
@@ -5,6 +5,16 @@ CACHE_DIR ?= cache
...
 
+# Download and extract function - separates download from extraction for stability
+# Usage: $(call download_and_extract,URL,TARGET_DIR)
+# This approach prevents pipe failures when wget encounters recoverable errors
+define download_and_extract
+	@echo "Downloading $(1)..."
+	wget -q $(1) -O $(CACHE_DIR)/tmp-archive.tar.gz
+	tar xzf $(CACHE_DIR)/tmp-archive.tar.gz -C $(2) --strip-components=1
+	rm -f $(CACHE_DIR)/tmp-archive.tar.gz
+endef
```
knqyf263, line 17: Consider using `mktemp` for safer temporary file handling:

```makefile
define download_and_extract
	@TMP_FILE=$$(mktemp) && \
	wget -q $(1) -O "$$TMP_FILE" && \
	tar xzf "$$TMP_FILE" -C $(2) --strip-components=1 && \
	rm -f "$$TMP_FILE"
endef
```

Benefits:
- Files are created in `/tmp`, so the OS auto-cleans them on reboot if the process fails mid-way
- Unique filenames prevent conflicts if parallel execution is ever introduced
- `&&` chaining ensures subsequent commands don't run if an earlier one fails (necessary since this is a single shell command with line continuations)

DmitriyLewen: make sense.
Updated in https://github.com/aquasecurity/trivy-db/pull/629/commits/9482eefcdb86cd349f16010061aec1fa48a8e171

knqyf263: Note that I just suggested my idea and didn't check if it works.

DmitriyLewen: I checked locally - it works correctly.
pipeline also correctly builds trivy-db - https://github.com/DmitriyLewen/trivy-db/actions/runs/21857919225/job/63079143620

PR #636, opened 2026-02-18
feat(bottlerocket): add vulnerability source for Bottlerocket OS

0intro:
Parse Bottlerocket security advisories from vuln-list JSON files and store per-CVE per-package fixed versions in the database, using a single flat "bottlerocket" bucket with no version subdivision.

## Related PRs
- [ ] https://github.com/aquasecurity/trivy/pull/10207
- [ ] https://github.com/aquasecurity/vuln-list-update/pull/406


mohamed-challal (approved):

pkg/vulnsrc/bottlerocket/bottlerocket.go
```diff
@@ -0,0 +1,162 @@
...
+	Version string `json:"version,omitempty"`
+	Release string `json:"release,omitempty"`
+	Arch    string `json:"arch,omitempty"`
+}
+
+type Reference struct {
+	Href string `json:"href,omitempty"`
+	ID   string `json:"id,omitempty"`
+	Type string `json:"type,omitempty"`
+}
```
DmitriyLewen: nit:
Move these structures into types.go

DmitriyLewen: BTW - we can decode only used fields.

pkg/vulnsrc/bottlerocket/bottlerocket.go
```diff
@@ -0,0 +1,162 @@
...
+
+func (vs VulnSrc) Name() types.SourceID {
+	return source.ID
+}
+
+func (vs VulnSrc) Update(dir string) error {
+	rootDir := filepath.Join(dir, "vuln-list", bottlerocketDir)
+	eb := oops.In("bottlerocket").With("root_dir", rootDir)
+
+	err := utils.FileWalk(rootDir, vs.walkFunc)
```
DmitriyLewen: nit:
I don't think that we need to create separate function (take a look other sources).

pkg/vulnsrc/bottlerocket/bottlerocket.go
```diff
@@ -0,0 +1,162 @@
...
+	return nil
+}
+
+func (vs VulnSrc) commit(tx *bolt.Tx) error {
+	if err := vs.dbc.PutDataSource(tx, platformName, source); err != nil {
+		return oops.Wrapf(err, "failed to put data source")
+	}
+
+	for _, adv := range vs.advisories {
+		for _, cveID := range adv.CveIDs {
```
DmitriyLewen: let's move [this logic](https://github.com/0intro/vuln-list-update/blob/ed818f36855baddca448381a3d9d914167ec3c4e/bottlerocket/bottlerocket.go#L97-L103) to here.
see  https://github.com/aquasecurity/vuln-list-update/pull/406#discussion_r3452718403

pkg/vulnsrc/bottlerocket/bottlerocket.go
```diff
@@ -0,0 +1,162 @@
...
+)
+
+const bottlerocketDir = "bottlerocket"
+
+var (
+	platformName = bucket.NewBottlerocket().Name()
+	source       = types.DataSource{
+		ID:   vulnerability.Bottlerocket,
+		Name: "Bottlerocket Security Advisories",
+		URL:  "https://advisories.bottlerocket.aws/",
```
DmitriyLewen: nit:
https://advisories.bottlerocket.aws/ doesn't return anything (AccessDenied), so better use link to archive file.
```suggestion
		URL:  "https://advisories.bottlerocket.aws/updateinfo.xml.gz",
```

pkg/vulnsrc/bottlerocket/bottlerocket.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+)
+
+type VulnSrc struct {
+	dbc        db.Operation
+	advisories []Advisory
```
DmitriyLewen: nit:
https://advisories.bottlerocket.aws/ doesn't return anything (AccessDenied), so better use link to archive file.
```suggestion
		URL:  "https://advisories.bottlerocket.aws/updateinfo.xml.gz",
```

pkg/vulnsrc/bucket/bucket.go
```diff
@@ -83,6 +83,9 @@ func newLang(ecoType ecosystem.Type, dataSource types.DataSource) (DataSourceBuc
 // NewAlma creates a bucket for Alma Linux
 func NewAlma(version string) Bucket { return newOS(ecosystem.AlmaLinux, version) }
 
+// NewBottlerocket creates a bucket for Bottlerocket
+func NewBottlerocket() Bucket { return newOS(ecosystem.Bottlerocket, "") }
```
DmitriyLewen, line 89: nit:
Can you add comment why Bottlerocket doesn't have version argument (as for other sources).

pkg/vulnsrc/bottlerocket/bottlerocket.go
```diff
@@ -0,0 +1,162 @@
...
+func (vs VulnSrc) Get(params db.GetParams) ([]types.Advisory, error) {
+	eb := oops.In("bottlerocket")
+	advisories, err := vs.dbc.GetAdvisories(platformName, params.PkgName)
+	if err != nil {
+		return nil, eb.Wrapf(err, "failed to get advisories")
+	}
+	return advisories, nil
+}
+
+func severityFromPriority(priority string) types.Severity {
```
DmitriyLewen: I think it makes sense to change the `priority` word for the argument and function name (the advisory uses the Severity field).

Also, perhaps it makes sense to add protection — strings.ToLower. But up to you.

pkg/vulnsrc/bottlerocket/bottlerocket.go
```diff
@@ -0,0 +1,162 @@
...
+	case "medium":
+		return types.SeverityMedium
+	case "important":
+		return types.SeverityHigh
+	case "critical":
+		return types.SeverityCritical
+	default:
+		return types.SeverityUnknown
+	}
+}
```
DmitriyLewen, line 152: Should we separate advisories by arches?

pkg/vulnsrc/bottlerocket/bottlerocket.go
0intro: Done.

pkg/vulnsrc/bottlerocket/bottlerocket.go
0intro: Done.

pkg/vulnsrc/bottlerocket/bottlerocket.go
0intro: Done.

pkg/vulnsrc/bottlerocket/bottlerocket.go
0intro: Done.

pkg/vulnsrc/bottlerocket/bottlerocket.go
0intro: Done.

pkg/vulnsrc/bucket/bucket.go
0intro: Done.

pkg/vulnsrc/bottlerocket/bottlerocket.go
0intro: Done.

pkg/vulnsrc/bottlerocket/bottlerocket.go
0intro: The feed lists the same fixed version for x86_64 and aarch64 and Trivy's advisory key isn't arch-scoped, so I dropped the unused Arch field instead of separating by arch.

pkg/ecosystem/ecosystem.go
DmitriyLewen, line 1: run `go fmt` for this file.

pkg/vulnsrc/bottlerocket/bottlerocket.go
```diff
@@ -0,0 +1,146 @@
...
+		return oops.Wrapf(err, "failed to put data source")
+	}
+
+	for _, adv := range advisories {
+		// Bottlerocket advisories reference the CVEs they fix; derive the CVE
+		// IDs from those references instead of relying on a pre-computed field.
+		var cveIDs []string
+		var references []string
+		for _, ref := range adv.References {
+			if ref.Type == "cve" {
```
DmitriyLewen, line 90: Is this possible that advisory doesn't have cve-id?
e.g. only ghsa-id.

DmitriyLewen (commented):
LGTM.

left small comments.

Also I tested changes in my fork:
- test run - https://github.com/DmitriyLewen/trivy-db/actions/runs/28162389004/job/83405849052
- new db - https://github.com/DmitriyLewen/trivy-db/pkgs/container/trivy-db/974672707?tag=2
-  Image

pkg/ecosystem/ecosystem.go
0intro: Done.

pkg/vulnsrc/bottlerocket/bottlerocket.go
0intro: Yes, some advisories carry no CVE. I've fixed to fall back to GHSA/BRSA.

pkg/vulnsrc/bottlerocket/bottlerocket.go
```diff
@@ -0,0 +1,153 @@
...
+				cveIDs = append(cveIDs, ref.ID)
+			} else {
+				fallbackIDs = append(fallbackIDs, ref.ID)
+			}
+			references = append(references, ref.Href)
+		}
+		vulnIDs := cveIDs
+		if len(vulnIDs) == 0 {
+			vulnIDs = fallbackIDs
+		}
```
DmitriyLewen: a bit simplify:

```suggestion
		var vulnIDs, fallbackIDs []string
		var references []string
		for _, ref := range adv.References {
			if ref.Type == "cve" {
				vulnIDs = append(vulnIDs, ref.ID)
			} else {
				fallbackIDs = append(fallbackIDs, ref.ID)
			}
			references = append(references, ref.Href)
		}

		if len(vulnIDs) == 0 {
			vulnIDs = fallbackIDs
		}
```

DmitriyLewen: I'm worried about the case when a file has both a BRSA and a GHSA for the same advisory.

In this case we will be showing duplicates in the report.

I checked the current database and currently there are no such cases. So we can leave it as is, but if such cases are planned — we need to add priority for different IDs to exclude duplicates.

0intro: > a bit simplify:

Done.

DmitriyLewen:
Please recreate only the trivy PR.
This PR is correct. Reopen this PR.

DmitriyLewen:
new DB (after last changes):
test run - https://github.com/DmitriyLewen/trivy-db/actions/runs/28177318209/job/83457120659
new DB contains missed (https://github.com/aquasecurity/trivy-db/pull/636#discussion_r3474092383) advisories:
изображение

0intro:
I've re-opened https://github.com/aquasecurity/trivy-db/pull/679.

PR #639, opened 2026-02-28
fix(redhat-csaf): merge CPE list with existing OVAL CPEs

gshanbhag525:
-  added a param to skip adding mappings when csaf (rhel 10) runs alongside oval (rhel 6-9)
- merge CPE list with existing OVAL CPEs instead of overwriting. Prevents CSAF from corrupting OVALs CPE indices by reading existingCPEs and appending new CSAF-only CPEs at the end.
- parse and add rhsa release date from csaf to CustomPutInput
- get PublishDate from advisory

### DIFF
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -55,7 +55,7 @@ type Operation interface {
 	PutRedHatCPEs(tx *bolt.Tx, cpeIndex int, cpe string) (err error)
 	RedHatRepoToCPEs(repository string) (cpeIndices []int, err error)
 	RedHatNVRToCPEs(nvr string) (cpeIndices []int, err error)
-	GetAllRedHatCPEs(tx *bolt.Tx) (cpes map[int]string, err error)
+	GetAllRedHatCPEs(tx *bolt.Tx) (cpes []string, err error)
 }
 
 type Getter interface {
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -55,6 +55,7 @@ type Operation interface {
 	PutRedHatCPEs(tx *bolt.Tx, cpeIndex int, cpe string) (err error)
 	RedHatRepoToCPEs(repository string) (cpeIndices []int, err error)
 	RedHatNVRToCPEs(nvr string) (cpeIndices []int, err error)
+	GetAllRedHatCPEs(tx *bolt.Tx) (cpes map[int]string, err error)
 }
 
 type Getter interface {
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -55,7 +55,7 @@ type Operation interface {
 	PutRedHatCPEs(tx *bolt.Tx, cpeIndex int, cpe string) (err error)
 	RedHatRepoToCPEs(repository string) (cpeIndices []int, err error)
 	RedHatNVRToCPEs(nvr string) (cpeIndices []int, err error)
-	GetAllRedHatCPEs(tx *bolt.Tx) (cpes []string, err error)
+	RedHatCPEs(tx *bolt.Tx) (cpes []string, err error)
 }
 
 type Getter interface {
--- a/pkg/db/redhat_cpe.go
+++ b/pkg/db/redhat_cpe.go
@@ -65,3 +65,37 @@ func (dbc Config) getCPEs(bucket, key string) ([]int, error) {
 	}
 	return cpes, nil
 }
+
+// GetAllRedHatCPEs returns all stored CPE mappings (index -> CPE string).
+// This is used to merge CSAF CPEs with existing OVAL CPEs.
+func (dbc Config) GetAllRedHatCPEs(tx *bolt.Tx) (map[int]string, error) {
+	result := make(map[int]string)
+
+	root := tx.Bucket([]byte(redhatCPERootBucket))
+	if root == nil {
+		return result, nil
+	}
+
+	cpeBucket := root.Bucket([]byte(redhatCPEBucket))
+	if cpeBucket == nil {
+		return result, nil
+	}
+
+	err := cpeBucket.ForEach(func(k, v []byte) error {
+		index, err := strconv.Atoi(string(k))
+		if err != nil {
+			return oops.Wrapf(err, "invalid CPE index key: %s", string(k))
+		}
+		var cpe string
+		if err := json.Unmarshal(v, &cpe); err != nil {
+			return oops.Wrapf(err, "failed to unmarshal CPE value")
+		}
+		result[index] = cpe
+		return nil
+	})
+	if err != nil {
+		return nil, oops.Wrapf(err, "failed to iterate CPE bucket")
+	}
+
+	return result, nil
+}
--- a/pkg/db/redhat_cpe.go
+++ b/pkg/db/redhat_cpe.go
@@ -66,9 +66,9 @@ func (dbc Config) getCPEs(bucket, key string) ([]int, error) {
 	return cpes, nil
 }
 
-// GetAllRedHatCPEs returns all stored CPE strings ordered by their index.
+// RedHatCPEs returns all stored CPE strings ordered by their index.
 // This is used to merge CSAF CPEs with existing OVAL CPEs.
-func (dbc Config) GetAllRedHatCPEs(tx *bolt.Tx) ([]string, error) {
+func (dbc Config) RedHatCPEs(tx *bolt.Tx) ([]string, error) {
 	root := tx.Bucket([]byte(redhatCPERootBucket))
 	if root == nil {
 		return nil, nil
--- a/pkg/db/redhat_cpe.go
+++ b/pkg/db/redhat_cpe.go
@@ -66,21 +66,20 @@ func (dbc Config) getCPEs(bucket, key string) ([]int, error) {
 	return cpes, nil
 }
 
-// GetAllRedHatCPEs returns all stored CPE mappings (index -> CPE string).
+// GetAllRedHatCPEs returns all stored CPE strings ordered by their index.
 // This is used to merge CSAF CPEs with existing OVAL CPEs.
-func (dbc Config) GetAllRedHatCPEs(tx *bolt.Tx) (map[int]string, error) {
-	result := make(map[int]string)
-
+func (dbc Config) GetAllRedHatCPEs(tx *bolt.Tx) ([]string, error) {
 	root := tx.Bucket([]byte(redhatCPERootBucket))
 	if root == nil {
-		return result, nil
+		return nil, nil
 	}
-
 	cpeBucket := root.Bucket([]byte(redhatCPEBucket))
 	if cpeBucket == nil {
-		return result, nil
+		return nil, nil
 	}
-
+	// BoltDB iterates keys in lexicographic order ("0","1","10","2",...),
+	// so we parse each key as an integer and assign by index directly.
+	var result []string
 	err := cpeBucket.ForEach(func(k, v []byte) error {
 		index, err := strconv.Atoi(string(k))
 		if err != nil {
@@ -90,12 +89,14 @@ func (dbc Config) GetAllRedHatCPEs(tx *bolt.Tx) (map[int]string, error) {
 		if err := json.Unmarshal(v, &cpe); err != nil {
 			return oops.Wrapf(err, "failed to unmarshal CPE value")
 		}
+		for len(result) <= index {
+			result = append(result, "")
+		}
 		result[index] = cpe
 		return nil
 	})
 	if err != nil {
 		return nil, oops.Wrapf(err, "failed to iterate CPE bucket")
 	}
-
 	return result, nil
 }
--- a/pkg/vulnsrc/redhat-csaf/csaf.go
+++ b/pkg/vulnsrc/redhat-csaf/csaf.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"path/filepath"
 
+	"github.com/samber/lo"
 	"github.com/samber/oops"
 	bolt "go.etcd.io/bbolt"
 
@@ -154,7 +155,7 @@ func (vs VulnSrc) update(tx *bolt.Tx, dir string) error {
 // mergeCPEList merges the CSAF CPE list with existing OVAL CPEs in the database.
 // Existing OVAL CPE indices are preserved; new CPEs (e.g. RHEL 10-only, not in OVAL) are appended.
 func (vs VulnSrc) mergeCPEList(tx *bolt.Tx, csafCPEs redhatoval.CPEList) (redhatoval.CPEList, error) {
-	existingCPEs, err := vs.dbc.GetAllRedHatCPEs(tx)
+	existingCPEs, err := vs.dbc.RedHatCPEs(tx)
 	if err != nil {
 		return nil, oops.Wrapf(err, "failed to get existing CPEs")
 	}
@@ -163,27 +164,7 @@ func (vs VulnSrc) mergeCPEList(tx *bolt.Tx, csafCPEs redhatoval.CPEList) (redhat
 	if len(existingCPEs) == 0 {
 		return nil, oops.Errorf("no OVAL CPEs in DB; run redhat-oval before redhat-csaf when using WithRunAlongsideOVAL")
 	}
-
-	// Build a set of existing CPE strings (index-ordered slice from DB)
-	existingSet := make(map[string]bool, len(existingCPEs))
-	for _, cpe := range existingCPEs {
-		if cpe != "" {
-			existingSet[cpe] = true
-		}
-	}
-
-	// Start with existing CPEs in order
-	merged := make(redhatoval.CPEList, len(existingCPEs))
-	copy(merged, existingCPEs)
-
-	// Append new CSAF CPEs that don't exist in OVAL
-	for _, cpe := range csafCPEs {
-		if !existingSet[cpe] {
-			merged = append(merged, cpe)
-			existingSet[cpe] = true
-		}
-	}
-
+	merged := redhatoval.CPEList(lo.Uniq(append(existingCPEs, csafCPEs...)))
 	log.Info("Merged CPE list", "oval_count", len(existingCPEs), "csaf_count", len(csafCPEs), "merged_count", len(merged))
 	return merged, nil
 }
--- a/pkg/vulnsrc/redhat-csaf/csaf.go
+++ b/pkg/vulnsrc/redhat-csaf/csaf.go
@@ -51,21 +51,20 @@ func WithCustomPut(put db.CustomPut) Option {
 	}
 }
 
-// WithSkipRepoNVRMappings skips storing repository-to-CPE and NVR-to-CPE mappings.
-// Use this when OVAL already stores these mappings and CSAF should not overwrite them.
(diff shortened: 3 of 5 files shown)


pkg/vulnsrc/redhat-csaf/csaf.go
```diff
@@ -94,7 +105,18 @@ func (vs VulnSrc) update(tx *bolt.Tx, dir string) error {
...
+		// OVAL runs alongside CSAF: merge to preserve OVAL CPE indices.
+		var err error
+		cpeList, err = vs.mergeCPEList(tx, vs.parser.CPEList())
+		if err != nil {
+			return eb.Wrapf(err, "failed to merge CPE list")
+		}
+	} else {
+		// use CSAF CPE list as-is; no OVAL in DB.
+		cpeList = vs.parser.CPEList()
+	}
```
DmitriyLewen: nit:

```suggestion
	cpeList := vs.parser.CPEList()
	if vs.skipRepoNVRMappings {
		// OVAL runs alongside CSAF: merge to preserve OVAL CPE indices.
		var err error
		cpeList, err = vs.mergeCPEList(tx, cpeList)
		if err != nil {
			return eb.Wrapf(err, "failed to merge CPE list")
		}
	}
```

pkg/vulnsrc/redhat-csaf/csaf.go
```diff
@@ -128,6 +155,46 @@ func (vs VulnSrc) update(tx *bolt.Tx, dir string) error {
...
 
+// mergeCPEList merges the CSAF CPE list with existing OVAL CPEs in the database.
+// Existing OVAL CPE indices are preserved, and new CSAF-only CPEs are appended.
+func (vs VulnSrc) mergeCPEList(tx *bolt.Tx, csafCPEs redhatoval.CPEList) (redhatoval.CPEList, error) {
+	existingCPEs, err := vs.dbc.GetAllRedHatCPEs(tx)
+	if err != nil {
+		return nil, oops.Wrapf(err, "failed to get existing CPEs")
+	}
+
+	if len(existingCPEs) == 0 {
```
DmitriyLewen, line 164: We expected that existingCPEs contains CPEs.
Should we return error, if we didn't find CPEs?

pkg/db/redhat_cpe.go
```diff
@@ -65,3 +65,37 @@ func (dbc Config) getCPEs(bucket, key string) ([]int, error) {
...
+		}
+		result[index] = cpe
+		return nil
+	})
+	if err != nil {
+		return nil, oops.Wrapf(err, "failed to iterate CPE bucket")
+	}
+
+	return result, nil
+}
```
DmitriyLewen: Do we need map here?
We can fill slice.
something like that:
```suggestion
// GetAllRedHatCPEs returns all stored CPE strings ordered by their index.
// This is used to merge CSAF CPEs with existing OVAL CPEs.
func (dbc Config) GetAllRedHatCPEs(tx *bolt.Tx) ([]string, error) {
	root := tx.Bucket([]byte(redhatCPERootBucket))
	if root == nil {
		return nil, nil
	}

	cpeBucket := root.Bucket([]byte(redhatCPEBucket))
	if cpeBucket == nil {
		return nil, nil
	}

	// BoltDB iterates keys in lexicographic order ("0","1","10","2",...),
	// so we parse each key as an integer and assign by index directly.
	var result []string
	err := cpeBucket.ForEach(func(k, v []byte) error {
		index, err := strconv.Atoi(string(k))
		if err != nil {
			return oops.Wrapf(err, "invalid CPE index key: %s", string(k))
		}
		var cpe string
		if err := json.Unmarshal(v, &cpe); err != nil {
			return oops.Wrapf(err, "failed to unmarshal CPE value")
		}
		// Or you can create map and convert to slice after iteration, but this way we can avoid an extra loop and potential sorting.
		for len(result) <= index {
			result = append(result, "")
		}
		result[index] = cpe
		return nil
	})
	if err != nil {
		return nil, oops.Wrapf(err, "failed to iterate CPE bucket")
	}

	return result, nil
}

```
It should simplify the `mergeCPEList` fucntion.

pkg/vulnsrc/redhat-csaf/csaf.go
```diff
@@ -50,11 +51,21 @@ func WithCustomPut(put db.CustomPut) Option {
...
+	}
+}
+
 type VulnSrc struct {
-	put        db.CustomPut
-	dbc        db.Operation
-	parser     Parser
-	aggregator Aggregator
+	put                 db.CustomPut
+	skipRepoNVRMappings bool
```
DmitriyLewen: We need to think about this.

We use this function for 2 points:
- don't write into DB RepoToCPE and RepoToNVR.
- merge CPEs from OVAL and CSAF.

Or we need to rename variable to avoid confusing, or we need to create two variables.

pkg/vulnsrc/redhat-csaf/parse.go
```diff
@@ -432,6 +440,19 @@ func (p *Parser) NVRToCPE() iter.Seq2[string, []string] {
...
+func (p *Parser) ReleaseDate(vulnID VulnerabilityID) string {
+	return p.releaseDates[vulnID]
+}
+
+// formatDate extracts the date portion (YYYY-MM-DD) from an RFC3339 timestamp.
+func (p *Parser) formatDate(timestamp string) string {
+	if len(timestamp) >= 10 {
+		return timestamp[:10]
+	}
+	return timestamp
```
DmitriyLewen: I think better to use `time` package.
The following logic should work:
```suggestion
	t, err := time.Parse(time.RFC3339, timestamp)
	if err != nil {
		return ""
	}
	return t.Format(time.DateOnly)
```

pkg/vulnsrc/redhat-csaf/parse.go
```diff
@@ -432,6 +440,19 @@ func (p *Parser) NVRToCPE() iter.Seq2[string, []string] {
 	return maps.All(p.nvrToCPE)
 }
 
+// ReleaseDate returns the advisory release date for a given vulnerability ID.
+func (p *Parser) ReleaseDate(vulnID VulnerabilityID) string {
+	return p.releaseDates[vulnID]
+}
+
+// formatDate extracts the date portion (YYYY-MM-DD) from an RFC3339 timestamp.
+func (p *Parser) formatDate(timestamp string) string {
```
DmitriyLewen, line 450: I think make sense to add small test for this function.

pkg/vulnsrc/redhat-csaf/csaf.go
```diff
@@ -128,6 +155,46 @@ func (vs VulnSrc) update(tx *bolt.Tx, dir string) error {
 	return nil
 }
 
+// mergeCPEList merges the CSAF CPE list with existing OVAL CPEs in the database.
+// Existing OVAL CPE indices are preserved, and new CSAF-only CPEs are appended.
+func (vs VulnSrc) mergeCPEList(tx *bolt.Tx, csafCPEs redhatoval.CPEList) (redhatoval.CPEList, error) {
```
DmitriyLewen, line 157: Can we add tests for this function?

pkg/vulnsrc/redhat-csaf/csaf.go
gshanbhag525: here if No OVAL CPEs exist in DB (e.g. CPEs only affect RHEL 10) then we return csafCVEs of redhat10.
we dont need to reutn error here.

DmitriyLewen: Allow me to clarify a bit.
We run the mergeCPEList function only when the skipRepoNVRMappings option is enabled.
If I understand correctly, we enable the skipRepoNVRMappings option only when the database contains advisories from OVAL.
And if we expect there to be OVAL advisories in the database, but we didn't find CPEs from OVAL, this looks like an unexpected case.

pkg/vulnsrc/redhat-csaf/parse.go
gshanbhag525: sure, added

pkg/vulnsrc/redhat-csaf/csaf.go
gshanbhag525: renamed the variable to "runAlongsideOVAL"

pkg/vulnsrc/redhat-csaf/csaf.go
gshanbhag525: thanks for the explanation, updated to return err.

pkg/vulnsrc/redhat-csaf/csaf.go
```diff
@@ -128,6 +151,43 @@ func (vs VulnSrc) update(tx *bolt.Tx, dir string) error {
...
+	for _, cpe := range csafCPEs {
+		if !existingSet[cpe] {
+			merged = append(merged, cpe)
+			existingSet[cpe] = true
+		}
+	}
+
+	log.Info("Merged CPE list", "oval_count", len(existingCPEs), "csaf_count", len(csafCPEs), "merged_count", len(merged))
+	return merged, nil
+}
```
DmitriyLewen, line 170: I think it can be simpler.
WDYT?
```suggestion
func (vs VulnSrc) mergeCPEList(tx *bolt.Tx, csafCPEs redhatoval.CPEList) (redhatoval.CPEList, error) {
	existingCPEs, err := vs.dbc.GetAllRedHatCPEs(tx)
	if err != nil {
		return nil, oops.Wrapf(err, "failed to get existing CPEs")
	}
	if len(existingCPEs) == 0 {
		return nil, oops.Errorf("no OVAL CPEs in DB; run redhat-oval before redhat-csaf when using WithRunAlongsideOVAL")
	}

	merged := lo.Uniq(append(existingCPEs, csafCPEs...))
	log.Info("Merged CPE list", "oval_count", len(existingCPEs), "csaf_count", len(csafCPEs), "merged_count", len(merged))
	return merged, nil
}
```

pkg/db/db.go
```diff
@@ -55,6 +55,7 @@ type Operation interface {
 	PutRedHatCPEs(tx *bolt.Tx, cpeIndex int, cpe string) (err error)
 	RedHatRepoToCPEs(repository string) (cpeIndices []int, err error)
 	RedHatNVRToCPEs(nvr string) (cpeIndices []int, err error)
+	GetAllRedHatCPEs(tx *bolt.Tx) (cpes []string, err error)
```
DmitriyLewen: nit:
I think `RedHatCPEs` will be better (see https://go.dev/doc/effective_go#Getters)

pkg/vulnsrc/redhat-csaf/parse_test.go
```diff
@@ -167,3 +211,54 @@ func TestParser_DetectStatus(t *testing.T) {
...
+			require.NoError(t, dbc.PutRedHatCPEs(tx, 1, "cpe:/o:redhat:el9::baseos"))
+			var err error
+			merged, err = vs.mergeCPEList(tx, csafCPEs)
+			return err
+		})
+		require.NoError(t, err)
+		want := redhatoval.CPEList{"cpe:/o:redhat:el8::baseos", "cpe:/o:redhat:el9::baseos", "cpe:/o:redhat:el10::baseos"}
+		assert.Equal(t, want, merged)
+	})
+}
```
DmitriyLewen, line 276: I think this is more readable:

```suggestion
func TestMergeCPEList(t *testing.T) {
	tests := []struct {
		name         string
		existingCPEs redhatoval.CPEList
		csafCPEs     redhatoval.CPEList
		want         redhatoval.CPEList
		wantErr      string
	}{
		{
			name:     "no existing OVAL CPEs returns error",
			csafCPEs: redhatoval.CPEList{"cpe:/o:redhat:el10", "cpe:/o:redhat:el10::appstream"},
			wantErr:  "no OVAL CPEs in DB",
		},
		{
			name:         "appends new CSAF CPEs to OVAL CPEs",
			existingCPEs: redhatoval.CPEList{"cpe:/o:redhat:el8::baseos", "cpe:/o:redhat:el9::baseos"},
			csafCPEs:     redhatoval.CPEList{"cpe:/o:redhat:el10::baseos"},
			want: redhatoval.CPEList{
				"cpe:/o:redhat:el8::baseos", "cpe:/o:redhat:el9::baseos", "cpe:/o:redhat:el10::baseos",
			},
		},
		{
			name:         "deduplicates CPEs already present in OVAL",
			existingCPEs: redhatoval.CPEList{"cpe:/o:redhat:el8::baseos", "cpe:/o:redhat:el9::baseos"},
			csafCPEs:     redhatoval.CPEList{"cpe:/o:redhat:el9::baseos", "cpe:/o:redhat:el10::baseos"},
			want: redhatoval.CPEList{
				"cpe:/o:redhat:el8::baseos", "cpe:/o:redhat:el9::baseos", "cpe:/o:redhat:el10::baseos",
			},
		},
	}

	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			tempDir := t.TempDir()
			require.NoError(t, db.Init(tempDir))
			t.Cleanup(func() {
				_ = db.Close()
			})

			dbc := db.Config{}
			vs := NewVulnSrc(WithRunAlongsideOVAL())
			vs.dbc = dbc

			if len(tt.existingCPEs) > 0 {
				require.NoError(t, dbc.BatchUpdate(func(tx *bolt.Tx) error {
					for i, cpe := range tt.existingCPEs {
						require.NoError(t, dbc.PutRedHatCPEs(tx, i, cpe))
					}
					return nil
				}))
			}

			var got redhatoval.CPEList
			err := dbc.BatchUpdate(func(tx *bolt.Tx) error {
				var err error
				got, err = vs.mergeCPEList(tx, tt.csafCPEs)
				return err
			})

			if tt.wantErr != "" {
				require.ErrorContains(t, err, tt.wantErr)
				return
			}
			require.NoError(t, err)
			assert.Equal(t, tt.want, got)
		})
	}
}
```

gshanbhag525: updated tests code

pkg/vulnsrc/redhat-csaf/csaf.go
gshanbhag525: updated code

DmitriyLewen (approved):
LGTM

@knqyf263 can you quick a look?

knqyf263 (commented):
I have some concerns about mixing OVAL and CSAF data in the same DB buckets.

**1. How do you filter RHEL 10 only?**

The CSAF parser parses all RHEL versions, not just RHEL 10. I don't see any version filtering in this PR. Is the filtering done in the `CustomPut` implementation?

**2. Advisory overwrite risk**

`mergeCPEList` addresses CPE index conflicts, but `PutAdvisoryDetail` overwrites existing data without merging. If the same CVE affects both RHEL 9 (OVAL) and RHEL 10 (CSAF), wouldn't the CSAF write overwrite the OVAL advisory entries since they share the same bucket path?

```
advisory-detail → CVE-2025-X → "Red Hat" → kernel
```

**3. Separate DB or separate buckets instead of merging**

Mixing OVAL and CSAF in the same DB buckets was not the original design intent, and I'm concerned about unexpected edge cases. Would it be feasible to either:

- Use a separate DB file for CSAF, or
- Store CSAF data in separate buckets (e.g., `"Red Hat CSAF"` for advisories and `"Red Hat CSAF CPE"` for CPE mappings) and have `Get()` query both

Either approach would:
- Eliminate CPE index conflicts entirely (no merge needed)
- Prevent advisory overwrites
- Keep OVAL data completely untouched
- Make future migration from OVAL to CSAF cleaner

The trade-off is increased DB size due to duplicated data, but if that's acceptable, it keeps the code and DB structure much cleaner.

gshanbhag525:
=> 1. How do you filter RHEL 10 only?
filtering is done in the CustomPut implementation


=> 2. Advisory overwrite risk

No, by design. The code explicitly prevents this. Here's why:
The CustomPut does a read-merge-write cycle, not a blind overwrite:

The flow is:
1. getAdvisoryFromDB reads the existing OVAL data from advisory-detail → {cveID} → Red Hat → {pkgName}. This returns the full advisory struct including all RHEL 7/8/9 entries that OVAL already wrote.

2. mergeAdvisories copies all existing OVAL entries into a new struct, then appends CSAF entries. If a CSAF entry matches an existing entry on FixedVersion + Arches, it unions the CPE indices rather than duplicating:

3. PutAdvisoryDetail then writes the merged result. Yes, it's a full overwrite at the key level (bkt.Put([]byte(key), v)), but the value being written already contains the OVAL data because it was read and merged in step 1-2.
This is the same pattern used by debian.CustomPut() and ubuntu.CustomPut() in aqua-db -- they all do read-merge-write via CustomPut.

we have added a safeguard at premium side to always run oval first then csaf.


=> 3. Separate DB or separate buckets instead of merging

This would work but introduces more complexity than it removes, and doesn't align with the OSS design direction.
Here's why the current merged approach was chosen:

1. The OSS trivy-db already designed CSAF as a drop-in replacement for OVAL using the same buckets.
Both OVAL and CSAF in OSS write to the same "Red Hat" bucket and share the same CPEList type, PutAdvisoryDetail path, and advisory-detail intermediate bucket. The WithRunAlongsideOVAL option was specifically created in OSS to support the transition period where both run together. Separate buckets would diverge from OSS and create a parallel data path that doesn't exist upstream.

2. The scanner's Get() only queries one bucket per vendor.
The scanner queries "Red Hat" → {pkgName} → {cveID} to get advisories. If CSAF data were in "Red Hat CSAF", the scanner would need to query both "Red Hat" and "Red Hat CSAF" and merge the results at scan time. This means:
Every Red Hat scan call would double its DB lookups
The merge logic moves from build time (once) to scan time (every scan)
CPE index resolution would need to handle two separate CPE namespaces

3. CPE indices are inherently shared.
The scanner resolves repo-to-CPE mappings into CPE indices, then matches those indices against advisory entries' Affected arrays. If CSAF had its own CPE bucket, the same CPE string would have different indices in OVAL vs CSAF. The scanner would need to know which index namespace to use for each advisory, adding complexity at the critical path.

4. The "overwrite risk" is already mitigated by the read-merge-write pattern.
As shown above, CustomPut reads existing data before writing. This is a well-established pattern in aqua-db (Debian and Ubuntu use it too). The data integrity is guaranteed by BoltDB's serializable transactions.

5. Migration path is cleaner with shared buckets.
When OVAL is eventually deprecated, removing it is a simple deletion of the OVAL tracker from vulnsrc.go. The CSAF data is already in the right buckets. With separate buckets, deprecating OVAL would require a migration step to move data from "Red Hat CSAF" to "Red Hat", or permanent scanner changes to query both.

For reference: 
PR from premium side
https://bitbucket.org/scalock/aqua-db/pull-requests/594/overview

knqyf263:
The advisory merge logic lives inside `CustomPut`, but the CPE merge logic (`mergeCPEList`) is added directly to the OSS codebase. This inconsistency makes the code harder to maintain — some premium-specific behavior is injected via `CustomPut`, while other premium-specific behavior is hardcoded with a `runAlongsideOVAL` flag.

What if we made both behaviors customizable through a single abstraction? I've created #648 which introduces a `Store` interface with two methods — `PutMappings` (for CPE handling) and `Put` (for advisory writes) — so that premium can implement both without adding premium-specific logic to OSS.

Could you take a look and see if it meets the requirements?

PR #648, opened 2026-03-16
refactor(redhat-csaf): replace CustomPut with Store interface

knqyf263:
## Summary
- Replace the untyped `db.CustomPut` function with a typed `Store` interface in the CSAF package
- The `Store` interface has two methods: `PutMappings` (handles CPE mapping writes and returns the CPE list) and `Put` (handles advisory writes)
- OSS uses a `defaultStore` implementation; premium can inject a custom `Store` via `WithStore()`

### Background
When running CSAF alongside OVAL (e.g. RHEL 10 from CSAF, RHEL 7–9 from OVAL), the premium side needs to customize both advisory writes and CPE mapping behavior. The previous `CustomPut` allowed overriding advisory writes, but CPE handling (mapping writes, CPE list construction) was not customizable. This PR introduces a `Store` interface that covers both, so the premium side can implement the full merge logic.

### Out of scope (separate PR)
- `ReleaseDate` field on `PutInput` and parser changes (`ReleaseDate()`, `FormatDate()`)

## Test plan
- [x] All existing tests pass (no behavioral change for OSS)
- [x] Added `TestVulnSrc_Update_WithCustomStore` integration test that verifies a custom Store implementation receives correct calls and can customize both CPE and advisory handling

PR #649, opened 2026-03-18
feat: parse and format csaf Publish Date with unit tests

gshanbhag525:
parse and format csaf Publish Date with unit tests

Custom Store implementations (e.g. premium) can now use input.ReleaseDate in their Put implementation (e.g. for PublishDate). The default OSS store does not use it but receives it in PutInput.

addition to #648

### DIFF
--- a/pkg/vulnsrc/redhat-csaf/csaf.go
+++ b/pkg/vulnsrc/redhat-csaf/csaf.go
@@ -52,9 +52,10 @@ type MappingsInput struct {
 }
 
 type PutInput struct {
-	Bucket   Bucket
-	Advisory Advisory
-	CPEList  redhatoval.CPEList
+	Bucket      Bucket
+	Advisory    Advisory
+	CPEList     redhatoval.CPEList
+	ReleaseDate string // Advisory initial release date (YYYY-MM-DD), used by CustomPut for PublishDate
 }
 
 // defaultStore is the OSS default implementation of Store.
@@ -196,9 +197,10 @@ func (vs VulnSrc) update(tx *bolt.Tx, dir string) error {
 		advisory := Advisory{Entries: entries}
 
 		input := &PutInput{
-			Bucket:   bkt,
-			Advisory: advisory,
-			CPEList:  cpeList,
+			Bucket:      bkt,
+			Advisory:    advisory,
+			CPEList:     cpeList,
+			ReleaseDate: vs.parser.ReleaseDate(bkt.VulnerabilityID),
 		}
 		if err := vs.store.Put(vs.dbc, tx, input); err != nil {
 			return eb.Wrapf(err, "failed to put advisory")
--- a/pkg/vulnsrc/redhat-csaf/csaf.go
+++ b/pkg/vulnsrc/redhat-csaf/csaf.go
@@ -53,9 +53,9 @@ type MappingsInput struct {
 }
 
 type PutInput struct {
-	Bucket      Bucket
-	Advisory    Advisory
-	CPEList     redhatoval.CPEList
+	Bucket   Bucket
+	Advisory Advisory
+	CPEList  redhatoval.CPEList
 	// ReleaseDate is the RHSA vendor_fix remediation instant from CSAF (RFC3339). Zero means
 	// unknown or not applicable (e.g. CVE-only rows). Custom Store implementations may use
 	// it for PublishDate (calendar day: t.UTC().Format(time.DateOnly)).
--- a/pkg/vulnsrc/redhat-csaf/csaf.go
+++ b/pkg/vulnsrc/redhat-csaf/csaf.go
@@ -56,9 +56,9 @@ type PutInput struct {
 	Bucket   Bucket
 	Advisory Advisory
 	CPEList  redhatoval.CPEList
-	// ReleaseDate is the RHSA vendor_fix remediation instant from CSAF (RFC3339). Zero means
-	// unknown or not applicable (e.g. CVE-only rows). Custom Store implementations may use
-	// it for PublishDate (calendar day: t.UTC().Format(time.DateOnly)).
+	// ReleaseDate is the vendor_fix remediation timestamp from CSAF. Zero when there is
+	// no RHSA (CVE-only rows) or when the feed omits Remediation.Date. Intended for
+	// Store implementations that want to expose an advisory publish date.
 	ReleaseDate time.Time
 }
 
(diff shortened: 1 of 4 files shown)


gshanbhag525:
@DmitriyLewen  and @knqyf263 
please review

pkg/vulnsrc/redhat-csaf/parse.go
```diff
@@ -303,6 +306,11 @@ func (p *Parser) parseVulnerability(adv CSAFAdvisory, vuln *csaf.Vulnerability)
 			if status == types.StatusFixed {
 				vulnID = p.extractRHSAID(remediation)
 				alias = cveID
+				// Store the advisory release date for this RHSA ID (once per vulnID)
+				if _, exists := p.releaseDates[vulnID]; !exists &&
+					adv.Document != nil && adv.Document.Tracking != nil && adv.Document.Tracking.InitialReleaseDate != nil {
+					p.releaseDates[vulnID] = p.formatDate(*adv.Document.Tracking.InitialReleaseDate)
+				}
```
DmitriyLewen: I have two questions:
1. You save dates for fixed advisories only
2. You save adv.Document.Tracking.InitialReleaseDate instead of RHSA date  (Remediation.Date).

Is this expected logic?

gshanbhag525: > You save dates for fixed advisories only

yes, 
we also rely on CVE Vendor Publish Date, right now these are provided by oval since same cves are present in oval already. For redhat 10 unfixed cves, i see only 7 cves since the count is small, i have not included another parameter here.

> You save adv.Document.Tracking.InitialReleaseDate instead of RHSA date (Remediation.Date).

this I changed To Remediation date

pkg/vulnsrc/redhat-csaf/parse.go
```diff
@@ -432,6 +443,20 @@ func (p *Parser) NVRToCPE() iter.Seq2[string, []string] {
 	return maps.All(p.nvrToCPE)
 }
 
+// ReleaseDate returns the advisory release date (YYYY-MM-DD) for a given vulnerability ID.
+func (p *Parser) ReleaseDate(vulnID VulnerabilityID) string {
+	return p.releaseDates[vulnID]
+}
+
+// formatDate extracts the date portion (YYYY-MM-DD) from an RFC3339 timestamp.
+func (p *Parser) formatDate(timestamp string) string {
```
DmitriyLewen, line 452: We don't use parser for this function.
So we can create formatDate as package function:

```suggestion
func formatDate(timestamp string) string {
```

pkg/vulnsrc/redhat-csaf/parse.go
```diff
@@ -432,6 +443,20 @@ func (p *Parser) NVRToCPE() iter.Seq2[string, []string] {
...
+	t, err := time.Parse(time.RFC3339, timestamp)
+	if err != nil {
+		return ""
```
DmitriyLewen, line 455: Should we handle this case?
If I understand correctly — the error would mean that RedHat changed the date format.

If we add a log — we might still miss the error, as we rarely check logs without errors.
Maybe it makes sense to return an error so that we can release a fix right away?

pkg/vulnsrc/redhat-csaf/parse.go
```diff
@@ -303,6 +308,12 @@ func (p *Parser) parseVulnerability(adv CSAFAdvisory, vuln *csaf.Vulnerability)
 			if status == types.StatusFixed {
 				vulnID = p.extractRHSAID(remediation)
 				alias = cveID
+				// Store the RHSA release date (once per vulnID) from Remediation.Date.
+				if _, exists := p.releaseDates[vulnID]; !exists {
+					if remediation != nil && remediation.Date != nil {
```
DmitriyLewen, line 313: We already checked this here:
https://github.com/aquasecurity/trivy-db/blob/3e8ae373a31809426f2b3d3c1e4e8bf34a9cbbd3/pkg/vulnsrc/redhat-csaf/parse.go#L216

```suggestion
					if remediation.Date != nil {
```

DmitriyLewen (commented):
left comments.

Also I think we need to test ReleaseDate in `TestVulnSrc_Update_WithCustomStore`
you can add check here:
https://github.com/aquasecurity/trivy-db/blob/32625aa310272a1ff6904a68b78dddc43ac33872/pkg/vulnsrc/redhat-csaf/csaf_test.go#L318-L323

But re-check that emoty store.putInputs.ReleaseDate is correct.

gshanbhag525:
hi @knqyf263 can you take a look
thanks

pkg/vulnsrc/redhat-csaf/csaf.go
```diff
@@ -52,9 +52,10 @@ type MappingsInput struct {
 }
 
 type PutInput struct {
-	Bucket   Bucket
-	Advisory Advisory
-	CPEList  redhatoval.CPEList
+	Bucket      Bucket
+	Advisory    Advisory
+	CPEList     redhatoval.CPEList
+	ReleaseDate string // Advisory initial release date (YYYY-MM-DD), used by CustomPut for PublishDate
```
DmitriyLewen: I was just thinking — wouldn't you want to use time.Time here?
You can get only the date on your side, but in the future, if the time is needed — there will be no breaking changes.
This would be consistent with other fields [PublishedDate](https://github.com/aquasecurity/trivy-db/blob/e96046ef17aa64588b0be79a359bdc292a375d81/pkg/types/types.go#L84) and [Published](https://github.com/aquasecurity/trivy-db/blob/722777bbcf5e245b59d3f3e2545bcafedd3ca26c/pkg/vulnsrc/osv/osv.go#L49).

pkg/vulnsrc/redhat-csaf/parse.go
```diff
@@ -303,6 +308,16 @@ func (p *Parser) parseVulnerability(adv CSAFAdvisory, vuln *csaf.Vulnerability)
 			if status == types.StatusFixed {
 				vulnID = p.extractRHSAID(remediation)
 				alias = cveID
+				// Store the RHSA release date (once per vulnID) from Remediation.Date.
+				if _, exists := p.releaseDates[vulnID]; !exists {
+					if remediation.Date != nil {
+						formatted, err := formatDate(*remediation.Date)
+						if err != nil {
+							return oops.With("cve_id", cveID).With("vulnerability_id", vulnID).Wrapf(err, "remediation date")
```
DmitriyLewen: Should we return error here?
If one advisory has an error in the date — we will stop the entire build. Perhaps we should skip this date instead?
wdyt?

DmitriyLewen:
Hello @gshanbhag525 
I created small refactoring.
LMK if something is wrong.

DmitriyLewen (approved):

PR #654, opened 2026-03-27
feat(vulnsrc): add RapidFort Security Advisories source

vaibhav-rf:
## Summary

Adds `rapidfort` as a new vulnerability source in `trivy-db`, ingesting the advisories published by RapidFort in [rapidfort/security-advisories](https://github.com/rapidfort/security-advisories).

## How the data is fetched

- `Makefile` (`db-fetch-langs`) downloads the repo tarball directly into the build cache — the same pattern as `ghsa` / `ruby-advisory-db` / `bitnami-vulndb`, not through `vuln-list-update` (the data is already published as parseable JSON).
- Layout: `rapidfort-security-advisories/OS/{os}/{package}.json`, schema `{package_name, advisory: {distroVersion -> cveID -> CVEEntry}}`, where `Event = {introduced?, fixed?, identifier?}`.

### `pkg/ecosystem/ecosystem.go`
- New `RapidFort Type = "rapidfort"` ecosystem constant.

### `pkg/vulnsrc/vulnerability/const.go`
- `RapidFort` SourceID for the new source.

### `pkg/vulnsrc/bucket/bucket.go`
- New `NewFedora(version)` constructor — a Fedora bucket for the fedora ranges below.

### `pkg/vulnsrc/rapidfort/`
- **`types.go`** — `SourcePackageAdvisory`, `CVEEntry`, `Event` matching the feed format.
- **`bucket.go`** — `rapidFortBucket` wraps a base-OS bucket; the platform name is `"rapidfort {base} {version}"` (e.g. `"rapidfort ubuntu 22.04"`, `"rapidfort Red Hat 9"`), and RapidFort's own rebuilds use a bare `"rapidfort"` (no base distribution). `newBucket` is the single source of truth for which base ecosystems are supported (kept aligned with the companion detector in `trivy`).
- **`rapidfort.go`** — `VulnSrc` (DB builder) and `VulnSrcGetter` (scanner query path):
  - Walks the per-package JSON files with `utils.FileWalk`; derives the base OS from the directory path.
  - Splits by distro version in-memory (one file bundles every version of a package).
  - **RedHat:** a single file mixes ranges from different distributions, each tagged by a per-range identifier (`elN`/`fcNN`/`rf`). Every range is routed into its own single-distro bucket: `elN` → `rapidfort Red Hat N`, `fcNN` → `rapidfort fedora NN`, `rf` → the bare `rapidfort`. Ranges the feed replicates identically under every RHEL major are deduplicated.
  - Severity is stored per-package in `Advisory` (not in `VulnerabilityDetail`), so `FillInfo` can't override RapidFort's curated severity with the base-OS VendorSeverity.
  - An empty parse fails the build loudly (guards against silently shipping an empty integration).
- **`rapidfort_test.go`** — `TestVulnSrc_Update` / `TestVulnSrc_Get` covering Ubuntu, Alpine, Red Hat, Fedora and the bare `rapidfort` bucket; version-split; dedup; unsupported-OS skip; empty parse; invalid JSON; missing advisory.

### `pkg/vulnsrc/vulnsrc.go`
- `rapidfort.NewVulnSrc()` registered in the global source list.

### `Makefile`
- Fetches the RapidFort advisory repo into the build cache.

## Supported OSes
- Ubuntu
- Alpine
- Red Hat (RHEL) — its file additionally yields Fedora (`rapidfort fedora NN`) and RapidFort-rebuild (`rapidfort`) buckets.

Other OSes present in the feed (e.g. Debian) are skipped.

## Related
- Companion scanner PR: aquasecurity/trivy#10452 (should be merged in sync — without it the RapidFort data has no consumer).

## Testing

```bash
GOEXPERIMENT=jsonv2 go test ./pkg/vulnsrc/rapidfort/...
go build ./...
```


knqyf263:
Hi there, can you please contact us using this form https://trivy.dev/partners?

vaibhav-rf:
Hello @knqyf263 

The requested form has been completed and submitted by the team. Please let us know if there's anything else needed from our end. 

Thank you!

knqyf263:
@vaibhav-rf Thanks, we have confirmed the form was received. Our partnerships team reached out by email a few weeks ago and offered a meeting, but we have not heard back from RapidFort's side on that thread.

Could you check internally and make sure the right contact at RapidFort follows up on that email so the partnership conversation can move forward? The PR review proceeds once that discussion is in place.

pkg/vulnsrc/rapidfort/rapidfort.go
```diff
@@ -0,0 +1,261 @@
...
+func (vs VulnSrc) parse(rootDir string) ([]entry, error) {
+	eb := oops.In("rapidfort").With("root_dir", rootDir)
+	var entries []entry
+
+	err := utils.FileWalk(rootDir, func(r io.Reader, path string) error {
+		if !strings.HasSuffix(path, ".json") {
+			return nil
+		}
+
+		// Relative path: {osName}/{version}/{pkg}.json
```
DmitriyLewen: take a look https://github.com/aquasecurity/vuln-list-update/pull/424#pullrequestreview-4475439605

pkg/vulnsrc/rapidfort/rapidfort.go
```diff
@@ -0,0 +1,261 @@
...
+
+	severity := types.SeverityUnknown
+	if sev, err := types.NewSeverity(strings.ToUpper(cve.Severity)); err == nil {
+		severity = sev
+	}
+
+	adv := types.Advisory{
+		PatchedVersions:    patched,
+		VulnerableVersions: vulnerable,
+		Severity:           severity,
```
DmitriyLewen: You write the severity into both buckets (Advisory and VulnerabilityDetail).
But we use only one bucket:
We keep the severity in Advisory when the source has different severity for a single CVE for different packages (e.g. Debian).
For other cases use VulnerabilityDetail.

pkg/vulnsrc/rapidfort/rapidfort.go
```diff
@@ -0,0 +1,261 @@
...
+			}
+		}
+
+		for _, e := range entries {
+			eb := oops.With("platform", e.platform).With("package", e.pkgName).With("cve", e.cveID)
+
+			if err := vs.dbc.PutAdvisoryDetail(tx, e.cveID, e.pkgName, []string{e.platform}, e.advisory); err != nil {
+				return eb.Wrapf(err, "failed to save advisory")
+			}
+			if err := vs.dbc.PutVulnerabilityDetail(tx, e.cveID, source.ID, e.detail); err != nil {
```
DmitriyLewen, line 279: You don't check this in tests.

pkg/vulnsrc/rapidfort/rapidfort.go
DmitriyLewen: also make sense add list of supported OSes

pkg/vulnsrc/rapidfort/rapidfort.go
```diff
@@ -0,0 +1,261 @@
...
+			// Open vulnerability (no fix): write without space for the same reason.
+			vulnerable = append(vulnerable, fmt.Sprintf(">=%s", ev.Introduced))
+		default:
+			continue
+		}
+		// Track identifiers parallel to vulnerable versions.
+		identifiers = append(identifiers, ev.Identifier)
+		if ev.Identifier != "" {
+			hasIdentifier = true
+		}
```
DmitriyLewen: unneeded code. 
you can use `lo.EveryBy(ids, func(s string) bool { return s == "" }) ` code.
Will be readable.

pkg/vulnsrc/bucket/bucket.go
```diff
@@ -110,6 +110,30 @@ func NewUbuntu(version string) Bucket { return newOS(ecosystem.Ubuntu, version)
...
+
+func (r rapidFortBucket) BaseOS() string {
+	return r.baseOS
+}
+
+// NewRapidFort creates a bucket for RapidFort advisories for a given base OS and version.
+func NewRapidFort(baseOS, version string) rapidFortBucket {
+	return rapidFortBucket{baseOS: baseOS, version: version}
+}
+
```
DmitriyLewen: It is better to use the default OS bucket to keep the comparison between buckets consistent.
Create a bucket.go file for the rapidfort package, something like this:
```go
// rapidFortBucket wraps a base OS bucket. RapidFort advisories reuse the base OS
// naming (e.g. "ubuntu 20.04", "Red Hat 9") prefixed with "rapidfort",
// so the platform name stays consistent with the original distributions.
type rapidFortBucket struct {
	base       bucket.Bucket
	dataSource types.DataSource
}

func (r rapidFortBucket) Name() string {
	return "rapidfort " + r.base.Name()
}

func (r rapidFortBucket) Ecosystem() ecosystem.Type {
	return ecosystem.RapidFort
}

func (r rapidFortBucket) DataSource() types.DataSource {
	return r.dataSource
}

// newBucket resolves the base OS directory name (e.g. "ubuntu", "debian", "redhat")
// into a RapidFort bucket, including the DataSource (with BaseID) used when saving.
// An unsupported base OS results in an error so the caller can skip it.
func newBucket(baseOS, version string) (bucket.DataSourceBucket, error) {
	ds := source // copy of the package-level DataSource (ID, Name, URL)
	var base bucket.Bucket
	switch baseOS {
	case "ubuntu":
		base, ds.BaseID = bucket.NewUbuntu(version), vulnerability.Ubuntu
	case "debian":
		base, ds.BaseID = bucket.NewDebian(version), vulnerability.Debian
	case "redhat":
		base, ds.BaseID = bucket.NewRedHat(version), vulnerability.RedHat
	default:
		return nil, oops.With("base_os", baseOS).Errorf("unsupported base OS")
	}
	return rapidFortBucket{base: base, dataSource: ds}, nil
}
```

DmitriyLewen: Also, it will simplify the entry struct:
```diff
diff --git a/pkg/vulnsrc/rapidfort/rapidfort.go b/pkg/vulnsrc/rapidfort/rapidfort.go
index 302876f..1ff8244 100644
--- a/pkg/vulnsrc/rapidfort/rapidfort.go
+++ b/pkg/vulnsrc/rapidfort/rapidfort.go
@@ -66,8 +66,7 @@ func (vs VulnSrc) Update(dir string) error {
 }
 
 type entry struct {
-	platform string
-	baseOS   string
+	bucket   bucket.DataSourceBucket
 	pkgName  string
 	cveID    string
 	advisory types.Advisory
@@ -98,16 +97,20 @@ func (vs VulnSrc) parse(rootDir string) ([]entry, error) {
 		osName := parts[0]
 		version := parts[1]
 
+		b, err := newBucket(osName, version)
+		if err != nil {
+			vs.logger.Warn("Skipping unsupported base OS", "path", path, "base_os", osName)
+			return nil
+		}
+
 		var pkg PackageAdvisory
 		if err := json.NewDecoder(r).Decode(&pkg); err != nil {
 			return eb.With("path", path).Wrapf(err, "json decode error")
 		}
 
-		b := bucket.NewRapidFort(osName, version)
 		for cveID, cveEntry := range pkg.Advisories {
 			entries = append(entries, entry{
-				platform: b.Name(),
-				baseOS:   b.BaseOS(),
+				bucket:   b,
 				pkgName:  pkg.PackageName,
 				cveID:    cveID,
 				advisory: buildAdvisory(cveEntry),
@@ -129,29 +132,21 @@ func (vs VulnSrc) put(entries []entry) error {
 	}
 	vs.logger.Info("Saving RapidFort advisories", "count", len(entries))
 
-	// Track unique platform → baseOS mappings for DataSource registration.
-	platforms := map[string]string{}
-	for _, e := range entries {
-		platforms[e.platform] = e.baseOS
-	}
-
 	return vs.dbc.BatchUpdate(func(tx *bolt.Tx) error {
-		for platform, baseOS := range platforms {
-			ds := types.DataSource{
-				ID:     source.ID,
-				Name:   source.Name,
-				URL:    source.URL,
-				BaseID: types.SourceID(baseOS),
-			}
-			if err := vs.dbc.PutDataSource(tx, platform, ds); err != nil {
-				return oops.With("platform", platform).Wrapf(err, "failed to put data source")
-			}
-		}
-
+		addedDataSources := map[string]struct{}{}
 		for _, e := range entries {
-			eb := oops.With("platform", e.platform).With("package", e.pkgName).With("cve", e.cveID)
+			platform := e.bucket.Name()
+			eb := oops.With("platform", platform).With("package", e.pkgName).With("cve", e.cveID)
+
+			// Register the data source once per platform.
+			if _, ok := addedDataSources[platform]; !ok {
+				if err := vs.dbc.PutDataSource(tx, platform, e.bucket.DataSource()); err != nil {
+					return eb.Wrapf(err, "failed to put data source")
+				}
+				addedDataSources[platform] = struct{}{}
+			}
 
-			if err := vs.dbc.PutAdvisoryDetail(tx, e.cveID, e.pkgName, []string{e.platform}, e.advisory); err != nil {
+			if err := vs.dbc.PutAdvisoryDetail(tx, e.cveID, e.pkgName, []string{platform}, e.advisory); err != nil {
 				return eb.Wrapf(err, "failed to save advisory")
 			}
 			if err := vs.dbc.PutVulnerabilityDetail(tx, e.cveID, source.ID, e.detail); err != nil {
@@ -252,8 +247,11 @@ func NewVulnSrcGetter(baseOS string) VulnSrcGetter {
 func (vs VulnSrcGetter) Get(params db.GetParams) ([]types.Advisory, error) {
 	eb := oops.In("rapidfort").With("base_os", vs.baseOS).With("os_version", params.Release).With("package_name", params.PkgName)
 
-	platformName := bucket.NewRapidFort(vs.baseOS, params.Release).Name()
-	advs, err := vs.dbc.GetAdvisories(platformName, params.PkgName)
+	b, err := newBucket(vs.baseOS, params.Release)
+	if err != nil {
+		return nil, eb.Wrapf(err, "failed to create a bucket name")
+	}
+	advs, err := vs.dbc.GetAdvisories(b.Name(), params.PkgName)
 	if err != nil {
 		return nil, eb.Wrapf(err, "failed to get advisories")
 	}


```

pkg/vulnsrc/bucket/bucket.go
```diff
@@ -110,6 +110,30 @@ func NewUbuntu(version string) Bucket { return newOS(ecosystem.Ubuntu, version)
 // NewWolfi creates a bucket for Wolfi
 func NewWolfi(version string) Bucket { return newOS(ecosystem.Wolfi, version) }
 
+// rapidFortBucket for RapidFort advisories which span multiple base OSes.
+// The platform name encodes both the base OS and the version, e.g. "rapidfort ubuntu 22.04".
+type rapidFortBucket struct {
+	baseOS  string // e.g. "ubuntu", "debian", "redhat"
```
DmitriyLewen: What about Alpine advisories?
[Trivy PR](https://github.com/aquasecurity/trivy/pull/10452) contains Alpine.

pkg/vulnsrc/rapidfort/rapidfort.go
```diff
@@ -0,0 +1,261 @@
...
+		},
+	}
+}
+
+// Get returns RapidFort advisories for a given package and OS version (e.g. "22.04").
+func (vs VulnSrcGetter) Get(params db.GetParams) ([]types.Advisory, error) {
+	eb := oops.In("rapidfort").With("base_os", vs.baseOS).With("os_version", params.Release).With("package_name", params.PkgName)
+
+	platformName := bucket.NewRapidFort(vs.baseOS, params.Release).Name()
+	advs, err := vs.dbc.GetAdvisories(platformName, params.PkgName)
```
DmitriyLewen: When I was looking into the identifiers, I noticed that you also have advisories for the original packages (packages from vendors). For example, [this advisory](https://github.com/rapidfort/security-advisories/blob/2f1e7ef711ba2621680bce04725f24b30af9576b/OS/redhat/curl_advisory.json#L10860-L10874) contains fixed versions for both rh5 and fc19.

But I have to clarify: your database contains advisories for both your own packages and vendor packages, so you don't need a fallback inside trivy-db for vendor packages, right?

pkg/vulnsrc/rapidfort/rapidfort_test.go
```diff
@@ -0,0 +1,325 @@
...
+			},
+		},
+		{
+			name: "empty dir - no advisories",
+			dir:  filepath.Join("testdata", "empty"),
+			noBuckets: [][]string{
+				{"advisory-detail"},
+				{"vulnerability-id"},
+				{"vulnerability-detail"},
+			},
```
DmitriyLewen, line 434: I think we can remove this testcase.

pkg/vulnsrc/rapidfort/testdata/fixtures/broken.yaml
DmitriyLewen, line 1: Unused fixtures file

pkg/vulnsrc/bucket/bucket.go
rfJanit: It will support alpine as well.  I will update the comment line.

pkg/vulnsrc/rapidfort/rapidfort.go
rfJanit: Yes, we provide data for both our packages and vendor packages, and thus we dont need to fall back on trivy-db for vendor packages.

DmitriyLewen:
I found a strange case:
https://github.com/rapidfort/security-advisories/blob/ea90c8c4d04533250065c8bd180f1de41a3a24d7/OS/redhat/nodejs20-full-i18n_advisory.json#L13-L14
Affected range: `>= 1:20.19.5-2.fc43 < 20.20.0-2.fc43`.
This is an impossible case — epoch 1 is always greater than epoch 0.

pkg/vulnsrc/rapidfort/rapidfort.go
rfJanit: Addressed.

pkg/vulnsrc/rapidfort/rapidfort.go
rfJanit: Addressed.

pkg/vulnsrc/rapidfort/rapidfort.go
rfJanit: Addressed.

pkg/vulnsrc/rapidfort/rapidfort.go
rfJanit: Addressed.

pkg/vulnsrc/bucket/bucket.go
rfJanit: Addressed.

pkg/vulnsrc/rapidfort/rapidfort_test.go
rfJanit: Addressed.

pkg/vulnsrc/rapidfort/testdata/fixtures/broken.yaml
rfJanit: Addressed.

rfJanit:
> I found a strange case: https://github.com/rapidfort/security-advisories/blob/ea90c8c4d04533250065c8bd180f1de41a3a24d7/OS/redhat/nodejs20-full-i18n_advisory.json#L13-L14 Affected range: `>= 1:20.19.5-2.fc43 < 20.20.0-2.fc43`. This is an impossible case — epoch 1 is always greater than epoch 0.

This case has been addressed. Thanks for letting us know.

@DmitriyLewen Kindly review the PR again. All your review comments have been addressed. Thanks!

pkg/vulnsrc/rapidfort/rapidfort.go
```diff
@@ -0,0 +1,285 @@
...
+		// dedup here avoids a pre-pass over entries just to build a platform set.
+		addedDataSources := map[string]struct{}{}
+		for _, e := range entries {
+			// After the bucket refactor, entry no longer carries a pre-computed
+			// platform string — it holds a DataSourceBucket that composes the
+			// name from the base OS + version at call time (e.g. "rapidfort
+			// Red Hat 9"). Name() does a string concat internally, so cache
+			// it once here and reuse it four times below: the oops error
+			// context, the addedDataSources dedup key, PutAdvisoryDetail's
+			// nestedBktNames arg, and (indirectly, as the platform) PutDataSource.
```
DmitriyLewen: I think the comment about the refactoring is unnecessary here.

pkg/vulnsrc/rapidfort/rapidfort.go
```diff
@@ -0,0 +1,285 @@
...
+		return nil, oops.Wrapf(err, "walk error")
+	}
+	return entries, nil
+}
+
+func (vs VulnSrc) put(entries []entry) error {
+	if len(entries) == 0 {
+		vs.logger.Info("No RapidFort advisories found")
+		return nil
+	}
```
DmitriyLewen, line 258: Should we return the error here?
This is an unexpected case, and we'll silently skip it (we don't check all the logs for each build run).

pkg/vulnsrc/rapidfort/rapidfort.go
```diff
@@ -0,0 +1,285 @@
...
+}
+
+// buildVulnerabilityDetail constructs a VulnerabilityDetail from a CVEEntry for
+// enriching the vulnerability bucket with title, description and severity.
+// buildVulnerabilityDetail carries only prose metadata (title, description).
+// Severity is intentionally NOT set here — RapidFort is a curated derivative
+// (like root.io), so per-package severity lives in Advisory and is read
+// directly by the scanner. Writing it here as well would risk FillInfo
+// overriding RapidFort's curated severity with the base-OS severity via
+// VendorSeverity[BaseID] when the same CVE also exists in the base feed.
```
DmitriyLewen: There's a contradiction between lines 241 and 242 (severity). 
On top of that, I think we can simplify this whole comment.

pkg/vulnsrc/rapidfort/rapidfort.go
```diff
@@ -0,0 +1,285 @@
...
+
+		var src SourcePackageAdvisory
+		if err := json.NewDecoder(r).Decode(&src); err != nil {
+			return eb.With("path", path).Wrapf(err, "json decode error")
+		}
+
+		// The source file's shape is {version: {cveID: CVEEntry}}. We fan it
+		// out here into per-(platform, cveID) entries so downstream put() can
+		// write one advisory-detail per (version, package, cveID) tuple —
+		// the same shape the previous vuln-list-update-based parser produced.
```
DmitriyLewen: IIUC we can remove this line.

pkg/vulnsrc/rapidfort/rapidfort.go
```diff
@@ -0,0 +1,285 @@
...
+func (vs VulnSrc) put(entries []entry) error {
+	if len(entries) == 0 {
+		vs.logger.Info("No RapidFort advisories found")
+		return nil
+	}
+	vs.logger.Info("Saving RapidFort advisories", "count", len(entries))
+
+	return vs.dbc.BatchUpdate(func(tx *bolt.Tx) error {
+		// Register the data source once per unique platform. Inlining the
+		// dedup here avoids a pre-pass over entries just to build a platform set.
```
DmitriyLewen:
```suggestion
		// Register the data source once per platform.
```

pkg/vulnsrc/rapidfort/rapidfort.go
```diff
@@ -0,0 +1,285 @@
...
+// VendorSeverity[BaseID] when the same CVE also exists in the base feed.
+func buildVulnerabilityDetail(cve CVEEntry) types.VulnerabilityDetail {
+	return types.VulnerabilityDetail{
+		Title:       cve.Title,
+		Description: cve.Description,
+	}
+}
+
+// VulnSrcGetter is used by trivy (the scanner) to query advisories from the DB
+// for a specific base OS (e.g. "ubuntu" or "debian").
```
DmitriyLewen: Debian is not supported now, so it's better to switch from Debian to another OS.

pkg/vulnsrc/rapidfort/rapidfort.go
```diff
@@ -0,0 +1,285 @@
...
+	"github.com/aquasecurity/trivy-db/pkg/utils"
+	"github.com/aquasecurity/trivy-db/pkg/vulnsrc/bucket"
+	"github.com/aquasecurity/trivy-db/pkg/vulnsrc/vulnerability"
+)
+
+// rapidfortDir is the subdirectory under the DB-build cache where the RapidFort
+// security-advisories git repo is extracted. It must match the directory name
+// used by trivy-db's Makefile db-fetch-langs target (see the download_and_extract
+// line for github.com/rapidfort/security-advisories). The "rapidfort-" prefix
+// disambiguates it from other "*-security-advisories" caches (e.g. php-).
```
DmitriyLewen: I think you can remove this.

pkg/vulnsrc/rapidfort/rapidfort.go
```diff
@@ -0,0 +1,285 @@
...
+		var src SourcePackageAdvisory
+		if err := json.NewDecoder(r).Decode(&src); err != nil {
+			return eb.With("path", path).Wrapf(err, "json decode error")
+		}
+
+		// The source file's shape is {version: {cveID: CVEEntry}}. We fan it
+		// out here into per-(platform, cveID) entries so downstream put() can
+		// write one advisory-detail per (version, package, cveID) tuple —
+		// the same shape the previous vuln-list-update-based parser produced.
+		for version, cveMap := range src.Advisory {
```
DmitriyLewen: I think we should add validation for the `version` keys.
E.g. I found a case with an `el4` version (not supported by Trivy) — https://github.com/rapidfort/security-advisories/blob/666e85cb41b187ec1934ed623142c4e90e69759f/OS/redhat/p7zip-plugins_advisory.json#L19-L33

pkg/ecosystem/ecosystem.go
```diff
@@ -25,27 +25,27 @@ const (
 	Kubernetes Type = "k8s"
 
 	// OS ecosystems
-	Bottlerocket Type = "bottlerocket"
```
DmitriyLewen, line 28: It looks like this line was removed by mistake.

pkg/vulnsrc/rapidfort/rapidfort.go
```diff
@@ -0,0 +1,285 @@
...
+				vulnerable = append(vulnerable, fmt.Sprintf("<%s", ev.Fixed))
+			}
+		case ev.Introduced != "":
+			// Open vulnerability (no fix): write without space for the same reason.
+			vulnerable = append(vulnerable, fmt.Sprintf(">=%s", ev.Introduced))
+		default:
+			continue
+		}
+		// Track identifiers parallel to vulnerable versions.
+		identifiers = append(identifiers, ev.Identifier)
```
DmitriyLewen: I saw that you filled empty identifiers from the feed, but should we add a check empty `ev.Identifier` here?

rfJanit: I kept it the same on purpose. The identifiers list and the VulnerableVersions list need to have the same length, because the scanner matches them by position — item 0 in one list goes with item 0 in the other, item 1 goes with item 1, and so on.   

If we skip empty identifiers, the two lists get different lengths, and the positions stop matching correctly. 

In real data, empty identifiers only happen for ubuntu and alpine — and in those cases, all identifiers are empty. The EveryBy check right below catches this and skips saving Custom completely. So empty strings never actually get saved to the DB.

DmitriyLewen: I should have been a bit more detailed in this comment. I meant a check for emptiness and returning an error right away.
The logic with `lo.EveryBy` does indeed skip adding the identifiers if there are no empty values in the slice. But let's think about whether this is correct.

If we don't add an identifier for such cases, we'll get both FPs (which may be acceptable) and, if I understand correctly, FNs — which could be critical. 
On the other hand, users won't be able to get new advisories until the error is fixed.

Since the case where an advisory has an empty identifier shouldn't occur in practice, I would still treat it as an error and stop the database update. 
However, if you're confident in the current approach, we can leave it as is and accept the risks of FPs and FNs.

pkg/ecosystem/ecosystem.go
rfJanit: Addressed.

pkg/vulnsrc/rapidfort/rapidfort.go
rfJanit: Thanks for informing about this case. We will add a data side check for this instead of adding it here.

pkg/vulnsrc/rapidfort/rapidfort.go
rfJanit: Addressed.

pkg/vulnsrc/rapidfort/rapidfort.go
rfJanit: Addressed.

pkg/vulnsrc/rapidfort/rapidfort.go
rfJanit: Addressed.

pkg/vulnsrc/rapidfort/rapidfort.go
rfJanit: Addressed.

pkg/vulnsrc/rapidfort/rapidfort.go
rfJanit: Addressed.

pkg/vulnsrc/rapidfort/rapidfort.go
rfJanit: Yes, this is a better approach. Addressed.

pkg/vulnsrc/rapidfort/rapidfort.go
rfJanit: Addressed.

pkg/vulnsrc/rapidfort/rapidfort.go
rfJanit: Went with the fail-loud approach. `buildAdvisory` now errors out when a single CVE's events mix empty and non-empty identifiers — allEmpty (ubuntu/alpine) and noneEmpty (redhat) both pass, only the misalignment case is rejected. 

Added test cases as well for the same.

DmitriyLewen:
Hello @vaibhav-rf @rfJanit 
I refactored the RedHat part of this PR. In short, what changed:

- A RedHat feed file bundles ranges for several distributions in one file, tagged by a per-range identifier (`elN`, `fcNN`, `rf`). Each range is now routed into its own bucket by identifier: `elN` → `rapidfort Red Hat N`, `fcNN` → `rapidfort fedora NN`, `rf` → `rapidfort` (RapidFort's own rebuilds, which have no base distribution).
- Removed the `Custom.Identifiers` mechanism (and the `RapidFortCustom` type): previously ranges from different distributions shared one redhat bucket and were told apart by a parallel identifier array indexed alongside them — now each range lives in its own single-distro bucket, so the scanner no longer needs to tell them apart by identifier.
- Deduplicated the `fcNN`/`rf` ranges that the feed repeats identically under every RHEL major — they are now stored once.
- Updated the tests and testdata (added a multiversion fixture for the dedup, removed `mixed_identifiers`).

Could you double-check that I didn't miss anything or break the logic somewhere?

One more thing: this refactor splits the redhat feed into separate buckets (`rapidfort Red Hat N` / `rapidfort fedora NN` / bare `rapidfort`) and drops `Custom.identifiers`, so the companion Trivy PR ([#10452](https://github.com/aquasecurity/trivy/pull/10452)) needs work on the RedHat/Fedora/rf side.
Can you do this?

pkg/vulnsrc/rapidfort/rapidfort.go
```diff
@@ -0,0 +1,371 @@
...
+			return eb.With("path", path).Wrapf(err, "json decode error")
+		}
+
+		// ubuntu/alpine feeds already hold one distribution per file. The RedHat
+		// feed mixes RHEL/Fedora/rf ranges, so split it into one advisory set per
+		// distribution first; both shapes then convert the same way.
+		sources := map[ecosystem.Type]SourcePackageAdvisory{
+			osName: src,
+		}
+		if osName == ecosystem.RedHat {
```
knqyf263: The Ubuntu feed can contain both `rf` and `ubuntu` ranges in the same CVE entry. For example, [CVE-2025-69648 for rf-binutils](https://github.com/rapidfort/security-advisories/blob/56e5868f921b0b3a116cf60837cd8154de634c19/OS/ubuntu/rf-binutils_advisory.json#L5-L46) is fixed at `0:2.46-1ubuntu1` for Ubuntu, while the `rf` range remains vulnerable below `0:2.46-10rfubu`. With the current ingestion, an Ubuntu package at `0:2.46-3ubuntu2` is newer than the Ubuntu fix, but it still satisfies the `rf` range because `0:2.46-3ubuntu2 < 0:2.46-10rfubu` under Debian version comparison. Trivy would therefore report this CVE even though the Ubuntu package is fixed. Since the identifier is discarded here, should these ranges be separated?

DmitriyLewen: Thanks for catching this.

When I did the refactor, the `identifier` field existed only in the RedHat feed — it showed up in the Ubuntu feed while the work was already in progress.

Because of these changes we can no longer route `rf` ranges into the bare `rapidfort` bucket.
Previously `rf` ranges came only from the RedHat feed; now the same package carries `rf` ranges in both feeds with incompatible version schemes.
For example, `rf-curl` / CVE-2026-8932 is fixed at `0:8.21.0-10rfubu` in the Ubuntu feed and at `8.21.0-10.rf` in the RedHat feed — in a single bare `rapidfort` bucket these two collide (nine packages overlap like this today).

One option is to key the bucket by OS but not by version for RapidFort's own packages, and keep the version for upstream ranges: `rapidfort <os>` for `rf` ranges and `rapidfort <os> <version>` for upstream ones.
For example, `rapidfort ubuntu` and `rapidfort ubuntu 22.04`.

@vaibhav-rf @rfJanit — a question on this: do identifiers bound to a specific OS version exist today, or are they planned (analogous to `el8` for RHEL — something like `rfel8` / `rfubu2204`)?
Whether the version-less `rapidfort <os>` bucket works depends on that.

This will also require changes in the [Trivy PR](https://github.com/aquasecurity/trivy/pull/10452) — we need a way to tell upstream packages apart from RapidFort ones.
For RPM packages we currently look at the dist tag in the version (`.el` / `.fc` / `.rf`); as far as I can tell, deb packages need a similar marker.

rfJanit: Thank you, @knqyf263 and @DmitriyLewen , for raising this concern and proposing a solution.

@knqyf263 : Yes, the ranges should be separated in a manner similar to the existing implementation for Red Hat, following the approach Dmitriy outlined in the comment below.

@DmitriyLewen : Your concern regarding placing all rf ranges within a single RapidFort bucket is valid, and segregation is therefore appropriate. Your proposed approach appears suitable, and I have implemented it accordingly. The existing rapidfort bucket has been renamed to rapidfort redhat, and a new rapidfort ubuntu bucket has been added to accommodate the RapidFort version ranges for Ubuntu.

Regarding your question, there are currently no plans of that nature. However, we will keep you informed should any such plans arise in the future.

Thank you once again for your valuable feedback.

pkg/vulnsrc/rapidfort/bucket.go
```diff
@@ -12,24 +12,32 @@ import (
...
+// still split into RedHat- and Ubuntu-flavored buckets so their ranges never
+// share a bucket across package formats — that would let the wrong version
+// comparator interpret a range string.
 type rapidFortBucket struct {
+	ecosystem  ecosystem.Type
 	base       bucket.Bucket
 	dataSource types.DataSource
+	// rfSuffix identifies the rf bucket flavor when base is nil ("redhat" or
+	// "ubuntu"). Empty for base-backed buckets (Ubuntu/Alpine/RedHat/Fedora).
+	rfSuffix string
```
DmitriyLewen: We try to reuse the existing buckets. In your case, you're creating additional names that will differ from the upstream names — for example, "redhat" and "Red Hat".

Instead of adding rfSuffix, I suggest simply dropping the version for the rt identifier.

e.g.:
https://github.com/DmitriyLewen/trivy-db/commit/6ba1405b37400e26ec2bf5d3ccd5debbb045db7f#diff-1dbbca0b8082b4afec1fc77a34e50e7096bd40966fe410a8d0a4315bf85a6156

- For "rapidfort Red Hat 9" — newBucket(ecosystem.RedHat, "9")
- For "rapidfort Red Hat" — newBucket(ecosystem.RedHat, "")

pkg/vulnsrc/rapidfort/rapidfort.go
```diff
@@ -209,13 +213,85 @@ func (vs VulnSrc) splitRedHat(src SourcePackageAdvisory, path string) map[ecosys
 	return out
 }
 
+// splitUbuntu re-keys a mixed Ubuntu feed file into one advisory set per
+// distribution it targets — "ubuntu" and "rf" — keyed by that distribution's
+// own version. rf ranges land in the distribution-less "rapidfort" bucket
+// alongside RedHat's rf events; ubuntu ranges stay in "rapidfort ubuntu <ver>".
+// A missing identifier is treated as "ubuntu" for backward compatibility with
+// pre-annotation feed files.
+func (vs VulnSrc) splitUbuntu(src SourcePackageAdvisory, path string) map[ecosystem.Type]SourcePackageAdvisory {
```
DmitriyLewen: splitUbuntu contains a lot of duplicate code here. 
I suggest making a single common function for all OSes. 

Something like this — https://github.com/DmitriyLewen/trivy-db/commit/6ba1405b37400e26ec2bf5d3ccd5debbb045db7f#diff-fbd5dc7d1463ae1de299be7528d685da1fcb73cb3f45248f0f2bf65990fc0196

pkg/vulnsrc/rapidfort/bucket.go
rfJanit: Thanks for the idea. Addressed.

pkg/vulnsrc/rapidfort/rapidfort.go
rfJanit: Have implemented the code on the similar pattern as the attached commit.

Addressed.

rfJanit:
@DmitriyLewen Kindly review. Thanks!

knqyf263 (approved):

DmitriyLewen (approved):

PR #657, opened 2026-04-15
feat(rootio): migrate source to per-advisory OSV v1.6.0 format

chait-slim:
Migrates the Root.io vulnerability source from the legacy monolithic
  cve_feed.json to per-advisory OSV v1.6.0 files under rootio/**/*.json,
  using the shared pkg/vulnsrc/osv parser.

  Each advisory's bucket is resolved by a single "Root:" bucket resolver
  driven by the OSV ecosystem string (e.g. Root:Alpine:3.18) plus
  database_specific.distro / distro_version. Only OS ecosystems are
  supported.

  Trivy resolution is unchanged: VulnSrcGetter.Get pulls every base-OS
  advisory (Debian/Ubuntu/Alpine), then overlays Root.io's patched
  versions on top, with Root.io winning only on vulnerability-ID
  collision. Unfixed CVEs continue to surface from the upstream source.


pkg/vulnsrc/rootio/types.go
DmitriyLewen, line 1: i think you can use [osv packages](https://github.com/aquasecurity/trivy-db/tree/main/pkg/vulnsrc/osv).

You can see `seal` as example

chait-slim: fixed

chait-slim:
@DmitriyLewen can you please review again the two PRs?

pkg/vulnsrc/rootio/rootio.go
```diff
@@ -4,181 +4,125 @@ import (
...
+func (vs VulnSrc) Update(root string) error {
+	eb := oops.In("rootio").With("root", root)
 
-	feedFile, err := os.Open(feedFilePath)
-	if err != nil {
-		return eb.Wrapf(err, "failed to open feed file %s", feedFilePath)
+	dataSources := map[ecosystem.Type]types.DataSource{
+		ecosystem.Npm:      source,
+		ecosystem.Pip:      source,
+		ecosystem.RubyGems: source,
```
DmitriyLewen: We migrate to OSV in this PR.

We don't need to add language ecosystems here.

pkg/vulnsrc/rootio/rootio.go
```diff
@@ -4,181 +4,125 @@ import (
...
 	}
-	defer feedFile.Close()
 
-	var rawFeed RawFeed
-	if err = json.NewDecoder(feedFile).Decode(&rawFeed); err != nil {
-		return eb.With("file_path", feedFilePath).Wrapf(err, "json decode error")
+	o := osv.New(vulnsDir, source.ID, dataSources,
+		osv.WithBucketResolver("alpine", resolveAlpineBucket),
+		osv.WithBucketResolver("debian", resolveDebianBucket),
+		osv.WithBucketResolver("ubuntu", resolveUbuntuBucket),
```
DmitriyLewen: IIUC we can create a single function to resolve the Bucket based on the ecosystem.

It also makes sense to move this function into bucket.go.

pkg/vulnsrc/rootio/rootio.go
```diff
@@ -4,181 +4,125 @@ import (
...
-				return err
-			}
+// TransformAdvisories applies Root.io's distro/distro_version (carried in
+// entry.DatabaseSpecific) to OS advisories and drops entries that have no
+// fixed version.
+func (t *transformer) TransformAdvisories(advs []osv.Advisory, entry osv.Entry) ([]osv.Advisory, error) {
+	var dbSpec struct {
+		Distro        string `json:"distro"`
+		DistroVersion string `json:"distro_version"`
+	}
```
DmitriyLewen: I would move this structure outside of the function — it's more readable that way.

pkg/vulnsrc/rootio/rootio.go
```diff
@@ -4,181 +4,125 @@ import (
...
+		if len(adv.PatchedVersions) == 0 {
+			continue
 		}
-	}
-	return nil
-}
 
-func (vs VulnSrc) put(tx *bolt.Tx, platform string, feed Feed) error {
-	eb := oops.With("platform", platform).With("package", feed.PkgName).With("cve", feed.VulnerabilityID)
+		if dbSpec.Distro != "" {
```
DmitriyLewen: Since we currently only support OS ecosystems — I think we should return an error if Distro is empty.

pkg/vulnsrc/rootio/rootio.go
```diff
@@ -4,181 +4,125 @@ import (
...
+		}
+		out = append(out, adv)
 	}
+	return out, nil
+}
 
-	return nil
+type config struct {
+	dbc    db.Operation
+	logger *log.Logger
```
DmitriyLewen: unused now.

pkg/vulnsrc/rootio/rootio.go
DmitriyLewen: can we use `bucket` now?

pkg/vulnsrc/rootio/testdata/happy/vuln-list/rootio/rootio-curl/ROOT-OS-ALPINE-318-CVE-2023-38473.json
DmitriyLewen, line 1: can you update test files with severity?

chait-slim:
@DmitriyLewen addressed all comments

pkg/vulnsrc/rootio/rootio_test.go
```diff
@@ -72,64 +40,41 @@ func TestVulnSrc_Update(t *testing.T) {
...
+						"CVE-2023-38473",
 					},
 					Value: map[string]any{},
 				},
 			},
-		},
-		{
-			name: "happy path with unsupported OS",
-			dir:  filepath.Join("testdata", "unsupported-os"),
+			// Skipped entries (empty affected, no fixed) must not create buckets
```
DmitriyLewen: Could you elaborate on why you are skipping these CVEs (both with an empty affected and without a fixed version)?
Is this an internal root.io decision?
I assume that an empty affected means it is not yet known whether the package is vulnerable (correct me if I'm wrong), but I think we should still show vulnerabilities that don't have a fix yet, even if it's a potential vulnerability — users should be aware.

DmitriyLewen (commented):
Can you update PR description?

Also i left one question.

pkg/vulnsrc/rootio/rootio_test.go
chait-slim: You're right that filtering these is questionable. The Root.io feed only publishes advisories where Root.io has shipped a patch: a "no fix" or "empty affected" entry shouldn't occur in real feed data, so the filter was just defensive code against malformed entries.
Removed it (and the test fixtures that only existed to exercise it).

DmitriyLewen:
Hi @chait-slim 
I updated the PR a bit:
- restored the comments
- replaced sourceID with ecosystem (this will simplify the code and be consistent with other packages). For these changes I have already prepared a PR for Trivy —github.com/aquasecurity/trivy/pull/10629

Could you take a quick look?
Also, please review and fix the comments in https://github.com/aquasecurity/vuln-list-update/pull/438.

Thanks in advance!

idoberko2:
Hey @DmitriyLewen! Sorry for the delay, we were recently had some changes in the company that lead to some changes in priority. We're now getting back to this, I'll be helping @chait-slim to bring this to the finish line

idoberko2:
@DmitriyLewen, I merged in the recent `main`. I believe there are currently no open comments on this one, merely waiting for https://github.com/aquasecurity/vuln-list-update/pull/438 to be merged. Correct?

PR #659, opened 2026-04-19
feat(echo): add echo language package advisories

orizerah:
Add support for downloading Echo's OSV advisory.

Issue: https://github.com/aquasecurity/trivy/discussions/10519

This PR relies on [this](https://github.com/aquasecurity/vuln-list-update/pull/436) PR to be merged first:

### DIFF
--- a/pkg/vulnsrc/echo-osv/echo_osv.go
+++ b/pkg/vulnsrc/echo-osv/echo_osv.go
@@ -1,0 +1,64 @@
+package echoosv
+
+import (
+	"path/filepath"
+	"strings"
+
+	"github.com/samber/oops"
+
+	"github.com/aquasecurity/trivy-db/pkg/ecosystem"
+	"github.com/aquasecurity/trivy-db/pkg/types"
+	"github.com/aquasecurity/trivy-db/pkg/vulnsrc/osv"
+	"github.com/aquasecurity/trivy-db/pkg/vulnsrc/vulnerability"
+)
+
+var (
+	vulnsDir = filepath.Join("vuln-list", "echo-osv")
+
+	source = types.DataSource{
+		ID:   vulnerability.EchoOSV,
+		Name: "Echo OSV",
+		URL:  "https://advisory.echohq.com/osv",
+	}
+
+	dataSources = map[ecosystem.Type]types.DataSource{
+		ecosystem.Pip: source,
+	}
+)
+
+type VulnSrc struct{}
+
+func NewVulnSrc() VulnSrc {
+	return VulnSrc{}
+}
+
+func (VulnSrc) Name() types.SourceID {
+	return source.ID
+}
+
+func (VulnSrc) Update(root string) error {
+	o := osv.New(vulnsDir, source.ID, dataSources, osv.WithTransformer(&transformer{}))
+	if err := o.Update(root); err != nil {
+		return oops.In("echo-osv").Wrapf(err, "failed to update Echo OSV vulnerability data")
+	}
+	return nil
+}
+
+// transformer filters out advisories that didn't resolve to a CVE ID.
+// The Echo OSV feed contains both ECHO-ID and CVE-ID entries for the same
+// vulnerability; keeping only CVE-keyed advisories avoids duplicates.
+type transformer struct{}
+
+func (t *transformer) PostParseAffected(adv osv.Advisory, _ osv.Affected) (osv.Advisory, error) {
+	return adv, nil
+}
+
+func (t *transformer) TransformAdvisories(advisories []osv.Advisory, _ osv.Entry) ([]osv.Advisory, error) {
+	var filtered []osv.Advisory
+	for _, adv := range advisories {
+		if strings.HasPrefix(adv.VulnerabilityID, "CVE-") {
+			filtered = append(filtered, adv)
+		}
+	}
+	return filtered, nil
+}
--- a/pkg/vulnsrc/echo-osv/echo_osv.go
+++ b/pkg/vulnsrc/echo-osv/echo_osv.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/aquasecurity/trivy-db/pkg/ecosystem"
 	"github.com/aquasecurity/trivy-db/pkg/types"
+	"github.com/aquasecurity/trivy-db/pkg/vulnsrc/bucket"
 	"github.com/aquasecurity/trivy-db/pkg/vulnsrc/osv"
 	"github.com/aquasecurity/trivy-db/pkg/vulnsrc/vulnerability"
 )
@@ -19,10 +20,6 @@ var (
 		ID:   vulnerability.EchoOSV,
 		Name: "Echo OSV",
 		URL:  "https://advisory.echohq.com/osv",
-	}
-
-	dataSources = map[ecosystem.Type]types.DataSource{
-		ecosystem.Pip: source,
 	}
 )
 
@@ -37,11 +34,21 @@ func (VulnSrc) Name() types.SourceID {
 }
 
 func (VulnSrc) Update(root string) error {
-	o := osv.New(vulnsDir, source.ID, dataSources, osv.WithTransformer(&transformer{}))
+	dataSources := map[ecosystem.Type]types.DataSource{
+		ecosystem.Pip: source,
+	}
+	o := osv.New(vulnsDir, source.ID, dataSources,
+		osv.WithTransformer(&transformer{}),
+		osv.WithBucketResolver("pypi", resolvePyPI),
+	)
 	if err := o.Update(root); err != nil {
 		return oops.In("echo-osv").Wrapf(err, "failed to update Echo OSV vulnerability data")
 	}
 	return nil
+}
+
+func resolvePyPI(_ string) (bucket.Bucket, error) {
+	return newPipBucket(source)
 }
 
 // transformer filters out advisories that didn't resolve to a CVE ID.
--- a/pkg/vulnsrc/echo-osv/echo_osv.go
+++ b/pkg/vulnsrc/echo-osv/echo_osv.go
@@ -39,7 +39,7 @@ func (VulnSrc) Update(root string) error {
 	}
 	o := osv.New(vulnsDir, source.ID, dataSources,
 		osv.WithTransformer(&transformer{}),
-		osv.WithBucketResolver("pypi", resolvePyPI),
+		osv.WithBucketResolver("echo", resolveEcho),
 	)
 	if err := o.Update(root); err != nil {
 		return oops.In("echo-osv").Wrapf(err, "failed to update Echo OSV vulnerability data")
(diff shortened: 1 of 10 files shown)


pkg/vulnsrc/echo-osv/echo_osv.go
```diff
@@ -0,0 +1,83 @@
...
+
+	"github.com/aquasecurity/trivy-db/pkg/ecosystem"
+	"github.com/aquasecurity/trivy-db/pkg/types"
+	"github.com/aquasecurity/trivy-db/pkg/vulnsrc/bucket"
+	"github.com/aquasecurity/trivy-db/pkg/vulnsrc/osv"
+	"github.com/aquasecurity/trivy-db/pkg/vulnsrc/vulnerability"
+)
+
+var (
+	vulnsDir = filepath.Join("vuln-list", "echo-osv")
```
DmitriyLewen, line 17: related comment - https://github.com/aquasecurity/vuln-list-update/pull/436#discussion_r3265221716

pkg/vulnsrc/osv/osv.go
```diff
@@ -29,6 +29,7 @@ type Advisory struct {
 	PkgName         string
 	VulnerabilityID string
 	Aliases         []string
+	Upstream        []string
```
DmitriyLewen, line 32: Unused variable.

pkg/vulnsrc/echo-osv/testdata/happy/vuln-list/echo-osv/ECHO-2024-1234.json
```diff
@@ -0,0 +1,49 @@
...
+  "severity": [
+    {
+      "type": "CVSS_V3",
+      "score": "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:N/A:N"
+    }
+  ],
+  "affected": [
+    {
+      "package": {
+        "ecosystem": "Echo:PyPi",
```
DmitriyLewen, line 20: https://github.com/aquasecurity/vuln-list-update/pull/436#discussion_r3871544074

pkg/vulnsrc/echo-osv/echo_osv.go
```diff
@@ -0,0 +1,83 @@
...
+type transformer struct{}
+
+func (t *transformer) PostParseAffected(adv osv.Advisory, _ osv.Affected) (osv.Advisory, error) {
+	return adv, nil
+}
+
+func (t *transformer) TransformAdvisories(advisories []osv.Advisory, _ osv.Entry) ([]osv.Advisory, error) {
+	var filtered []osv.Advisory
+	for _, adv := range advisories {
+		if strings.HasPrefix(adv.VulnerabilityID, "CVE-") {
```
DmitriyLewen, line 78: I found a case with an advisory that has a GHSA-id only — ECHO-c457-5e3c-26ac.
Can you recheck this?

pkg/vulnsrc/echo-osv/echo_osv.go
```diff
@@ -0,0 +1,83 @@
...
+// plus plain "Echo" for OS packages. resolveBucket lowercases and splits on ':',
+// so this is invoked with eco="echo" and the original suffix.
+func resolveEcho(suffix string) (bucket.Bucket, error) {
+	switch suffix {
+	case "pypi":
+		return newPipBucket(source)
+	default:
+		// Plain "Echo" entries are OS packages and are handled by the
+		// existing `echo` source from vuln-list/echo/. Skip them here so
+		// we don't double-store them under a language bucket.
```
DmitriyLewen, line 61: nit: We filter OS package advisories in vuln-list-update, so this comment is a bit confusing.

pkg/vulnsrc/echo-osv/echo_osv.go
orizerah: This is addressed on the other pr

pkg/vulnsrc/echo-osv/testdata/happy/vuln-list/echo-osv/ECHO-2024-1234.json
orizerah: Fixed

pkg/vulnsrc/echo-osv/echo_osv.go
orizerah: You are correct, sometimes there's no CVE id

pkg/vulnsrc/echo-osv/echo_osv.go
orizerah: I decided to add the filter because I don't want to rely on a filter existing in the vuln list update.

pkg/vulnsrc/echo-osv/echo_osv.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+func (VulnSrc) Name() types.SourceID {
+	return source.ID
+}
+
+func (VulnSrc) Update(root string) error {
+	dataSources := map[ecosystem.Type]types.DataSource{
+		ecosystem.Pip: source,
+	}
```
DmitriyLewen, line 40: Since all Echo ecosystems go through `resolveEcho`, we can pass `nil` instead of `dataSources` here.
This rules out Echo advisories ending up in `pip::Echo OSV` if the feed has an error (e.g. an entry with the plain `PyPI` ecosystem).

pkg/vulnsrc/echo-osv/testdata/happy/vuln-list/echo-osv/ECHO-2024-1234.json
```diff
@@ -0,0 +1,49 @@
+{
+  "schema_version": "1.4.0",
+  "id": "ECHO-2024-1234",
+  "modified": "2024-10-15T12:00:00Z",
+  "published": "2024-09-01T08:00:00Z",
+  "upstream": [
+    "CVE-2024-99999"
+  ],
```
DmitriyLewen, line 8: The most common real shape isn't covered: in the current feed almost all language advisories have `upstream: [CVE, GHSA]` (after a build from the real feed, 498 of 528 advisories have `VendorIDs: ["GHSA-…", "ECHO-…"]`), while the fixtures have only a CVE-only or GHSA-only `upstream`.
So, for example, a change that re-keys by GHSA even when a CVE is present wouldn't be caught by the tests.
Could you extend one of the existing happy fixtures (e.g. `ECHO-2024-1234.json`) with a GHSA in `upstream` and check the CVE key, `VendorIDs`, and `noBuckets` for the GHSA key?

It would also be good to bring the fixtures closer to the real feed format: it has no CVSS `severity` (the array is always empty), no `published`, `details` or `schema_version`, and `summary` is present only in withdrawn entries.

pkg/vulnsrc/echo-osv/echo_osv.go
```diff
@@ -0,0 +1,97 @@
...
+	switch suffix {
+	case "pypi":
+		return newPipBucket(source)
+	default:
+		// Only PyPI is wired up for now; the OSV parser skips ecosystems we
+		// don't resolve. OS package advisories (plain "Echo") are filtered
+		// out in vuln-list-update, so they normally never reach this point.
+		return nil, oops.Errorf("unsupported Echo ecosystem suffix: %q", suffix)
+	}
+}
```
DmitriyLewen, line 65: nit:
- The resolver is selected by the `echo` key and receives only the suffix — it has no `eco` argument.
- Plain `Echo` does reach this function (with an empty suffix) and is skipped here by the `default` branch, so trivy-db doesn't depend on the vuln-list-update filter — exactly what you wanted.

Suggestion:

```go
// resolveEcho maps an Echo OSV ecosystem to a bucket.
// resolveBucket lowercases the ecosystem and splits it on ':', so "Echo:PyPI"
// arrives here as suffix "pypi" and plain "Echo" (OS packages) as "".
func resolveEcho(suffix string) (bucket.Bucket, error) {
	switch suffix {
	case "pypi":
		return newPipBucket(source)
	default:
		// Only PyPI is supported for now. Plain "Echo" (OS packages, served by
		// the `echo` source) and other suffixes are skipped by the OSV parser.
		return nil, oops.Errorf("unsupported Echo ecosystem suffix: %q", suffix)
	}
}
```

pkg/vulnsrc/echo-osv/echo_osv.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/aquasecurity/trivy-db/pkg/vulnsrc/vulnerability"
+)
+
+var (
+	vulnsDir = filepath.Join("vuln-list", "echo-osv")
+
+	source = types.DataSource{
+		ID:   vulnerability.EchoOSV,
+		Name: "Echo OSV",
+		URL:  "https://advisory.echohq.com/osv",
```
DmitriyLewen, line 23: nit:
`https://advisory.echohq.com/osv` (as well as `https://advisory.echohq.com/`) returns 403; only `https://advisory.echohq.com/osv/all.zip` returns 200.
This URL is shown in Trivy output.
Is there a public page for the OSV feed? If not, `.../osv/all.zip` could be used.

pkg/vulnsrc/osv/osv.go
DmitriyLewen: The field is used now, but it's still not needed in the shared `osv.Advisory`: `TransformAdvisories` already receives `entry`, and `adv.Upstream` is always equal to `entry.Upstream`.
Could you revert the changes in `osv/osv.go` and use `entry.Upstream` in the transformer?
Since the decision depends only on the entry, the GHSA can be looked up once before the loop:

```go
func (t *transformer) TransformAdvisories(advisories []osv.Advisory, entry osv.Entry) ([]osv.Advisory, error) {
	ghsaID, hasGHSA := lo.Find(entry.Upstream, func(id string) bool {
		return strings.HasPrefix(id, "GHSA-")
	})

	var filtered []osv.Advisory
	for _, adv := range advisories {
		if !strings.HasPrefix(adv.VulnerabilityID, "CVE-") {
			if !hasGHSA {
				continue
			}
			// Key the advisory by the GHSA and keep the ECHO ID as a vendor ID.
			adv.VulnerabilityID = ghsaID
			adv.Aliases = lo.Uniq(append(lo.Without(adv.Aliases, ghsaID), entry.ID))
		}
		filtered = append(filtered, adv)
	}
	return filtered, nil
}
```

DmitriyLewen (commented):
Some `Echo:PyPI` ranges contain versions that aren't PEP 440 at all.
Of 560 `fixed` values in non-withdrawn `Echo:PyPI` entries, 113 look like Debian package versions (e.g. `2.2.2-3+deb12u1` for `Werkzeug` in `ECHO-b9b6-c39c-7946`, `3.7.4-1+deb11u1` for `aiohttp` in `ECHO-8549-82cb-62d1`), 24 of which have an epoch (`3:3.2.25-0+deb12u1` for `django` in `ECHO-91c5-85b0-1e52`).
There is also a Go pseudo-version: `0.0.0-20231218163308-9d2ee975ef9f` for `paramiko` in `ECHO-7d6b-5585-901d`.

Trivy compares echo pip versions using PEP 440, so constraints with an epoch fail to parse (`improper constraint: <3:3.2.25-0+deb12u1`) and never match, while the others describe the Debian package rather than the PyPI one.
Is this expected?
It looks like OS package data leaked into `Echo:PyPI` records — could this be fixed on your side?

---

nit: duplicated version ranges

32 entries list the same package in different case (`jinja2`/`Jinja2`, `werkzeug`/`Werkzeug`).
After name normalization both end up in one advisory and the ranges are appended twice, e.g. for `CVE-2024-22195` / `jinja2`: `["<2.11.3-1+deb11u1","<2.11.3-1+deb11u1"]`.
It doesn't affect matching, but it's worth either removing the duplicates in the feed or deduplicating versions (e.g. `lo.Uniq` in `osv.parseAffected`).

DmitriyLewen (approved):
LGTM

PR #664, opened 2026-05-07
feat(ubuntu): add missing stream release mappings to Ubuntu

Bit-Doctor:
## Description

This PR extends `UbuntuReleasesMapping` with missing Ubuntu release **stream** keys found in [aquasecurity/vuln-list](https://github.com/aquasecurity/vuln-list) CVE (for example `fips-updates/jammy`, `fips-preview/jammy`, and additional ESM stream entries).

Without these stream mappings, advisories keyed only under those release/stream names are skipped in `defaultPut`.

## References

- https://github.com/aquasecurity/vuln-list/blob/main/ubuntu/2026/CVE-2026-23099.json
- https://github.com/aquasecurity/vuln-list/blob/main/ubuntu/2026/CVE-2026-7737.json
- https://github.com/aquasecurity/vuln-list/blob/main/ubuntu/2026/CVE-2026-7598.json

## Request for review

Would a small normalization helper (parse codename segments; classify `esm-apps/`, `esm-infra/`, `esm-infra-legacy/` as `*-ESM`; map `fips*` streams to the base numeric suite) be preferable for long-term scalability, keeping the map only for base release?


pkg/vulnsrc/ubuntu/ubuntu.go
```diff
@@ -57,12 +57,29 @@ var (
...
+		"esm-apps/bionic":         "18.04-ESM",
+		"esm-infra/focal":         "20.04-ESM",
+		"esm-apps/focal":          "20.04-ESM",
+		"esm-infra/jammy":         "22.04-ESM",
+		"esm-apps/jammy":          "22.04-ESM",
+		"esm-infra/noble":         "24.04-ESM",
+		"esm-apps/noble":          "24.04-ESM",
+		"esm-infra/resolute":      "26.04-ESM",
+		"esm-apps/resolute":       "26.04-ESM",
+		// FIPS
```
DmitriyLewen: I think FIPS should not be merged with the regular release buckets.

The way `defaultPut` works in `pkg/vulnsrc/ubuntu/ubuntu.go`, it iterates `cve.Patches[pkg]` and calls `PutAdvisoryDetail(tx, cve, pkg, [platform], adv)` keyed on `(cve, pkg, platform)`. With this PR, both `jammy` and `fips-updates/jammy` map to the same `ubuntu 22.04` bucket, so when a CVE has `released` for both streams (e.g. CVE-2024-5535, CVE-2023-3446, CVE-2024-0727 for `openssl`), the two writes collide and the second one wins. Map iteration order in Go is randomized, so the stored `FixedVersion` becomes non-deterministic across DB builds — sometimes the regular fix `3.0.2-0ubuntu1.X`, sometimes the FIPS fix `3.0.2-0ubuntu1.17+Fips1`.

Even if we sorted the iteration deterministically, there is no FIPS-package detection on Ubuntu in Trivy itself (the Ubuntu scanner at `trivy/pkg/detector/ospkg/ubuntu/ubuntu.go` just queries by `pkg.SrcName` + `osVer` — compare with Oracle, which has `PackageFlavor` / `FipsPackageFlavor` in `pkg/vulnsrc/oracle-oval/oracle-oval.go`). So we would either show a FIPS-validated fixed version (`+Fips1`) to a regular openssl user, or a non-FIPS fix to a FIPS user — both wrong.

Properly supporting FIPS would require (a) a separate bucket like `22.04-FIPS` and (b) a corresponding change in Trivy to identify FIPS packages by version suffix (`+Fips`, `+fips`, `.fips.`). That is out of scope here. For this PR I'd suggest dropping the `fips/`, `fips-updates/`, `fips-preview/` entries and keeping the ESM additions only. We can track FIPS support as a separate issue.

pkg/vulnsrc/ubuntu/ubuntu.go
```diff
@@ -57,12 +57,29 @@ var (
...
+		"esm-infra/bionic":        "18.04-ESM",
+		"esm-apps/bionic":         "18.04-ESM",
+		"esm-infra/focal":         "20.04-ESM",
+		"esm-apps/focal":          "20.04-ESM",
+		"esm-infra/jammy":         "22.04-ESM",
+		"esm-apps/jammy":          "22.04-ESM",
+		"esm-infra/noble":         "24.04-ESM",
+		"esm-apps/noble":          "24.04-ESM",
+		"esm-infra/resolute":      "26.04-ESM",
+		"esm-apps/resolute":       "26.04-ESM",
```
DmitriyLewen: I don't think we need to add these names right now. Worth explaining why, because how Trivy queries this data is not obvious from the trivy-db side.

The Trivy Ubuntu detector (`trivy/pkg/detector/ospkg/ubuntu/ubuntu.go`, `versionFromEolDates`) decides which bucket to query based on EOL dates:

```go
func (s *Scanner) versionFromEolDates(ctx context.Context, osVer string) string {
    if _, ok := s.eolDates[osVer]; ok {
        return osVer
    }
    // if base version (not ESM) is still actual, use it
    ver := strings.TrimRight(osVer, "-ESM")
    if eol, ok := s.eolDates[ver]; ok && clock.Now(ctx).Before(eol) {
        return ver
    }
    return osVer
}
```

So while the base release is still under standard support, Trivy strips `-ESM` and queries the plain `22.04` / `24.04` bucket. The `*-ESM` bucket only starts being queried after standard EOL. Current EOL dates from the same file: jammy — 2027-04-23, noble — 2034-03-31, resolute — not in `eolDates` yet.

pkg/vulnsrc/ubuntu/ubuntu.go
Bit-Doctor: The src package for the fips stream from what I saw also had a change in the name and not only in the version.
ex: `linux-aws-fips` vs `linux-aws`. I assume that because of this there shouldn't be any override? 

But maybe they are example where the package would be the same and would create a clash silently indeed.
Checking if I can find one.

pkg/vulnsrc/ubuntu/ubuntu.go
Bit-Doctor: Would the current change just create buckets on the DB that would not be used until trivy consider that release as EOL?

pkg/vulnsrc/ubuntu/ubuntu.go
Bit-Doctor: Ok got a counter exemple from my assumption 👍 
https://github.com/aquasecurity/vuln-list/blob/7676ae4f172c5b86016b9877a320bd2fe5353263/ubuntu/2025/CVE-2025-9230.json#L143-L150
On the same file there's both `openssl-fips` and `openssl` but both have remediation for fips stream

pkg/vulnsrc/ubuntu/ubuntu.go
DmitriyLewen: right.
But the main problem is not the "bloating" of trivy-db (or unused buckets), but the fact that by the time a release reaches EOL, Ubuntu may make changes.
As you can see — Ubuntu periodically changes the key for a release in their advisories.
Therefore it is better to add keys when Ubuntu actually starts publishing advisories for them, to avoid the case where we expect to support some ESM release but Ubuntu has changed the key and we are not showing advisories.
I also hope that over time we will migrate to OSV, at which point these keys will no longer be needed.

pkg/vulnsrc/ubuntu/release_mapping_test.go
DmitriyLewen, line 1: I think this test is not needed anymore.

PR #668, opened 2026-05-19
fix(ubuntu): include "pending" status in vulnerability tracking

michaelr-orca:
## Summary

- CVEs with Ubuntu status `"pending"` (fix committed upstream but not yet released as an update) were silently dropped because `"pending"` was missing from the `targetStatuses` list in `pkg/vulnsrc/ubuntu/ubuntu.go`
- This meant Trivy users were not alerted to vulnerabilities that have known fixes in progress
- Added `"pending"` to `targetStatuses` alongside the existing `"needed"`, `"deferred"`, and `"released"` values

## References

Ubuntu CVE Tracker status definitions: https://wiki.ubuntu.com/SecurityTeam/FAQ#CVE_Statuses

- **needed** — fix is needed but not yet available
- **deferred** — fix is deferred to a later release
- **pending** — fix is committed upstream but not yet released as an update
- **released** — fix has been released

## Test plan

- [x] Added `"pending"` status entry to existing test fixture (`CVE-2020-1234.json`)
- [x] Added new test case `"pending status is included"` verifying the advisory is stored
- [x] All existing tests continue to pass

### DIFF
--- a/pkg/vulnsrc/ubuntu/testdata/vuln-list/ubuntu/CVE-2020-1234.json
+++ b/pkg/vulnsrc/ubuntu/testdata/vuln-list/ubuntu/CVE-2020-1234.json
@@ -23,6 +23,10 @@
       "focal": {
         "Status": "needs-triage",
         "Note": ""
+      },
+      "jammy": {
+        "Status": "pending",
+        "Note": "1.2.4"
       }
     }
   },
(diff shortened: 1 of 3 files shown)


DmitriyLewen (approved):
Thanks!

Issue #650, opened 2026-03-19
Inconsistent severity priority: `Severity` field uses CVSS-first, `VendorSeverity` uses advisory-first


DmitriyLewen:
## Description

`Normalize()` computes `Severity` and `VendorSeverity` using **opposite priority rules**
for the same source data, leading to inconsistent results.

### `getSeverity()` — populates top-level `Severity` field

```go
case d.CvssScoreV3 > 0:  return scoreToSeverity(d.CvssScoreV3)  // CVSS first
case d.Severity != 0:    return d.Severity                        // advisory second
```

### `getVendorSeverity()` — populates `VendorSeverity` map entries

```go
case detail.SeverityV3 != SeverityUnknown: vs[vendor] = detail.SeverityV3  // advisory first
case detail.CvssScoreV3 > 0:              vs[vendor] = scoreToSeverity(...) // CVSS second
```

## Real-world Impact

For **CVE-2026-0861** (RedHat data):
- RedHat advisory severity: **LOW**
- RedHat CVSS V3 score: **8.1 → HIGH**

| Field | Value | Rule used |
|-------|-------|-----------|
| `VendorSeverity["redhat"]` | **LOW** | advisory wins |
| `Severity` | **HIGH** | CVSS wins |

The same source data produces different severity values depending on which
field is used to read it.

## Expected Behavior

Both functions should use the same priority rule. The question is which is correct:

- **Advisory-first** (`VendorSeverity` behavior): respects the vendor's explicit
  severity assessment, uses CVSS only as a fallback when no advisory rating exists.
- **CVSS-first** (`Severity` behavior): uses the numeric score regardless of
  the vendor's own rating.

Advisory-first is likely the intended behavior, since the whole point of
`VendorSeverity` is to use vendor judgement over raw CVSS scores.

PR #669, opened 2026-05-20
fix: make Severity and VendorSeverity use consistent advisory first priority

nikpivkin:
Extracted a `detailToSeverity` helper with advisory-first priority (advisory severity → CVSS fallback). Both `getSeverity` and `getVendorSeverity` now delegate to it, giving consistent results.

- Added a CVSS-only fallback test (no advisory severity set)
- Added a regression test for the exact scenario from issue (vendor LOW advisory vs HIGH CVSS score)

Related issues:
- #650


DmitriyLewen (approved):
LGTM

PR #670, opened 2026-05-21
ci: declare workflow-level `contents: read` on cron and go workflows

arpitjain099:
Both workflows are scheduled / test steps; no GitHub API writes from the workflows. Workflow-level `contents: read` is the right cap for the default `GITHUB_TOKEN`.

Same post-CVE-2025-30066 (`tj-actions/changed-files`) hardening pattern. YAML validated locally.

### DIFF
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -10,6 +10,7 @@ env:
   VERSION: 2
 permissions:
   contents: read
+  packages: write  # for pushing to ghcr.io
 
 jobs:
   build:
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -8,6 +8,9 @@ env:
   REPO_OWNER: ${{ github.repository_owner }} # used in 'make db-build'
   GH_USER: aqua-bot
   VERSION: 2
+permissions:
+  contents: read
+
 jobs:
   build:
     name: Build DB
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -1,6 +1,9 @@
 ---
 name: Go
 on: [push, pull_request, merge_group]
+permissions:
+  contents: read
+
 jobs:
   test:
     name: Test


.github/workflows/cron.yml
```diff
@@ -8,6 +8,9 @@ env:
   REPO_OWNER: ${{ github.repository_owner }} # used in 'make db-build'
   GH_USER: aqua-bot
   VERSION: 2
+permissions:
+  contents: read
```
DmitriyLewen, line 12: You need to add `packages: write` to push into ghcr.io

DmitriyLewen (commented):
Hello @arpitjain099 
thanks for your work!

left a comment.

can you also run both workflows in your fork to see that workflows work correctly?

arpitjain099:
> Hello @arpitjain099 
> thanks for your work!
> 
> left a comment.
> 
> can you also run both workflows in your fork to see that workflows work correctly?

Hi @DmitriyLewen sure. Let me check them tonight.

arpitjain099:
Thanks @DmitriyLewen. Added `packages: write` to `cron.yml` since it pushes images to ghcr.io, kept `go.yml` at `contents: read` only.

On testing in fork: I'm running both workflows in my fork now to confirm they pass with these scopes. I'll come back with the run links shortly.

arpitjain099:
Quick follow-up on the fork test: I hit some friction.

`cron.yml` shows up in my fork's filesystem but isn't registered as an active workflow in the Actions API (`/actions/workflows` only lists `go.yml`, probably because GitHub doesn't auto-register workflows that were added upstream after I forked). `gh workflow run cron.yml` returns 404, and dispatch isn't an option from the API.

`go.yml` is registered but only triggers on `push/pull_request/merge_group` (no `workflow_dispatch`), and my PR-branch pushes don't seem to be triggering it on the fork either - might be the fork's Action policy on non-default-branch pushes.

I can either:
1. Add a temporary `workflow_dispatch` to both files on the PR branch just for testing (would be a separate commit I'd revert before merge).
2. Trust the static reasoning: `packages: write` is the documented minimum for `docker/login-action` against ghcr.io + `docker/build-push-action` push, both of which `cron.yml` uses. `go.yml` only runs unit tests, no ghcr.io push.
3. Wait for whatever your CI infra runs on the upstream PR (right now only `license/cla` shows up - might be gated on first-time-contributor approval).

Let me know which you'd prefer.

DmitriyLewen:
Hello @arpitjain099 
Thanks for update!

> go.yml is registered but only triggers on push/pull_request/merge_group (no workflow_dispatch), and my PR-branch pushes don't seem to be triggering it on the fork either - might be the fork's Action policy on non-default-branch pushes.

This workflow tested here (in this PR) - https://github.com/aquasecurity/trivy-db/actions/runs/26435967333/job/77848664406?pr=670

> cron.yml shows up in my fork's filesystem but isn't registered as an active workflow in the Actions API (/actions/workflows only lists go.yml, probably because GitHub doesn't auto-register workflows that were added upstream after I forked). gh workflow run cron.yml returns 404, and dispatch isn't an option from the API.

By default GitHub disables scheduled workflows for forks.
You need to enable it.

These steps should help:
First, make sure the workflow is disabled rather than missing:
```bash
gh workflow list --all -R arpitjain099/trivy-db
```
cron.yml should appear with the state disabled_fork. Then enable and dispatch it:
```bash
gh workflow enable cron.yml -R arpitjain099/trivy-db
gh workflow run cron.yml -R arpitjain099/trivy-db --ref chore/declare-workflow-perms
```
After that you will be able to run the cron workflow from the required branch.

A few notes regarding this workflow:
- In the repo settings you need to grant Actions write permissions for Packages (so that you can push to ghcr.io).
- In addition to ghcr.io there is also DockerHub and ECR. To avoid pushing everywhere, I usually remove them for testing.

Here is how I do it:

- Create a branch based on your fix (arpitjain099:chore/declare-workflow-perms).
- Remove the DockerHub and ECR login + remove the push from the script.
- Push this branch and run the workflow from the test branch.
- If changes are made to the base branch (arpitjain099:chore/declare-workflow-perms) — simply rebase the test branch.

arpitjain099:
@DmitriyLewen thanks for the walkthrough. Couple things to report from running the workflow on my fork:

**Workflow registration on the fork:** `cron.yml` wasn't visible to `gh workflow list --all` even after the upstream sync (looks like GitHub doesn't auto-register workflow files added upstream after a fork was created). I worked around it by pushing a tiny update to `cron.yml` on my fork's `main`, which got it registered.

**Dispatched run on `chore/declare-workflow-perms`:** [run 26499230611](https://github.com/arpitjain099/trivy-db/actions/runs/26499230611). It cleared the `actions/checkout` and 5 other steps (so the `contents: read + packages: write` block is doing its job), then failed on step 6 with:
```
Downloading https://github.com/arpitjain099/vuln-list/archive/main.tar.gz...
2026-05-27 08:16:43 ERROR 404: Not Found.
make: *** [Makefile:113: db-fetch-vuln-list] Error 8
```
That's a fork-data dependency, not a permissions issue - the Makefile resolves `REPO_OWNER` from `github.repository_owner`, which is `arpitjain099` on the fork, and `arpitjain099/vuln-list` doesn't exist. Same would happen for the DockerHub + ECR push steps later on (no secrets on the fork), per your earlier suggestion to strip those for fork testing.

So as far as the perms change in this PR goes, the auth surface is happy: `contents: read` is sufficient for checkout, and `packages: write` (which is already in the PR for `cron.yml`) covers the ghcr.io push path. Happy to do anything else you'd like checked, or merge whenever you're ready.

DmitriyLewen:
Sorry, I keep forgetting that not everyone has forks for all the Trivy-related repos 😄

To save us some time, I tested this on my fork.
Here are the run results: https://github.com/DmitriyLewen/trivy-db/actions/runs/26566271871
Everything builds and pushes to ghcr.io just fine — so `contents: read` + `packages: write` is enough.

And if you're curious — here's my testing branch (DockerHub + ECR login/push removed, ghcr.io only): https://github.com/DmitriyLewen/trivy-db/tree/test/declare-workflow-perms

DmitriyLewen (approved):
Thanks!

PR #671, opened 2026-06-08
feat(oracle-oval): align types with vuln-list-update source

manojkrishnanomula:
## Summary
- Add `Oval`, `Definition`, and `Issued` types to match the updated Oracle OVAL schema in [vuln-list-update](https://github.com/aquasecurity/vuln-list-update/blob/main/oracle/oval/types.go)
- Replace `IssuedDate`/`Date` with `Issued` to support the new JSON field format (`"Issued": {"Date": "..."}`)
- Keep `OracleOVAL` as a type alias for `Definition` so existing callers continue to compile without changes

This is a preparatory change to open the flow for upcoming vuln-list-update data format changes. No parsing or ingestion logic is updated in this PR.

## Test plan
- [x] `go build ./pkg/vulnsrc/oracle-oval/...`
- [x] `go test ./pkg/vulnsrc/oracle-oval/...`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/pkg/vulnsrc/oracle-oval/types.go
+++ b/pkg/vulnsrc/oracle-oval/types.go
@@ -22,7 +22,12 @@ type Reference struct {
 type Cve struct {
 	Impact string
 	Href   string
-	ID     string
+	Public string `json:",omitempty"`
+	// Oracle encodes cvss2 and cvss3 as "score/vector" (e.g. cvss3="7.3/CVSS:3.1/AV:N/...").
+	// Stored verbatim; downstream consumers split into score/vector themselves.
+	CVSS2 string `json:",omitempty"`
+	CVSS3 string `json:",omitempty"`
+	ID    string
 }
 
 type Criteria struct {


pkg/vulnsrc/oracle-oval/types.go
```diff
@@ -22,7 +22,12 @@ type Reference struct {
 type Cve struct {
 	Impact string
 	Href   string
-	ID     string
+	Public string `json:",omitempty"`
+	// Oracle encodes cvss2 and cvss3 as "score/vector" (e.g. cvss3="7.3/CVSS:3.1/AV:N/...").
+	// Stored verbatim; downstream consumers split into score/vector themselves.
+	CVSS2 string `json:",omitempty"`
+	CVSS3 string `json:",omitempty"`
```
DmitriyLewen, line 29: Is the omitempty tag needed here?
Go uses it only for encoding (https://pkg.go.dev/encoding/json#Marshal), but here we are using it only for decoding.
The exception is a custom Put, but in that case all fields from cve should have this tag.
If this tag is not needed — I suggest removing it.

DmitriyLewen (commented):
The PR description mentions more changes than what is actually there.
Is this an incorrect description or missing (unpushed) commits?

PR #678, opened 2026-06-23
fix(vulnerability): dedupe references using canonical URLs

h4r7w3l1:
## Summary

Vulnerability references from multiple upstream sources are merged in `getReferences()` and deduplicated by exact string equality. Equivalent URIs with different encodings can therefore survive into the trivy-db record and later appear as duplicate references in Trivy JSON, CycloneDX, and other output formats.

This change canonicalizes absolute reference URLs before deduplication and stores the canonical `String()` form. Non-absolute and unparseable references are left unchanged.

Fixing this at the trivy-db source level ensures all downstream Trivy exporters benefit without per-format deduplication logic.

## Background

This came from a Trivy PR review (aquasecurity/trivy#10877), where the duplicate references were confirmed to already exist in the DB-backed JSON report before CycloneDX conversion.

The reviewer suggested moving the fix to `trivy-db` so references are canonicalized and deduplicated at the source instead of only in a single Trivy output formatter.

## Example

Equivalent advisory references from different upstream sources may use different URI encodings, for example:

```text
https://lists.apache.org/thread.html/example%40%3Cdev.apache.org%3E   (NVD, fully encoded)
https://lists.apache.org/thread.html/example@%3Cdev.apache.org%3E
https://lists.apache.org/thread.html/example@<dev.apache.org>         (decoded)
```

These represent the same semantic URI but are different raw strings, so before this change all three could survive raw string deduplication as separate references.

After this change, they collapse to a single canonical reference:

```text
https://lists.apache.org/thread.html/example@%3Cdev.apache.org%3E
```

## Implementation details

`canonicalizeReferenceURL()` normalizes an absolute URL following [RFC 3986 §6.2.2](https://datatracker.ietf.org/doc/html/rfc3986#section-6.2.2):

* **Case normalization ([§6.2.2.1](https://datatracker.ietf.org/doc/html/rfc3986#section-6.2.2.1)):** `scheme` and `host` are lowercased.
* **Per-segment percent-encoding normalization:** each path segment is decoded and re-encoded, which normalizes percent-encoding case and collapses equivalent encodings of characters that are legal in a path (e.g. `%40` ↔ `@`).
* **Encoded slashes are kept distinct from the separator:** an encoded slash (`%2F`) is a data byte inside a segment, not a path separator, and denotes a different resource ([RFC 3986 §3.3](https://datatracker.ietf.org/doc/html/rfc3986#section-3.3)). Normalization is therefore done per segment so `%2F` is preserved rather than merged into `/`.
* **Query and fragment encoding are preserved:** the query is kept verbatim (`RawQuery`) and the fragment encoding is preserved (`RawFragment`).
* **Only `Scheme`/`Host`/`Path`/`Query`/`Fragment` are kept:** advisory links never carry `UserInfo` (`user:pass@`) or opaque data, so those components are intentionally dropped.

`getReferences()` already trims whitespace around each original reference and splits multiline reference values; this PR preserves that behavior and only adds canonicalization before the existing map-based deduplication step.

## Scope

This change does not alter non-absolute or unparseable references.
It also does not change vulnerability source parsing, severity calculation, CVSS handling, or any output-format-specific exporter code. 

To prevent unnecessary regressions and maintain high performance, full URI normalization rules are intentionally restricted. The following operations are **out of scope**:
* Sorting of query parameters (e.g., `?a=1&b=2` vs `?b=2&a=1`)
* Dot-segment removal / path cleaning (e.g., resolving `/../`)
* Default-port normalization (e.g., removing `:80` or `:443`)

The normalization happens once at the source-level reference aggregation step.

## Test plan

* [x] `go test ./pkg/vulnsrc/vulnerability/... -v -count=1`
* [x] Equivalent Apache mailing-list URL variants collapse to a single canonical reference (covered in `TestNormalize`, refs from NVD/RedHat/GHSA)
* [x] Distinct URLs remain distinct
* [x] Invalid or non-absolute references preserve existing behavior
* [x] Encoded slash (`%2F`) inside a segment is kept distinct from the `/` separator
* [x] Query strings and fragments (including `%2F`) are preserved after path canonicalization
* [x] Host casing is normalized (lowercased)


pkg/vulnsrc/vulnerability/references_test.go
```diff
@@ -0,0 +1,139 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/aquasecurity/trivy-db/pkg/types"
+)
+
+func TestCanonicalizeReferenceURL(t *testing.T) {
```
DmitriyLewen: Can you add/update test case with `%2F`?

pkg/vulnsrc/vulnerability/references_test.go
DmitriyLewen, line 1: I don't think that we should create separate file for these tests.

You can export  `canonicalizeReferenceURL` function in export.go and use it.
See https://github.com/aquasecurity/trivy/blob/c233735b023974c0a5c321c370c7f3ebe5f20eec/pkg/licensing/export_test.go as example.

pkg/vulnsrc/vulnerability/references_test.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert.Equal(t, tt.want, canonicalizeReferenceURL(tt.raw))
+		})
+	}
+}
+
+func TestGetReferences(t *testing.T) {
```
DmitriyLewen: I think we can remove this test and add refs from multiple sources into testcase from TestNormalize.

pkg/vulnsrc/vulnerability/vulnerability.go
```diff
@@ -154,6 +155,23 @@ func getCweIDs(details map[types.SourceID]types.VulnerabilityDetail) []string {
 	return nil
 }
 
+func canonicalizeReferenceURL(raw string) string {
```
DmitriyLewen, line 174: I think it makes sense to add a comment about:
- what function does
- examples with dedup
- info why we use only these parameters. e.g. advisory links don't have UserInfo.

pkg/vulnsrc/vulnerability/references_test.go
h4r7w3l1: Great catch - adding the `%2F` case surfaced an actual bug, not just a missing test.

The previous implementation copied the already-decoded `u.Path` without `RawPath`, so `u.String()` re-escaped the path in path mode where `/` is treated as a separator and is left unescaped. As a result, an encoded slash was silently turned into a real path separator:

~~~text
https://example.com/a%2Fb  →  https://example.com/a/b  # different resource (RFC 3986 §3.3)
~~~

I reworked `canonicalizeReferenceURL` to normalize per path segment: each segment is individually `url.PathUnescape`'d and `url.PathEscape`'d. Since `url.PathEscape` re-encodes a literal `/` back to `%2F`, the encoded slash stays distinct from the path separator while equivalent encodings still collapse (e.g., `%40` ↔ `@`, which is exactly what the Apache mailing-list references need).

Added test case: "encoded slash inside a segment is kept distinct from the separator" (`a%2Fb` stays `a%2Fb`).

pkg/vulnsrc/vulnerability/references_test.go
h4r7w3l1: Done. Removed the standalone `getReferences` test and added a `TestNormalize` case ("equivalent references from multiple sources collapse to one canonical URL") with the same Apache variants coming from NVD / RedHat / GHSA, asserting a single canonical reference in the normalized output.

I kept a focused unit test for `canonicalizeReferenceURL` itself (via `export_test.go`), since the RFC edge cases (`%2F`, fragment, host casing) are much easier to cover thoroughly at that level.

pkg/vulnsrc/vulnerability/references_test.go
h4r7w3l1: Done. Removed the standalone `getReferences` test and added a `TestNormalize` case ("equivalent references from multiple sources collapse to one canonical URL") with the same Apache variants coming from NVD / RedHat / GHSA, asserting a single canonical reference in the normalized output.

I kept a focused unit test for `canonicalizeReferenceURL` itself (via `export_test.go`), since the RFC edge cases (`%2F`, fragment, host casing) are much easier to cover thoroughly at that level.

pkg/vulnsrc/vulnerability/vulnerability.go
h4r7w3l1: Added a doc comment covering:
* What the function does
* The dedup example (the three Apache variants collapsing to one)
* The [RFC 3986 §6.2.2](https://datatracker.ietf.org/doc/html/rfc3986#section-6.2.2) rules applied (lowercase scheme/host, per-segment percent-encoding normalization, `%2F` kept distinct from the separator, and query/fragment preserved)
* Why only `Scheme`/`Host`/`Path`/`Query`/`Fragment` are kept (advisory links never carry `UserInfo` or opaque data, so those are dropped)

h4r7w3l1:
Hi @DmitriyLewen, 

Thanks a lot for the review and the great catch with the `%2F` case!

Summary of the update beyond the review notes: while addressing the `%2F` test I found the original canonicalization:
1. Decoded `%2F` into a real path separator `/` (which changes the identity of the resource according to RFC 3986 §3.3)
2. Did not actually preserve fragment encoding (only `RawQuery` was kept verbatim)

The new per-segment approach fixes both and also lowercases the host. PR description note about fragments is now accurate. 

While it intentionally leaves full RFC-normalization rules (such as query parameter sorting, default-port normalization, or dot-segment removal) out of scope as unnecessary overhead for advisory links, it perfectly solves the duplicate issues we faced.

Tests: `go test ./pkg/vulnsrc/vulnerability/...` passes.

pkg/vulnsrc/vulnerability/vulnerability.go
```diff
@@ -154,6 +155,61 @@ func getCweIDs(details map[types.SourceID]types.VulnerabilityDetail) []string {
...
+	}
+
+	segments := strings.Split(u.EscapedPath(), "/")
+	for i, s := range segments {
+		// url.PathEscape re-encodes a literal "/" back to %2F, so encoded
+		// slashes inside a segment are preserved rather than merged with the
+		// path separator.
+		if decoded, err := url.PathUnescape(s); err == nil {
+			segments[i] = url.PathEscape(decoded)
+		}
```
DmitriyLewen: nit:
this error is impossible, because we would get this error in url.Parse(raw).
```suggestion
		decoded, _ := url.PathUnescape(s)
		segments[i] = url.PathEscape(decoded)
```

pkg/vulnsrc/vulnerability/vulnerability.go
```diff
@@ -154,6 +155,61 @@ func getCweIDs(details map[types.SourceID]types.VulnerabilityDetail) []string {
...
+//     are legal in a path (e.g. %40 <-> @).
+//
+// Re-encoding is done per segment on purpose: an encoded slash (%2F) is a data
+// byte inside a segment and is kept distinct from the "/" separator, since they
+// denote different resources (RFC 3986 §3.3). The query string is preserved
+// verbatim (RawQuery) and the fragment encoding is preserved (RawFragment).
+//
+// Only Scheme/Host/Path/Query/Fragment are kept: advisory links never carry
+// UserInfo (user:pass@) or opaque data, so those components are intentionally
+// dropped. Non-absolute or unparseable references are returned unchanged.
```
DmitriyLewen: nit:
A bit smaller 😄 
wdyt?

```suggestion
// canonicalizeReferenceURL returns a canonical form of an absolute reference
// URL so that semantically equivalent links from different upstream sources
// deduplicate into a single reference. Non-absolute or unparseable references
// are returned unchanged.
//
// The same URI is often encoded differently per source; these collapse to one:
//
//	https://lists.apache.org/thread.html/example%40%3Cdev.apache.org%3E (NVD, fully encoded)
//	https://lists.apache.org/thread.html/example@%3Cdev.apache.org%3E
//	https://lists.apache.org/thread.html/example@<dev.apache.org>       (decoded)
//	=> https://lists.apache.org/thread.html/example@%3Cdev.apache.org%3E
//
// Following RFC 3986 §6.2.2, it lowercases scheme and host and re-encodes the
// path segment by segment, normalizing percent-encoding while keeping an encoded
// slash (%2F) distinct from the "/" separator. Query and fragment are preserved
// verbatim; UserInfo and opaque data are dropped (advisory links never carry them).
```

DmitriyLewen (commented):
LGTM.
left nit comments.

DmitriyLewen (approved):
@h4r7w3l1 Thanks for your work!

PR #680, opened 2026-06-27
fix(utils): skip dotfiles in FileWalk so .DS_Store does not break the DB build

### LINKED ISSUE #247: Mac users can not run successfully because of .DS_Store file
When I try to new and build a vulndb,there are some error like this
```
failed to decode vuln-list/oval/redhat/6/rhel-6-including-unpatched/definitions/.DS_Store:
    github.com/aquasecurity/trivy-db/pkg/vulnsrc/redhat-oval.parseOVALStream.func1
       trivy-db/pkg/vulnsrc/redhat-oval/redhat-oval.go:309
  invalid character '\x00' looking for beginning of value
```
Mac Almost every folder has a .DS_Store，Deleting them one by one is too cumbersome.Do you want to consider dealing with this situation?

arpitjain099:
Fixes #247.

On macOS a stray `.DS_Store` can end up inside a cloned vuln-list data dir. FileWalk hands every non-empty file to its callback, and most vulnsrc parsers then run it through json.Decode, so the binary `.DS_Store` fails with an "invalid character" error and the whole DB build aborts. A few sources (redhat-oval, debian, osv, wolfi, chainguard) already guard with a `.json` suffix check, but most (alpine, amazon, photon, etc.) decode whatever they're given, so one dotfile is enough to break them.

Rather than add a filter to each unguarded parser, I skip dotfiles once in the WalkDir callback. Dotfiles are never legitimate vulnerability data, so dropping them up front is safe for every caller. I avoided a `.json`-only filter on purpose: TestFileWalk asserts that the extensionless file foo3 still gets passed through, and skipping by leading dot keeps that working.

I extended TestFileWalk with a non-empty `.DS_Store` and an assertion that it never reaches the callback; that fails on main and passes here. `go test ./pkg/utils/...` and `go vet` are clean.

The issue is a few years old with no recent maintainer note, so I kept this minimal and focused on the concrete crash. Happy to push the filter into the callers instead if you'd prefer.

### DIFF
--- a/pkg/utils/file.go
+++ b/pkg/utils/file.go
@@ -8,6 +8,7 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/samber/oops"
 
@@ -29,6 +30,13 @@ func FileWalk(root string, walkFn func(r io.Reader, path string) error) error {
 		if err != nil {
 			return eb.Wrapf(err, "walk dir error")
 		} else if d.IsDir() {
+			return nil
+		}
+
+		// Skip dotfiles such as .DS_Store. They are not vulnerability data
+		// and most vulnsrc parsers json.Decode every file handed to walkFn,
+		// which fails on a binary .DS_Store with "invalid character".
+		if strings.HasPrefix(d.Name(), ".") {
 			return nil
 		}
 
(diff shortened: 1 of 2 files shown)


DmitriyLewen:
Hello @arpitjain099 and thanks for the PR.

  Issue #247 was created almost 4 years ago, and in all that time you're the first to run into this problem again.

  The trivy-db database is built in a pipeline from already "prepared" data that doesn't require any changes.
  So the likelihood of a .DS_Store file appearing is extremely low.

  For this reason I'm closing this PR: changes like this aren't needed, and excluding just the .DS_Store file doesn't look like a complete solution.

  Best regards,
  Dmitriy

PR #684, opened 2026-07-14
Feat/nvd cna metrics fallback

### LINKED ISSUE #575: Clarification on CVSS Data Source in Trivy DB
Hi team,

It seems Trivy DB only stores CVSS metrics from nvd@nist.gov, and ignores scores provided by other CNAs.

Example: [CVE-2025-26519](https://nvd.nist.gov/vuln/detail/CVE-2025-26519)
has no NVD score, but includes one from MITRE.

Was this an intentional choice, or a feature gap? If it’s the latter, I’d be happy to contribute support for fetching CVSS data from additional CNAs.

Thanks!

dor-hayun:
Adding a fallback to CNA/CISA-ADP CVSS metrics in the NVD source.

Today only `nvd@nist.gov` metrics are used, but many recent CVEs only have a
CNA/ADP metric — so they end up with no CVSS and no severity from NVD
(e.g. CVE-2026-39834: ADP says 9.1 CRITICAL, trivy-db shows MEDIUM).

Now, per CVSS version: use the NVD metric if it exists (unchanged), otherwise
take the first CNA/ADP metric.

Resolves #575

## Checklist
- [x] I've read the guidelines for contributing to this repository.
- [x] I've added tests that prove my fix is effective or that my feature works.


gshanbhag525:
shouldn't we add a new field for secondary metrics whcih are coming from CNA.
so in future if NIST adds metrics they go to primary field

dor-hayun:
> shouldn't we add a new field for secondary metrics whcih are coming from CNA. so in future if NIST adds metrics they go to primary field

I don't think so, as the logic is intentionally: NVD's own metric wins; CNA/ADP only fills the gap while NVD hasn't scored it. The DB is rebuilt from the raw NVD feed on every cycle, so the selection re-runs each build the moment NIST adds its own (primary) metric, the Source == "nvd" check makes it win automatically. The CNA value was only ever a fallback; it doesn't persist or "stick," so there's no stale carry-over to guard against and primary takes over on the next build.

dor-hayun:
HI @DmitriyLewen @knqyf263 
Would appreciate your review here

DmitriyLewen:
Hi @dor-hayun, thanks for the PR and for describing the problem in detail, and sorry for the late reply.

The `nvd` source in our DB means one specific thing — the assessment performed by NIST.
CNA and CISA-ADP metrics are marked as `"type": "Secondary"` by NVD itself, and we take only `nvd@nist.gov` so that `SeveritySource: nvd` and `CVSS.nvd` in the report mean exactly what they say.
If we put someone else's assessment there, the user sees `nvd` while the number comes from a vendor or from CISA — and comparing NVD against vendor sources becomes impossible, even though that comparison is the whole reason the sources are kept separate.
The same point was raised earlier in the review by @gshanbhag525.

The second point is scale.
I looked at CVEs published in NVD during June 2026 (a sample of 2000 records): only 14.5% have an `nvd@nist.gov` metric, while 85% have CNA/CISA-ADP metrics only.
So this is not about filling rare gaps — it replaces the contents of the source for the vast majority of recent CVEs.
On top of that, `nvd` comes first in the source precedence order, which makes it the default severity everywhere there is no vendor advisory.
In the same sample roughly half of the substituted assessments are HIGH and CRITICAL, so severity would shift for a large number of vulnerabilities at once, with no way for users to opt out.

Third, this takes away the only way to say "show me only what NIST actually assessed".
Today `--vuln-severity-source nvd` does that, and we recommend it exactly for these cases (see https://github.com/aquasecurity/trivy/discussions/9928, for example).
Once the metrics are merged, that option is gone.

It is also worth noting that the CNA assessment usually reaches Trivy through another path anyway: GHSA takes the CVSS from the CNA, and Red Hat, Debian and Ubuntu provide their own.
So in most cases that number is already in the report under its real source.

dor-hayun:
Hi @DmitriyLewen ,

The current Trivy DB data doesn’t have a CVSS score or severity coming from NVD for many CVEs. So, are you saying you would prefer missing data rather than using the metrics from an authorized CNA or CISA-ADP?

In my opinion, it would be better to prioritize NVD when the data is available. However, if there’s no data from NVD, I’d prefer to fall back to the other providers rather than have no data at all.

DmitriyLewen:
We have other supported sources for CVSS and severity. Using flags, people can control the sources, and the json report contains CVSS and severity for each source.

You, however, are suggesting mixing NVD and CNA/CISA-ADP. Even though CNA/CISA-ADP are shown on NVD pages, this could cause confusion for users.

We could think about adding this data under separate keys (as @gshanbhag525 also suggested), but before that, let's open a discussion in Trivy and discuss these changes with the community.

PR #686, opened 2026-07-29
feat(chainguard): use the Chainguard OSV v3 feed for Chainguard and Wolfi

plerion-sina:
## Description

Chainguard's secdb feed (`security.json`) is deprecated. This switches the Chainguard and Wolfi sources to [Chainguard's OSV v3 feed](https://github.com/chainguard-dev/vulnerability-scanner-support/blob/main/docs/osv_v3_feed.md), which covers both ecosystems and, unlike secdb, also publishes the advisories Chainguard has not resolved yet.

secdb published one rolled up entry per package: a map of fixed version to the CVEs fixed in it. The v3 feed publishes one record per **vulnerable component**, so the rollup secdb did for us is now the consumer's job. Up to 106 records cover a single package, architecture and CVE. Chainguard documents how they combine, and that is what `Aggregate` implements:

- any unresolved record means the package is affected, with no fix version
- otherwise the fix is the highest version any record names, because that is the first version in which every component is fixed
- if every record is a false positive determination, nothing is stored

Advisories are stored per architecture, since the two builds of a package are not always fixed in the same revision (`prism` CVE-2025-25289 is fixed in `5.14.3-r1` on aarch64 and `5.14.3-r8` on x86_64). Every architecture is returned from `Get` with `Arches` populated, and choosing between them is left to Trivy, because the feed covers only one architecture for about one package/CVE pair in eleven and dropping a non-matching one would hide it entirely.

For unresolved advisories the resolution status is kept, so `--ignore-status` can act on it:

| Chainguard status | Trivy status |
|---|---|
| `true_positive_determination` | `affected` |
| `pending_upstream_fix` | `fix_deferred` |
| `fix_not_planned` | `will_not_fix` |
| `analysis_not_planned` | `will_not_fix` |
| `detection` | `under_investigation` |

The CGA advisory ID that determined the outcome goes into `VendorIDs`.

Parsing is shared between the two ecosystems in the new `chainguardosv` package, because a Chainguard image and its Wolfi variant have to report the same vulnerabilities. `chainguard.VulnSrc` and `wolfi.VulnSrc` stay distinct types wrapping it.

## Backward compatibility

This changes the shape of two buckets without a schema bump, so both directions are handled:

- **New code, old database.** `Get` falls back to the bare `FixedVersion` when there are no entries, the way the Rocky source does. Without this a Trivy built from this code would accept the currently published database and report **no** Chainguard or Wolfi vulnerabilities at all, silently. Verified against the real released `trivy.db`: results are identical to an unpatched Trivy.
- **Old code, new database.** The top level `FixedVersion` is still written. Verified with an unpatched Trivy binary against a database built from this code: same 23 findings on `scanner-test:unfixed-vulnerabilities-chainguard`.

## Database size

Measured with `bbolt compact` then `tar czf`, the same as the release pipeline. The "before" reproduction came out at 102.5MiB against the 102.79MiB actually downloaded today, so the method matches.

| | `trivy.db` | `db.tar.gz` |
|---|---|---|
| before | 1.14GiB | 102.5MiB |
| after | 1.24GiB | 112.1MiB |
| delta | +99.6MiB (+8.5%) | **+9.6MiB (+9.4%)** |

The Chainguard and Wolfi buckets alone go from 32.2MiB to 128.6MiB uncompressed, 1.69MiB to 11.75MiB compressed. Three things compound: 2.3x more packages, because v3 files an advisory against the subpackage that ships the component so subpackages get their own entries; 1.5x more advisories, because unresolved ones are now expressible; and about 3x more bytes per advisory, because each one now carries `Arches`, `Status` and `VendorIDs` where secdb stored `{"FixedVersion":"1.2.3-r4"}`.

| | packages | advisories |
|---|---|---|
| chainguard secdb | 4,393 | 172,236 |
| chainguard v3 | 10,203 | 272,277 |
| wolfi secdb | 1,461 | 57,691 |
| wolfi v3 | 3,411 | 83,686 |

If the +9.6MiB is a problem there is easy headroom: most entries carry `Arches: ["aarch64","x86_64"]`, and an absent `Arches` already means "all architectures", so omitting it in that case would save roughly 8-9MiB uncompressed. Dropping `VendorIDs` would save more, at the cost of traceability back to the Chainguard advisory. Say the word.

## Merge order

**This must not merge before https://github.com/aquasecurity/vuln-list-update/pull/475 has merged and its target has run at least once.** `Update()` walks `vuln-list/chainguard-osv/v3/<ecosystem>` and `utils.FileWalk` surfaces a missing directory as a walk error, which fails the entire DB build rather than yielding an empty bucket. Failing loudly is the right behaviour, it just makes the ordering strict.

One thing a reviewer will probably ask: the mirror pulls the tarball from `packages.cgr.dev` while the spec says `advisories.cgr.dev` is the new domain and `packages.cgr.dev` will eventually be deprecated. The tarball is only documented on `packages.cgr.dev`, so there is no alternative today.

## Related PRs

- [x] https://github.com/aquasecurity/vuln-list-update/pull/475
- [ ] https://github.com/aquasecurity/trivy/pull/11015

## Checklist

- [x] I've read the [guidelines for contributing](https://trivy.dev/docs/latest/community/contribute/pr/) to this repository.
- [x] I've followed the [conventions](https://trivy.dev/docs/latest/community/contribute/pr/#title) in the PR title.
- [x] I've added tests that prove my fix is effective or that my feature works.


plerion-sina:
Closing this one. We're going to carry these changes downstream rather than upstream for now.

The branch stays up if anyone wants to pick it up later, and the description above has the full write-up: what changes between secdb and the v3 feed, the aggregation rules, and the validation against Chainguard's `scanner-audit` across all 10 test cases and both distro variants.

Thanks for the project.

PR #692, opened 2026-09-21
feat(vulnsrc): extend RapidFort source to Debian, Oracle, Rocky, Alma, and Amazon

rfJanit:
## Summary

Follow-up to #654, which introduced the RapidFort vulnsrc for Red Hat, Ubuntu, and Alpine. This PR extends the source so it ingests advisories for the five other base distributions RapidFort now ships curated images for: **Debian, Oracle Linux, Rocky Linux, AlmaLinux, and Amazon Linux**.

### `newBucket` covers the additional bases

`pkg/vulnsrc/rapidfort/bucket.go` adds cases for `ecosystem.Debian`, `ecosystem.OracleLinux`, `ecosystem.Rocky`, `ecosystem.AlmaLinux`, and `ecosystem.AmazonLinux`, each wrapping the corresponding base bucket (`bucket.NewDebian`, `bucket.NewOracle`, `bucket.NewRocky`, `bucket.NewAlma`, `bucket.NewAmazon`) and tagging the datasource with the matching `BaseID`. An `unsupported base ecosystem` error is still returned for anything outside the supported set, so a stray directory under `OS/` is dropped instead of producing a bogus bucket.

`rapidFortBucket.Name()` uses `strings.TrimRight(..., " ")` because `oracleBucket.Name` and `amazonBucket.Name` unconditionally append the version — a version-less rebuild otherwise came back as `"Oracle Linux "` with a trailing space. Trimming ensures every distro spells its rebuild bucket the same way (`rapidfort Oracle Linux`, `rapidfort Amazon Linux`, ...).

### `resolveBucket` generalized

`pkg/vulnsrc/rapidfort/rapidfort.go` replaces the hard-coded `ubuntuIdentifier == "ubuntu"` case with `identifier == string(eco)`. One clause now covers every feed that tags its own distribution's packages by name — `"ubuntu"`, `"debian"`, `"alpine"`, `"redhat"`, `"oracle"`, `"rocky"`, `"alma"`, `"amazon"` — collapsing what would otherwise have been five near-identical additions. Empty identifiers still fall through to the same "keep eco and ecoVer as listed" branch.

A new `amazonIdentifier = "amzn"` constant plus its `strings.HasPrefix(identifier, amazonIdentifier)` branch handles Amazon Linux's `amzn2`/`amzn2023` release tags, which name only the release (like `elN`) while the feed still names the distribution.

### Cross-format bucket separation for `rf` ranges

RapidFort's own rebuilds carry the `"rf"` identifier and drop the release, so a single `"rapidfort"` bucket would end up mixing RPM ranges (`1:3.2.2-19.rf`) with dpkg ranges (`3.5.7-10rfubu+rf.0`) — the wrong comparator would then read them at scan time. The bucket wrapper keeps the base ecosystem, so every feed now writes to its own family-specific `rf` bucket: `rapidfort Red Hat`, `rapidfort ubuntu`, `rapidfort debian`, `rapidfort Oracle Linux`, `rapidfort Rocky Linux`, `rapidfort AlmaLinux`, `rapidfort Amazon Linux`. Comparator boundaries are preserved and one feed's `rf` range cannot overwrite another's.

### Test coverage

- `pkg/vulnsrc/rapidfort/rapidfort_test.go` grows from 785 to 1,464 lines with happy-path cases for each new distribution and negative cases covering unsupported identifiers, malformed paths, and cross-format collision attempts.
- `pkg/vulnsrc/rapidfort/route_test.go` (new) covers `resolveBucket` in isolation: per-identifier routing (`rf`, `elN`, `amznN`, `fcNN`, feed name, empty) and rejection of unusable inputs.
- `pkg/vulnsrc/rapidfort/bucket_test.go` (new) covers `newBucket` for all supported bases plus the unsupported-ecosystem error path.
- New per-OS happy-path fixtures under `testdata/happy/rapidfort-security-advisories/OS/`: `alma/curl.json`, `amazon/curl.json`, `debian/curl.json`, `oracle/curl.json`, `rocky/curl.json`.
- `testdata/unsupported_os/OS/photon/openssl.json` replaces the previous `debian/openssl.json` since Debian is now supported; Photon takes its place as the unsupported-OS regression case.
- `testdata/fixtures/happy.yaml` and `testdata/fixtures/data-source.yaml` are extended with the new buckets and data sources.


pkg/vulnsrc/rapidfort/rapidfort.go
```diff
@@ -204,18 +259,19 @@ func resolveBucket(eco ecosystem.Type, ecoVer, identifier string) (bucket.DataSo
 	// RapidFort's rebuilds are not tied to a distro release: they keep the feed's OS but drop the version.
 	case identifier == rapidFortIdentifier:
 		ecoVer = ""
+	// A feed tags its own distribution's packages by name ("ubuntu", "debian") or
+	// not at all (alpine): either way they belong to the release the file lists them under.
+	case identifier == string(eco), identifier == "":
```
DmitriyLewen, line 264: It looks like some ranges in your Oracle feed came without an `identifier`, so they land in the `rapidfort Oracle Linux N` buckets even though they are not Oracle ranges.

The code treats an empty `identifier` as a range of the feed's own distribution, from the release it is listed under — which holds for Alpine, where the feed doesn't tag its ranges at all.
I ran the current `main` of your `rapidfort/security-advisories` across all 8 OSes:

| OS | total events | without `identifier` |
|---|---|---|
| alpine | 58,139 | 58,139 (all) |
| oracle | 1,725,386 | **1,590** |
| redhat, rocky, alma, amazon, debian, ubuntu | — | 0 |

So untagged ranges are the norm in Alpine, don't exist at all in the other feeds, and are an anomaly in Oracle: 46 `rf-*_advisory.json` files (`rf-curl`, `rf-openssl`, `rf-expat`, `rf-libxml2`, `rf-pcre2-*`, `rf-systemd-*`, …), 232 package/CVE pairs, 318 events under each of the `6`, `7`, `8`, `9`, `10` keys.
By content, these are not Oracle ranges:
- 1,125 carry Fedora versions (`...fc43`, `...fc44`, `...fc46`);
- 460 carry bare upstream versions (`3.5.4`, `8.16.0`, `2.12.10`), 30 of them open (`>=X` with no `fixed`);
- 5 carry `el` versions.

Example — `OS/oracle/rf-curl_advisory.json`, key `"9"`, `CVE-2025-13034`:

```json
[
  {"introduced": "0", "fixed": "0:8.18.0-11.rf", "identifier": "rf"},
  {"introduced": "8.16.0", "fixed": "8.17.0~rc1-1.fc44"},
  {"introduced": "8.17.0", "fixed": "8.18.0-1.fc44"},
  {"introduced": "8.18.0", "fixed": "8.18.0-1.fc44"},
  {"introduced": "8.19.0", "fixed": "8.22.0~rc2-1.fc46"},
  {"introduced": "8.20.0", "fixed": "8.22.0~rc2-1.fc46"},
  {"introduced": "8.21.0", "fixed": "8.22.0~rc2-1.fc46"}
]
```

For the same CVE, `OS/redhat/rf-curl_advisory.json` tags every Fedora range (`"identifier": "fc44"` etc.), so they go to `rapidfort fedora NN`.
For Oracle, a DB built from this PR ends up with:

```
rapidfort Oracle Linux 9 / rf-curl / CVE-2025-13034
  PatchedVersions:    [8.17.0~rc1-1.fc44 8.18.0-1.fc44 8.22.0~rc2-1.fc46]
  VulnerableVersions: [>=8.16.0, <8.17.0~rc1-1.fc44 ...]
```

That is 232 records with such ranges in each of `rapidfort Oracle Linux 6..10` (140 of them with Fedora versions).
Impact at scan time (given the routing in aquasecurity/trivy#11285):
- false negative: a package with `.fc43`/`.fc44` in its release is looked up in `rapidfort fedora NN`, which doesn't have these ranges;
- false positive: a package with `.el9` is looked up in `rapidfort Oracle Linux 9` and compared against a Fedora/upstream fixed version (for open ranges it is flagged unconditionally).

Is this expected?
It looks like a bug in the feed (the `fcNN`/`rf` identifiers got lost). I don't think trivy-db should work around it.

pkg/vulnsrc/rapidfort/rapidfort.go
```diff
@@ -73,12 +74,66 @@ func (vs VulnSrc) Update(dir string) error {
...
 	return nil
 }
 
+// mergeEntries folds entries targeting the same (platform, package, CVE) into
+// one. split already collapses the ranges within a single file, but two files
+// can still describe the same artifact: every RPM feed carries fcNN ranges, and
+// those all resolve to the shared "rapidfort fedora NN" buckets. put writes one
+// entry after another, so without this only whichever file the walk visited last
+// would survive. Feeds with no such overlap pass through unchanged.
+func mergeEntries(entries []entry) []entry {
```
DmitriyLewen, line 89: When we added RapidFort support, we discussed where to keep the `fcNN` entries and agreed that the Fedora data isn't tied to the OS/ecosystem (any `fc` version can be installed in any environment), so it is enough to store it once.
That's exactly why we moved them into separate `rapidfort fedora NN` buckets shared by all RPM feeds, and planned to merge and deduplicate them across OSes (Oracle included).

But judging by `mergeEntries` and by the feed itself, the entries for the same `fcNN` differ between OSes/ecosystems.
For example, `memcached-devel` / `CVE-2026-47784` / `fc41`:

```
OS/redhat: {"introduced": "1.6.24-1.fc41", "identifier": "fc41"}                            # open
OS/oracle: {"introduced": "1.6.24-1.fc41", "fixed": "1.6.32-2.fc41", "identifier": "fc41"}  # fixed
```

And it's not a one-off: out of 378,173 (`fcNN`, package, CVE) combinations that appear in more than one of your feeds, the ranges differ in 327,072.

Is this a sync delay between the feeds?
Or is this expected behavior?
If it is expected, there are problems with `mergeEntries`:
1. False positives: for the example above, `rapidfort fedora 41` ends up with `VulnerableVersions: [">=1.6.24-1.fc41", ">=1.6.24-1.fc41, <1.6.32-2.fc41"]`, so `1.6.32-2.fc41` is still reported as vulnerable, even though the oracle feed says it's fixed.
2. Severity differs between OSes/ecosystems (259 cases), e.g. `rf-python3` / `CVE-2024-9287` / `fc41`: `HIGH` in redhat and `MEDIUM` in alma. `mergeEntries` takes the maximum, so a user scanning a RHEL image gets a severity rated for another OS.

pkg/vulnsrc/rapidfort/rapidfort.go
```diff
@@ -204,18 +259,19 @@ func resolveBucket(eco ecosystem.Type, ecoVer, identifier string) (bucket.DataSo
 	// RapidFort's rebuilds are not tied to a distro release: they keep the feed's OS but drop the version.
 	case identifier == rapidFortIdentifier:
 		ecoVer = ""
+	// A feed tags its own distribution's packages by name ("ubuntu", "debian") or
+	// not at all (alpine): either way they belong to the release the file lists them under.
+	case identifier == string(eco), identifier == "":
+		// Keep eco and ecoVer as the feed listed them.
 	// "elN" is the dist tag of the Enterprise Linux family, which RedHat, Oracle and the other rebuilds all share, so it names the release while the feed still names the distribution.
 	case strings.HasPrefix(identifier, redHatIdentifier):
```
DmitriyLewen, line 267: In the redhat and oracle feeds, `dnsmasq` and `dnsmasq-utils` have ranges with `"identifier": "el2"`, even though the versions are Fedora ones:

```
OS/oracle: {"introduced": "2.92rel2-9.fc45", "fixed": "2.93-3.fc45", "identifier": "el2"}
OS/redhat: {"introduced": "2.92rel2-9.fc45", "identifier": "el2"}
```

It looks like `el2` came from `rel2` in the upstream version `2.92rel2`, while it's actually `fc45`.
As a result, the DB gets `rapidfort Red Hat 2` and `rapidfort Oracle Linux 2` buckets (6 records each) that nobody will ever query, and `rapidfort fedora 45` is missing these ranges.
Is this expected, or was the `identifier` detected incorrectly here?

pkg/vulnsrc/rapidfort/route_test.go
```diff
@@ -0,0 +1,272 @@
...
+			eco:     ecosystem.PhotonOS,
+			ecoVer:  "5.0",
+			event:   Event{Identifier: "rf", Fixed: "8.11.1-1.rf"},
+			wantErr: "unsupported base ecosystem",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := resolveBucket(tt.eco, tt.ecoVer, tt.event.Identifier)
```
DmitriyLewen, line 261: nit: only `Identifier` is used from `event` — `resolveBucket` gets `tt.event.Identifier`, while `Fixed`/`Introduced` aren't passed anywhere.
This makes the cases look as if the version affects the bucket choice, which it doesn't.
I'd replace `event Event` with `identifier string` (and `event: Event{Identifier: "el9", Fixed: ...}` → `identifier: "el9"` in the cases):

```suggestion
			got, err := resolveBucket(tt.eco, tt.ecoVer, tt.identifier)
```

pkg/vulnsrc/rapidfort/route_test.go
```diff
@@ -0,0 +1,272 @@
...
+			event:    Event{Identifier: "ubuntu", Fixed: "0:2.46-10rfubu"},
+			wantName: "rapidfort ubuntu 22.04",
+		},
+		{
+			name:     "explicit rf identifier without a marker still splits out",
+			eco:      ecosystem.Ubuntu,
+			ecoVer:   "22.04",
+			event:    Event{Identifier: "rf", Fixed: "0:2.46-10"},
+			wantName: "rapidfort ubuntu",
+		},
```
DmitriyLewen, line 162: nit: these cases refer to a "marker heuristic", but there is none in trivy-db — `resolveBucket` only looks at `identifier`.
Their inputs match "ubuntu identifier on the ubuntu feed" and "rf on the ubuntu feed" (only `Fixed` differs, and it isn't used), so they don't check anything new.
I'd remove them:

```suggestion
```

pkg/vulnsrc/rapidfort/route_test.go
```diff
@@ -0,0 +1,272 @@
...
+
+	"github.com/aquasecurity/trivy-db/pkg/ecosystem"
+)
+
+// TestResolveBucket pins which bucket a single range lands in. One feed file can
+// mix distributions, so the range's identifier — not the version key it sits
+// under — decides, and the feed's own OS supplies whatever the identifier leaves
+// unsaid. These bucket names are the keys Trivy queries at scan time, so a
+// change here silently stops matching.
+func TestResolveBucket(t *testing.T) {
```
DmitriyLewen, line 17: nit: there is no `route.go` in the package — `resolveBucket` lives in `rapidfort.go`.
Let's move `TestResolveBucket` to `rapidfort_test.go`, so the test sits next to the code it covers.

pkg/vulnsrc/rapidfort/bucket_test.go
```diff
@@ -0,0 +1,201 @@
...
+		ecosystem.Type("nonexistent"),
+	} {
+		t.Run(string(eco), func(t *testing.T) {
+			got, err := newBucket(eco, "1")
+			require.Error(t, err)
+			assert.Nil(t, got)
+			assert.Contains(t, err.Error(), "unsupported base ecosystem")
+		})
+	}
+}
```
DmitriyLewen, line 201: nit: `TestNewBucket_UnsupportedEcosystem` can be folded into the `TestNewBucket` cases — it's the same function with the same inputs, only the expected result differs.
I'd add a `wantErr string` field to `TestNewBucket`, cases for the unsupported ecosystems, and an error check in the loop:

```go
		{
			name:      "unsupported ecosystem",
			ecosystem: ecosystem.PhotonOS,
			version:   "5.0",
			wantErr:   "unsupported base ecosystem",
		},
	}

	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			got, err := newBucket(tt.ecosystem, tt.version)
			if tt.wantErr != "" {
				require.ErrorContains(t, err, tt.wantErr)
				assert.Nil(t, got)
				return
			}
			require.NoError(t, err)
			// ...
		})
	}
```

And remove this function:

```suggestion
```

pkg/vulnsrc/rapidfort/rapidfort.go
```diff
@@ -27,12 +27,13 @@ const rapidfortDir = "rapidfort-security-advisories"
 // repo that groups advisory JSON files by operating system.
 const osSubDir = "OS"
 
-// Supported `Event.Identifier` values.
+// Supported `Event.Identifier` values. A feed's own distribution is named by
+// its ecosystem string ("ubuntu", "debian"), so it needs no constant here.
```
DmitriyLewen, line 31: nit: this comment describes the current code in terms of the previous behavior ("so it needs no constant here" only makes sense to someone who remembers `ubuntuIdentifier`).
I'd describe what lives here now: identifiers that differ from the feed's ecosystem name.

```suggestion
// Supported `Event.Identifier` values that differ from the feed's ecosystem name.
// Identifiers equal to the ecosystem name (e.g. "ubuntu", "debian") are matched against the feed directly.
```

pkg/vulnsrc/rapidfort/bucket.go
```diff
@@ -16,7 +18,10 @@ type rapidFortBucket struct {
 }
 
 func (r rapidFortBucket) Name() string {
-	return "rapidfort " + r.base.Name()
+	// Some base buckets append the version unconditionally, so a version-less
+	// rebuild comes back with a trailing space ("Oracle Linux "). Trim it so
+	// every distro spells its rebuild bucket the same way.
+	return strings.TrimRight("rapidfort "+r.base.Name(), " ")
 }
```
DmitriyLewen, line 25: nit: `TrimRight` treats the symptom: the trailing space appears because `oracleBucket.Name()` and `amazonBucket.Name()` in `pkg/vulnsrc/bucket/bucket.go` always append the version, even an empty one.
Let's fix the buckets themselves, following `redHatBucket.Name()` (don't append an empty version), and keep a plain concatenation here:

```suggestion
func (r rapidFortBucket) Name() string {
	return "rapidfort " + r.base.Name()
}
```

After that, the `strings` import in `bucket.go` is no longer needed.
Nothing changes for the other sources: `oracle-oval` and `amazon` always pass a non-empty version.

pkg/vulnsrc/rapidfort/rapidfort_test.go
```diff
@@ -298,6 +306,405 @@ func TestVulnSrc_Update(t *testing.T) {
...
+					},
+					Value: map[string]any{},
+				},
+			},
+			noBuckets: [][]string{
+				// The malformed "el5" top-level key is never used as a release:
+				// the range identifier decides the bucket, so no bucket is keyed
+				// on the raw key.
+				{"advisory-detail", "CVE-2024-EL5KEY", "rapidfort Oracle Linux el5"},
 			},
```
DmitriyLewen, line 708: nit: this `noBuckets` doesn't check anything: the `rapidfort Oracle Linux el5` bucket can't appear at all — the range has `identifier: "el5"`, and without it the `el5` key would be rejected by `isVersionNumber`.
That the range lands in `rapidfort Oracle Linux 5` is already checked above (L353-358), so this block can be removed:

```suggestion
```

pkg/vulnsrc/rapidfort/route_test.go
```diff
@@ -0,0 +1,272 @@
...
+			wantName: "rapidfort Red Hat 8",
+		},
+		// Amazon Linux tags releases "amznN" rather than "elN".
+		{
+			name:     "amzn2023 on the amazon feed",
+			eco:      ecosystem.AmazonLinux,
+			ecoVer:   "2023",
+			event:    Event{Identifier: "amzn2023", Fixed: "8.5.0-1.amzn2023"},
+			wantName: "rapidfort amazon linux 2023",
+		},
```
DmitriyLewen, line 72: nit: `TestResolveBucket` is missing a few edge cases: `amzn` with no release, `amzn1`, `el9.` (the `part == ""` branch in `isVersionNumber`), and `identifier == string(eco)` on an RPM feed.
I'd add:

```suggestion
			name:     "amzn2023 on the amazon feed",
			eco:      ecosystem.AmazonLinux,
			ecoVer:   "2023",
			event:    Event{Identifier: "amzn2023", Fixed: "8.5.0-1.amzn2023"},
			wantName: "rapidfort amazon linux 2023",
		},
		{
			name:     "amzn1 on the amazon feed",
			eco:      ecosystem.AmazonLinux,
			ecoVer:   "2023",
			event:    Event{Identifier: "amzn1", Fixed: "8.5.0-1.amzn1"},
			wantName: "rapidfort amazon linux 1",
		},
		{
			name:    "amzn with no release",
			eco:     ecosystem.AmazonLinux,
			ecoVer:  "2023",
			event:   Event{Identifier: "amzn", Fixed: "8.5.0-1"},
			wantErr: "unusable distribution version",
		},
		{
			name:    "el release with a trailing dot",
			eco:     ecosystem.RedHat,
			ecoVer:  "9",
			event:   Event{Identifier: "el9.", Fixed: "7.76.1-26"},
			wantErr: "unusable distribution version",
		},
		{
			name:     "redhat identifier on the redhat feed",
			eco:      ecosystem.RedHat,
			ecoVer:   "9",
			event:    Event{Identifier: "redhat", Fixed: "7.76.1-26.el9"},
			wantName: "rapidfort Red Hat 9",
		},
```

(If you accept the suggestion to replace `event` with `identifier`, these cases need the same change.)

DmitriyLewen (commented):
`pkg/vulnsrc/rapidfort/rapidfort.go` L166:

nit: Debian is supported now, so the example is outdated.

```go
		// RapidFort owns which OSes its feed ships, so an OS this build doesn't ingest (e.g. photon) is expected rather than something to warn about on every file.
```

---

`pkg/vulnsrc/rapidfort/rapidfort.go` L280:

nit: the list of bare dist tags is missing `"amzn"` — the same check rejects it.

```go
	// Only the rebuilds are release-less; every other range has to name a release, so a bare dist tag ("el", "amzn", "fc") or an empty version key is rejected rather than folded into a rebuild bucket.
```

PR #693, opened 2026-09-21
fix(ci): prevent OOM in the DB build with GOMEMLIMIT

DmitriyLewen:
## Description

The scheduled DB build fails on roughly half of its runs.
The failures all look the same: the build is killed part way through `make db-build`, the runner loses its agent, and every remaining step is skipped — including the failure notification, as noted in #691.

```
INFO [oracle-oval] Saving Oracle Linux OVAL
##[error]The runner has received a shutdown signal...
make: *** [Makefile:94: db-build] Killed
```

`Killed` with no error from the program is the OOM killer.
Disk is not involved — 119GB of 145GB are free after the cleanup step — and the job takes ~25 minutes against a 6 hour limit.

## Root cause

The build was profiled on a throwaway branch with `GODEBUG=gctrace=1`, `/usr/bin/time -v` and a sampler writing `MemAvailable` and the builder's RSS every 5 seconds.

The peak is not one source holding too much data.
It is the garbage collector: with the default `GOGC=100` the heap may grow to twice the live data before a collection runs.

```
12655->12660->586 MB, 13417 MB goal
```

The heap reached 12.7GB while only 586MB was live.
The largest live heap is ~7GB, which sets a GC goal of ~14GB — more than the 16GB runner has, and the cleanup step removes the swap file.
The build peaked at 13.5GB with 1.3GB left, so any small variation tips a run over.

Note that the process is killed while `oracle-oval` is running, but that source holds only ~1.6GB.
The peak is reached earlier and the memory is simply never returned to the OS, so the source named in the log is not the one to optimise.

`GOMEMLIMIT` caps the footprint instead of letting it scale with the live data.

## Measurements

Two runs per configuration, same commit, same workflow:

| | Wall | Peak RSS | Memory left | GC cycles |
|---|---|---|---|---|
| before | 21:21 | 13509MB | 1458MB | 336 |
| before | 25:30 | 13685MB | 1251MB | 323 |
| after | 50:17 | 10525MB | 4905MB | 330 |
| after | 31:04 | 10151MB | 4858MB | 335 |

Headroom grows from ~1.3GB to ~4.9GB, and the number of GC cycles is unchanged, so the limit does not push the collector into a death spiral.

The build gets slower: +22% on the cleanest pair, which started in the same second and spent the same time downloading the data.
The 50:17 run landed on a slow machine — `cocoapods`, which is IO bound and holds under 1GB, took 927s there against 314-475s in the other three runs.
On a ~25 minute build against a 6 hour job limit that cost is acceptable.

The limit is set on the step rather than in the Makefile, because it is tuned for the runner and would be wrong for anyone building the DB locally.

## Alternatives considered

Restoring the swap file that the cleanup step removes was also considered, and it would probably have worked.
The peak is ~12GB of garbage, and the mark phase never walks unreachable objects, so those pages are cold and would swap out well.

It only postpones the problem, though — for a long time, but not indefinitely.
`GOGC` makes the goal a multiple of the live heap, so the goal grows with the data: at ~8GB live it reaches ~16GB, and 4GB of swap no longer covers it.

`GOMEMLIMIT` is not a permanent fix either: as the live heap grows the value will have to be revisited.
The difference is that it is an explicit ceiling rather than one that drifts upwards with the data, and that exceeding it makes the build slower rather than less complete: the limit has no effect on what ends up in the DB.


DmitriyLewen:
Closed in favor of https://github.com/aquasecurity/trivy-db/pull/695 and https://github.com/aquasecurity/trivy-db/pull/694

